PR #12, opened 2024-05-01
RSS Feed Sync CMS Plugin

niekert:
### Description

This Plugin adds a simple RSS feed sync plugin. It uses the new [CMS Collections API](https://developers.framer.wiki/docs/cms) 

image

### Testing

- [ ] You can sync an RSS feed
  - [ ] In the CMS add a new collection
  - [ ] Load the plugin
  - [ ] Select one of the data sources
  - [ ] Click "Import"
  - [ ] The RSS data is imported in the CMS
- [ ] It synchronizes when you press "Synchronize"
  - [ ] Press "Synchronize" after the RSS feed changed
  - [ ] Notice how the CMS is synchronized with the RSS feed
- [ ] It lets you mange the data source
  - [ ] Click "Manage" in the toolbar
  - [ ] Select a different data source
  - [ ] Notice how the CMS now has different items as a data source


plugins/rss-feeds/src/main.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const rssSourceId = await collection.getPluginData("rssSourceId")
+  if (mode === "syncCollection" && rssSourceId) {
+    try {
+      await importData(collection, rssSourceId)
+      await framer.closePlugin()
+    } catch {
+      // Failed to synchronize - show configuration
+    }
+  }
```
huntercaron, line 20: Why is this logic in this file? I find it nice when everything is inside the App.tsx for these smaller examples

niekert: When a plugin is launched in `syncCollection` mode we can synchronize the CMS without displaying any UI.

The reason this is placed here is that we check if we can synchronize the data without displaying any UI.

If it is possible we skip mounting the React app altogether and close the plugin after syncing 🙏

huntercaron: ok makes sense! lets keep it then

huntercaron:
LGTM! Thank you sir

huntercaron (approved):
LGTM! Thank you sir

PR #14, opened 2024-05-07
Add Notion plugin

niekert:
### Description

This PR adds a Notion plugin. You can connect this Plugin to your Notion workspace and use it to sync Notion databases to the Framer CMS

https://github.com/framer/plugins/assets/5508905/4cfdc818-86e1-4788-9094-e2def7d449a8

### Testing

- [ ] Synchronizing pages works with the example provided by the plugin
  - [ ] Open the plugin from the URL https://plugins-git-notion-plugin-framer-app.vercel.app/notion
  - [ ] Authenticate with Notion and choose to use the template
  - [ ] Return to the Plugin and select the "Movies Examples" template
  - [ ] Continue and map the fields that you want to include
  - [ ] Press "Import"
  - [ ] it imports data in notion
- [ ] Synchronizing data
  - [ ] After initial setup add or remove an item from the notion database
  - [ ] Press "Synchronize" in the Framer CMS view
  - [ ] It removes the item you deleted from the Notion database
- [ ] Adding fields
 - [ ] Add a new field / column to the notion database
 - [ ] Press synchronize again
 - [ ] It now displays the field configuration again because a new field was added
 - [ ] Include this field in the data
 - [ ] Press "import"
 - [ ] Notice how the new field was added to each item


plugins/notion/src/MapFields.tsx
```diff
@@ -0,0 +1,284 @@
...
+                                <div
+                                    className={classNames(
+                                        "flex items-center justify-center",
+                                        isUnsupported && "opacity-50"
+                                    )}
+                                >
+                                    <IconChevron />
+                                </div>
+                                <input
+                                    type=""
```
anthonyec, line 239: Is `type=""` something special? Or is meant to be `type="text"`?

plugins/notion/src/blocksToHTML.ts
```diff
@@ -0,0 +1,106 @@
...
+            case "code":
+                htmlContent += `<pre><code>${richTextToHTML(block.code.rich_text)}</code></pre>`
+                break
+            default:
+                // TODO: More block types can be added here!
+                break
+        }
+    }
+
+    console.log("html", htmlContent)
```
anthonyec, line 103: Just making sure, keeping console logs on purpose as logs?

niekert: Ah, this one should not be there. I think the sync results are kind of useful to keep there for now.

anthonyec (approved):

PR #19, opened 2024-08-09
Add Doodles

benjamindenboer:
### Description
Adds the Doodles plugin.

image

### Testing

- [ ] Check dark and light modes
- [ ] Test basic insertion, history, and clearing
- [ ] Insert doodles of different widths and colors


plugins/doodles/readme.md
```diff
@@ -0,0 +1,7 @@
+# Color Extract
```
niekert, line 1: Could you update this readme? 🙏

benjamindenboer: Oopsie doopsie

plugins/doodles/src/App.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const canvasRef = useRef<ReactSketchCanvasRef>(null);
+  const strokeInputRef = useRef<HTMLInputElement>(null);
+  const strokeHueInputRef = useRef<HTMLInputElement>(null);
+
+  const [strokeValue, setStrokeValue] = useState(5);
+  const [strokeInputValue, setStrokeInputValue] = useState("5");
+
+  const [strokeHueInputValue, setStrokeHueInputValue] = useState("0");
+  const [strokeSaturateInputValue, setStrokeSaturateInputValue] = useState("0");
+  const [strokeLightInputValue, setStrokeLightInputValue] = useState("50");
```
niekert, line 56: Is there a reason to have separate states for the input and rendered values? They seem to always be in sync

benjamindenboer: I think so, because these inputs mimic the framer ux of allowing you to type in the input and it not insta setting the value until you hit enter. Otherwise values would update as you type?

niekert: ahh I see, that makes sense

niekert (approved):

PR #21, opened 2024-08-14
Update Notion and CSV export plugin for new Collection API

anthonyec:
image

### Description

- Add simple plugin to export any collection as a CSV
- Update Notion plugin to work with new `ManagedCollections`
- Update all example plugins to use latest version

### DIFF
--- a/plugins/csv-export/src/csv.ts
+++ b/plugins/csv-export/src/csv.ts
@@ -1,0 +1,126 @@
+import type { Collection, CollectionField, CollectionItem } from "framer-plugin"
+
+import { isFileAsset, isImageAsset } from "framer-plugin"
+import { assertNever } from "framer-plugin/utils"
+
+function downloadFile(file: File) {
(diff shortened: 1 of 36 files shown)


plugins/csv-export/src/csv.ts
```diff
@@ -0,0 +1,126 @@
+import type { Collection, CollectionField, CollectionItem } from "framer-plugin"
+
+import { isFileAsset, isImageAsset } from "framer-plugin"
+import { assertNever } from "framer-plugin/utils"
```
niekert, line 4: We are not exporting this from the package so you have to define it inside the plugin itself

anthonyec:
> image
> 
> UI looks a little off, maybe missing some shared CSS from devkit?
> 
> I also think the input field next to the button doesn't do anything?

Haha wow yeah. I need to get it working locally to fix this!

niekert (approved):
🚀

PR #26, opened 2024-08-29
Add Airtable plugin

Sakib25800:
### Description

This PR adds an Airtable plugin. It can be connected to your Airtable base and can sync a table from that base to the Framer CMS.

Backend: https://github.com/Sakib25800/framer-airtable-api

https://github.com/user-attachments/assets/5d5d055f-d442-446d-a30a-b4ae065595f2

### Scopes
- `data.records:read`
- `schema.bases:read`

### Testing

- [ ] Confirm the above example works
    - [ ] Clone the [🍿 Movies & TV Logs and Rating](https://airtable.com/appc7EdYpGlMTZfnb/shrcXmrFrvlaPGa2P) base 
  - [ ] Open the plugin 
  - [ ] Choose a table and map the fields to be included
  - [ ] Press "import"
  - [ ] Data should be imported
- [ ] Synchronizing 
    - [ ] After the above setup, remove a row from the synced Airtable table 
    - [ ] Press the "Sync" button in the Framer CMS view 
    - [ ] The item removed from Airtable should be removed from Framer
- [ ] Adding fields
    - [ ] Add a new field / column to the notion database
    - [ ] Press synchronize again
    - [ ] It now displays the field configuration again because a new field was added
    - [ ] Include this field in the data
    - [ ] Press "import"
    - [ ] Notice how the new field was added to each item
 - [ ] Empty field
    - [ ] Empty fields will not prevent the row from syncing
 - [ ] Auth
    - [ ] Attempting to sync the plugin, whilst not authenticated, will cause an error and allow the user to sign back in 
 - [ ] Pressing ctrl + shift + l should toggle logging


plugins/airtable/src/pages/MapTableFields.tsx
```diff
@@ -0,0 +1,250 @@
...
+        createFieldConfig(tableSchema.primaryFieldId, tableSchema.fields, pluginContext)
+    )
+    const [disabledFieldIds, setDisabledFieldIds] = useState(
+        () => new Set<string>(pluginContext.type === "update" ? pluginContext.ignoredFieldIds : [])
+    )
+    const [fieldNameOverrides, setFieldNameOverrides] = useState<Record<string, string>>(() =>
+        getFieldNameOverrides(pluginContext)
+    )
+
+    // TODO: Render progress in UI
```
Sakib25800, line 121: Similar to the Notion plugin, the progress UI can be added here.

PR #32, opened 2024-09-11
Add copy CSV to clipboard button

madebyisaacr:
### Description

This pull request adds a button to the Export CSV plugin that copies the CSV text to the clipboard, for cases where you don't need to download it as a file.

image

### DIFF
--- a/plugins/export-csv/src/csv.ts
+++ b/plugins/export-csv/src/csv.ts
@@ -102,7 +102,7 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
     return rows
 }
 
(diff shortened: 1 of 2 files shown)


plugins/export-csv/src/csv.ts
```diff
@@ -102,7 +102,7 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
     return rows
 }
 
-export async function exportCollectionAsCSV(collection: Collection, filename: string) {
+export async function convertCollectionToCSV(collection: Collection) {
```
anthonyec, line 105: 🙏

anthonyec (approved):

PR #54, opened 2024-09-19
Chore/assorted cleanups

ankon:
### Description

This pull request fixes two things I noticed when reading:
1. We're using npm here, so having a `.yarnrc.yml` is misleading and confusing, and might lead people to use yarn and then end up with odd package dependencies.
2. The google sheets plugin readme labeled the image as "Airtable"

### Testing

- [ ] Code Review

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,0 @@
-nodeLinker: node-modules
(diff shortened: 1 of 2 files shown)


huntercaron (approved):
Thank you!

PR #55, opened 2024-09-19
Add background remover plugin

Sakib25800:
Remove the background from an image (locally)

https://github.com/user-attachments/assets/77154537-2dc1-4f80-a545-db6ab60a63e5


plugins/background-remover/framer.json
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "ot052l",
+    "name": "Background Remover",
+    "modes": ["editImage"],
```
niekert, line 4: I think it could be nice to support this on `canvas` mode as well and apply it to the selection if a layer with an image is selected 🙏

Sakib25800: Agreed! Canvas mode has now been added, multi-image select is also supported.

PR #57, opened 2024-09-23
Improve HubSpot plugin

Sakib25800:
### Description

This PR does the following:
- Adds a subtle animation to the plugin page heading
- Makes all initial requests in the menu - if the user is on the `/forms`, `/tracking` or `/chat` the request will be made again
- Uses a `HubSpot Form` component to look cleaner 
- Skips the auth page if the user is already logged in 
- Changes the toggle tracking button from secondary / primary for enabled / disabled and does not use framer red
- Fixes dark mode 
- Fixes navigation for prod


plugins/hubspot/src/pages/Meetings.tsx
```diff
@@ -1,3 +1,4 @@
+import { useEffect } from "react"
```
anthonyec, line 1: I think this import is unused

anthonyec: (I'll fix this when I merge it in)

anthonyec (approved):

PR #60, opened 2024-09-23
Dither and ASCII plugins improvements

clementroche:
This PR improves both Dither and ASCII plugins:

Dither:
- add resolution picker
- reorganize gui
- new matrices
- CSS tweaks

ASCII:
- bug fixing
- maximum export resolution


niekert:
LGTM.

One thing: The Dither and Ascii plugin should only support `canvas`, `editImage` mode and not `image` mode. Will push a separate commit for that fix

PR #63, opened 2024-09-25
Prevent exporting CMS reference fields

jamesandres:
### Description

The current CMS export plugin makes a CSV where reference fields are included, but the values are all blank. The problem with this is the CSV import will set each item's reference field value to blank.

The basic solution to this is to adjust the CMS export plugin to ignore reference fields, ie: don't include them in the CSV export.

This is a stepping stone towards properly supporting references in CSV exports, see: https://github.com/framer/company/issues/30409

### Testing

- [ ] Be in a project with the collection references experiement enabled
- [ ] Create the sample collection
- [ ] Export each collection to CSV using this plugin
- [ ] The CSV data should not contain any columns for fields which are single or multi reference


jonastreub (approved):
Nice one 👏

PR #67, opened 2024-09-30
Upgrade plugins to latest

anthonyec:
### Description

- This is to use the new assigning Color Style changes

- [ ] Plugins are upgraded and work

### DIFF
--- a/plugins/cms-export/src/csv.ts
+++ b/plugins/cms-export/src/csv.ts
@@ -1,6 +1,6 @@
 import type { Collection, CollectionField, CollectionItem } from "framer-plugin"
 
-import { isFileAsset, isImageAsset } from "framer-plugin"
+import { isFileAsset, isImageAsset, isColorStyle } from "framer-plugin"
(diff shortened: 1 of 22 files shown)


plugins/cms-export/src/csv.ts
```diff
@@ -79,9 +79,23 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
...
+                        continue
+                    }
+
+                    if (isColorStyle(value)) {
+                        columns.push(value.light)
+                        continue
+                    }
+
+                    columns.push(`${value}`)
+                    continue
```
jonastreub, line 94: I think it would be simpler without the first early return:

```ts
if (isColorStyle(value)) {
  columns.push(value.light)
  continue
}

columns.push(typeof value === "string" ? value : "")
```

jonastreub: And maybe add a `isString` util

jonastreub (approved):
LGTM besides that one nit

PR #70, opened 2024-10-02
Notion without refresh button

jonastreub:
### Description

Removes the Refresh button of the Notion plugin. And instead automatically refreshes when the plugin window gets focus.

### Testing

- [ ] Notion plugin should automatically refresh every time you focus it (e.g. switching windows)

### DIFF
--- a/plugins/notion/src/SelectDatabase.tsx
+++ b/plugins/notion/src/SelectDatabase.tsx
@@ -12,6 +12,22 @@ export function SelectDatabase({ onDatabaseSelected }: SelectDatabaseProps) {
     const { data, refetch, isRefetching, isLoading } = useDatabasesQuery()
     const [selectedDatabase, setSelectedDatabase] = useState<string | null>(null)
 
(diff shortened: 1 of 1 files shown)


plugins/notion/src/SelectDatabase.tsx
```diff
@@ -55,39 +73,26 @@ export function SelectDatabase({ onDatabaseSelected }: SelectDatabaseProps) {
...
-                        onChange={e => setSelectedDatabase(e.target.value)}
+                        onChange={event => setSelectedDatabase(event.target.value)}
                         className="flex-1 shrink-1"
-                        disabled={data && data.length === 0}
+                        disabled={!selectEnabled}
                     >
-                        {isLoading && (
-                            <option disabled selected>
-                                Loading…
-                            </option>
```
jonastreub, line 65: I saw warnings in the console that you should not make select options selected like this. It already automatically gets selected so not a problem I think

anthonyec (approved):

PR #76, opened 2024-10-03
Select sheet using URL instead of dropdown

anthonyec:
### Description

Improvements to the Sheets Plugin:

- Allow entering a Sheets URL so we do not need access to the Drive permission and hopefully won't need an audit
- If you don't have access to the Sheet you are trying to sync you will see a screen that explains this
- When auth tokens are missing and you press "sync" you have to re-authenticate. After authentication it will sync immediately instead of displaying field configuration
- UX polish

### Testing

- [ ] The flow works with several users
  - [ ] Create a spreadsheet with some headers
  - [ ] Open the Plugin and authenticate
  - [ ] Enter the spreadsheet URL
  - [ ] Select the Sheet you want to sync
  - [ ] Configure fields and sync
  - [ ] It works
  - [ ] Update some fields in the sheet
  - [ ] Press sync
  - [ ] It syncs
  - [ ] Fill in only the slug column in the sheet
  - [ ] Press sync again
  - [ ] It syncs with a partially completed row
- [ ] It displays a screen when you don't have access
  - [ ] Now attempt to synchronize the sheet with another Google account in another browser profile
  - [ ] Notice you see a screen that you do not have access
  - [ ] Press "View Sheet" and request permissions
  - [ ] Grant the permissions
  - [ ] Press retry in the plugin
  - [ ] It syncs immediately now


plugins/google-sheets/src/pages/Authenticate.tsx
```diff
@@ -56,12 +65,25 @@ export function Authenticate({ onAuthenticated }: AuthenticationProps) {
     return (
         <div className="col-lg pb-[15px]">
             <Hero />
-            <ol className="list-decimal list-inside space-y-2.5 marker:text-secondary *:text-tertiary *:leading-none *:tracking-normal py-[7px]">
-                <li>Log in to your Google account</li>
-                <li>Ensure your sheet has a header row</li>
-                <li>Map the column fields to the CMS</li>
+            <ol className="list-decimal list-inside space-y-2.5 marker:text-secondary *:text-content *:leading-none *:tracking-normal py-[7px]">
+                <li>
+                    <span className="pl-[10px]">Log in to your Google account</span>
```
niekert, line 70: I can't believe its this annoying to style the list marker but unfortunately with `    list-style-position: inside;` you don't have control over the padding of the text after the marker.

And using outside would require some hardcoded value which could differ between browsers.

https://css-tricks.com/everything-you-need-to-know-about-the-gap-after-the-list-marker/

plugins/google-sheets/src/components/CheckboxTextField.tsx
```diff
@@ -27,7 +29,7 @@ export function CheckboxTextfield({ value, darken, checked, onChange }: Props) {
                 checked={checked}
                 onChange={handleCheckboxChange}
                 onClick={e => e.stopPropagation()}
-                className="checked:!bg-sheets-green focus:ring-1 focus:ring-sheets-green"
+                className="checked:!bg-sheets-green focus:ring-1 focus:ring-sheets-green checked:border-none dark:!bg-[#777] !bg-[#CCC]"
```
niekert, line 32: These colors are different because we render this checkbox on a non standard background.

plugins/google-sheets/src/main.tsx
```diff
@@ -21,23 +19,10 @@ export const queryClient = new QueryClient({
...
-    if (pluginContext.hasChangedFields) return false
-
-    return true
-}
-
 function renderPlugin(app: ReactNode) {
     const root = document.getElementById("root")
     if (!root) throw new Error("Root element not found")
 
-    framer.showUI({
```
niekert, line 36: Moving this inside react components because synchronizing is now also done in the `<App />` component. The reason is this flow:

1. Press sync
2. You are not authenticated
3. You authenticate and we call `getContext` again
4. It now synchronizes without UI as expected.

anthonyec (commented):
Can't press approve because I created the PR, but looks good to me! 👍

PR #77, opened 2024-10-03
Add support for references to cms export

tom-james-watson:
### Description

This PR adds support for collection references to CMS Export.

This depends on https://github.com/framer/FramerStudio/pull/19215, which adds support for references to plugins.

### Testing

- [ ] See https://github.com/framer/FramerStudio/pull/19215

### DIFF
--- a/plugins/cms-export/src/csv.ts
+++ b/plugins/cms-export/src/csv.ts
@@ -59,6 +59,7 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
                     continue
                 }
 
+                case "collectionReference":
                 case "formattedText": {
                     if (typeof value !== "string") {
                         columns.push("")
@@ -66,6 +67,16 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
                     }
 
                     columns.push(`${value}`)
+                    continue
+                }
+
+                case "multiCollectionReference": {
+                    if (!Array.isArray(value) || value.some(v => typeof v !== "string")) {
+                        columns.push("")
+                        continue
+                    }
+
+                    columns.push(`${value.join(",")}`)
                     continue
                 }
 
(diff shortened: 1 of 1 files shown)


plugins/cms-export/src/csv.ts
jonastreub, line 1: I wonder if users should have the option to setup references by `slug`. Do we currently even include ids in CSV exports? If not, there would be no way to manage references externally

tom-james-watson: We don't export ids at the moment, no. I think the use case here is downloading, applying some edits, and reuploading, which is what I've seen some customers wanting to do in the past. I don't think we necessarily need to support full offline managing of references for this to be useful.

jonastreub: But why would we include references if there is zero ways to know what these ids mean?

tom-james-watson: I guess we could just always export ids, and then require that optionally allow ids when importing

plugins/cms-export/src/csv.ts
```diff
@@ -69,6 +70,16 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
                     continue
                 }
 
+                case "multiCollectionReference": {
+                    if (!Array.isArray(value) || value.some(v => typeof v !== "string")) {
+                        columns.push("")
+                        continue
+                    }
+
+                    columns.push(`${value.join(",")}`)
```
jonastreub, line 79: Can we add an extra space after the comma?

tom-james-watson: I think that would be incorrect. Standard way to add an array of strings to a csv would be to quote and comma-join.

jonastreub: Okay. I didn't know 👌

PR #81, opened 2024-10-07
Support images in Notion and general improvements

niekert:
### Description

- Supports Images and Files when a Notion database has a "Files and Media" column
- Allows you to switch between File or Image for that column
- Will pick only the first item in the media column. CMS arrays alter.
- Allow switching between Rich Text and Regular text for content fields
- Fix a bug where Notion would paginate endlessly because the next cursor would match the current cursor.
- Display errors as toasts instead of inline in the UI

### Testing

- [ ] Create a Notion database
  - [ ] Add a column "Files and Media"
  - [ ] Upload some images to the database and fill in other columns
  - [ ] Launch Notion & select the database
  - [ ] Notice you can select "File" or "Image" for the media column
  - [ ] It synchronizes with the selected file type

### DIFF
--- a/plugins/notion/src/App.tsx
+++ b/plugins/notion/src/App.tsx
@@ -22,6 +22,9 @@ export function AuthenticatedApp({ context }: AppProps) {
         framer.showUI({
             width: databaseConfig ? 360 : 325,
             height: databaseConfig ? 425 : 370,
+            // Only allow resizing when mapping fields as the default size could not be enough.
+            // This will keep the given dimensions in the Splash Screen.
+            resizable: Boolean(databaseConfig),
         })
     }, [databaseConfig])
 
--- a/plugins/notion/src/App.tsx
+++ b/plugins/notion/src/App.tsx
@@ -19,9 +19,13 @@ export function AuthenticatedApp({ context }: AppProps) {
     )
 
     useEffect(() => {
+        const width = databaseConfig ? 360 : 325
+        const height = databaseConfig ? 425 : 370
         framer.showUI({
-            width: databaseConfig ? 360 : 325,
-            height: databaseConfig ? 425 : 370,
+            width,
+            height,
+            minWidth: width,
+            minHeight: height,
             // Only allow resizing when mapping fields as the default size could not be enough.
             // This will keep the given dimensions in the Splash Screen.
             resizable: Boolean(databaseConfig),
--- a/plugins/notion/src/App.tsx
+++ b/plugins/notion/src/App.tsx
@@ -20,7 +20,7 @@ export function AuthenticatedApp({ context }: AppProps) {
 
     useEffect(() => {
         framer.showUI({
-            width: databaseConfig ? 450 : 325,
+            width: databaseConfig ? 360 : 325,
             height: databaseConfig ? 425 : 370,
         })
     }, [databaseConfig])
@@ -30,7 +30,7 @@ export function AuthenticatedApp({ context }: AppProps) {
             logSyncResult(result)
 
             if (result.status === "success") {
-                // framer.closePlugin("Synchronization successful")
+                framer.closePlugin("Synchronization successful")
                 return
             }
         },
--- a/plugins/notion/src/App.tsx
+++ b/plugins/notion/src/App.tsx
@@ -20,7 +20,7 @@ export function AuthenticatedApp({ context }: AppProps) {
 
     useEffect(() => {
         framer.showUI({
-            width: databaseConfig ? 340 : 325,
+            width: databaseConfig ? 450 : 325,
             height: databaseConfig ? 425 : 370,
         })
     }, [databaseConfig])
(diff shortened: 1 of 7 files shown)


plugins/notion/src/MapFields.tsx
```diff
@@ -117,14 +130,21 @@ export function MapDatabaseFields({
...
         () => new Set<string>(pluginContext.type === "update" ? pluginContext.ignoredFieldIds : [])
     )
+    const [fieldTypeByFieldId, setFieldTypeByFieldId] = useState(() =>
+        getInitialFieldTypeState(database, pluginContext)
+    )
     const [fieldNameOverrides, setFieldNameOverrides] = useState<Record<string, string>>(() =>
         getFieldNameOverrides(pluginContext)
     )
 
+    // TODO: Render progress in UI.
```
niekert, line 144: Adding this in a followup PR because there's no UI for it yet.

plugins/notion/src/components/CheckboxTexfield.tsx
```diff
@@ -12,22 +12,33 @@ export function CheckboxTextfield({ value, disabled, checked: initialChecked, on
...
+        if (disabled) return
 
-      setChecked(!checked)
-      onChange()
+        setChecked(!checked)
+        onChange()
     }
 
     return (
-        <div className={classNames("flex bg-tertiary rounded-[8px] items-center pl-[10px]", disabled && "opacity-50 cursor-default")} onClick={toggle} role="button">
```
niekert, line 22: This feature was broken. The `onClick` on the div causes the checkbox to be toggled twice immediately. Turning the div in a label ensures that the checkbox is toggled when the whole thing is clicked.

plugins/notion/src/App.tsx
```diff
@@ -20,12 +20,18 @@ export function AuthenticatedApp({ context }: AppProps) {
...
             height: databaseConfig ? 425 : 370,
+            // Only allow resizing when mapping fields as the default size could not be enough.
+            // This will keep the given dimensions in the Splash Screen.
+            resizable: Boolean(databaseConfig),
         })
     }, [databaseConfig])
 
     const synchronizeMutation = useSynchronizeDatabaseMutation(databaseConfig, {
+        onError(error) {
+            framer.notify(error.message, { variant: "error" })
```
niekert, line 33: Instead of displaying an error in the field mapping UI we display a toast.

This seems way better because the error text was getting cut off. Sheets Plugin does the same. Probably we built this before we had such an API 😄

plugins/notion/src/components/CheckboxTexfield.tsx
```diff
@@ -12,22 +12,33 @@ export function CheckboxTextfield({ value, disabled, checked: initialChecked, on
...
+        <label
```
niekert, line 22: This feature was broken. The onClick on the div causes the checkbox to be toggled twice immediately. Turning the div in a label ensures that the checkbox is toggled when the whole thing is clicked.

plugins/notion/src/notion.ts
```diff
@@ -371,15 +508,14 @@ async function processItem(
                 fieldId: field.id,
                 message: `Value is missing for field ${field.name}`,
             })
-            continue
```
niekert, line 374: Removing this on purpose. A single field not having a value should not exclude it entirely from synchronizing.

plugins/notion/src/notion.ts
```diff
@@ -679,3 +838,51 @@ export function isUnchangedSinceLastSync(lastEditedTime: string, lastSyncedTime:
...
+interface PaginatedList<T> {
+    object: "list"
+    results: T[]
+    next_cursor: string | null
+    has_more: boolean
+}
+
+/**
+ * Copied from:
+ * https://github.com/makenotion/notion-sdk-js/blob/main/src/helpers.ts#L47
```
alecmev, line 855: Some anti-link-rot 😛 

```suggestion
 * https://github.com/makenotion/notion-sdk-js/blob/7950edc034d3007b0612b80d3f424baef89746d9/src/helpers.ts#L47
```

alecmev (commented):
Will have time to test only on Friday, sorry!

plugins/notion/src/MapFields.tsx
```diff
@@ -1,67 +1,65 @@
...
+function getSortedProperties(database: GetDatabaseResponse): NotionProperty[] {
+    return getNotionProperties(database).sort((propertyA, propertyB) => {
+        // Properties that are not supported in the Plugin are displayed at the bottom of the list.
+        if (isSupportedNotionProperty(propertyA) && isSupportedNotionProperty(propertyB)) {
+            return -1
+        } else if (isSupportedNotionProperty(propertyA)) {
+            return -1
+        } else if (!isSupportedNotionProperty(propertyB)) {
+            return 1
+        }
```
jonastreub, line 34: I might have flipped the logic here but isn't this the same?
```ts
const a = isSupportedNotionProperty(propertyA) ? 1 : 0
const b = isSupportedNotionProperty(propertyB) ? 1 : 0
return a - b
```

plugins/notion/src/notion.ts
niekert: Thanks!

plugins/notion/src/MapFields.tsx
niekert: Yep, works for me 👌

huntercaron:
Works good in my testing ✅

PR #83, opened 2024-10-10
Prevent infinite rerender of CMS Export plugin

tom-james-watson:
### Description

Prevents an infinite rerender that was caused by this `useEffect` having an unnecessary dependency on `previewCSV`, which itself is being rewritten every time the hook runs.

### Testing

- [ ] Plugin still works like before


niekert (approved):

tom-james-watson:
Not sure what the process is here - do I just merge and that's it?

niekert:
yep feel free to merge. For first party plugins we submit them to the marketplace by hand after merge. I'm happy to do it

PR #94, opened 2024-10-31
Add support to notion for relations via multi references

tom-james-watson:
### Description

This PR adds support for [notion relations](https://www.notion.so/help/relations-and-rollups).

This depends on https://github.com/framer/FramerStudio/pull/19215, which adds support for references to plugins.

### Testing

- [ ] See https://github.com/framer/FramerStudio/pull/19215


tom-james-watson:
It was pretty simple in the end honestly, so I don't think so.

PR #98, opened 2024-11-05
Reuse refresh token when Google authentication expires

anthonyec:
### Description

Turns out Google refresh tokens [never expire unless](https://stackoverflow.com/a/8954103) the user invokes access. So if a new refresh token is not provided, reuse the existing one stored in local storage.

### Testing

Unless you want to wait 2 hours, 1 hour for the initial expiry, another for the refresh, clone the code and edit the [`expiresIn`](https://github.com/framer/plugins/blob/0e610277c229ea7cf353f740ac12576392902c32/plugins/google-sheets/src/auth.ts#L127) to be 10 seconds.

- [ ] Login with Google and wait for token to expire. Reopen the plugin multiple times and ensure you are still loggedin
- [ ] Ensure the login flow is triggered when the refresh token has been lost (delete it from local storage)

### DIFF
--- a/plugins/google-sheets/src/auth.ts
+++ b/plugins/google-sheets/src/auth.ts
@@ -8,10 +8,11 @@ interface Tokens {
 
 interface StoredTokens {
     createdAt: number
+    /** Duration in seconds. */
     expiredIn: number
     accessToken: string
-    // When you have previously logged into the app with oAuth Google no longer gives a refresh token.
-    // https://stackoverflow.com/a/10857806
+    // After the initial login, Google does not provide new refresh tokens since
+    // the original does not expire. https://stackoverflow.com/a/8954103
     refreshToken: string | undefined
 }
 
@@ -48,7 +49,6 @@ class Auth {
             })
 
             const newTokens = (await res.json()) as Tokens
-
             this.tokens.save(newTokens)
 
             return newTokens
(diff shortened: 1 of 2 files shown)


plugins/google-sheets/src/auth.ts
```diff
@@ -8,10 +8,11 @@ interface Tokens {
 
 interface StoredTokens {
     createdAt: number
+    /** Duration in seconds. */
     expiredIn: number
     accessToken: string
-    // When you have previously logged into the app with oAuth Google no longer gives a refresh token.
-    // https://stackoverflow.com/a/10857806
+    // After the initial login, Google does not provide new refresh tokens since
+    // the original does not expire. https://stackoverflow.com/a/8954103
```
niekert, line 15: I think that because we now pass `prompt=consent` this is no longer the case and the value should technically always be there?

anthonyec: 😱

anthonyec: Let me try

niekert: Sorry I may be wrong, that would be the `Tokens` interface which already has that

anthonyec: Just checked, yeah, it doesn't provide it on subsequent refreshes. We still need this

niekert (approved):
Nice one 👏  Also did QA steps and looks good!

PR #101, opened 2024-11-07
Use sheet ID when its available to reference a sheet

anthonyec:
### Description

This pull request …

### Testing

  - [ ] Step 1
  - [ ] Step 2
  - [ ] Step 3
  - [ ] Step 1
  - [ ] Step 2
  - [ ] Step 3


plugins/google-sheets/src/pages/SelectSheet.tsx
```diff
@@ -23,17 +24,22 @@ export function SelectSheetPage({ onError, onSheetSelected }: Props) {
         }
     }, [isSpreadSheetInfoError, onError])
 
-    const handleSheetSelect = (e: SelectChangeEvent) => {
-        setSelectedSheetId(e.target.value)
+    const handleSheetSelect = (event: SelectChangeEvent) => {
+        const id = parseInt(event.currentTarget.value)
```
alecmev, line 28: Leave as a string maybe? Especially considering that later it gets stringified a few times anyway? Who knows if these numbers actually fit into 53 bits or not.

anthonyec: Hmm yeah feels a bit messy, let me try change that

plugins/google-sheets/src/sheets.ts
```diff
@@ -379,7 +388,11 @@ export async function syncSheet({
 
     const unsyncedItemIds = new Set(await collection.getItemIds())
 
-    const sheet = fetchedSheet ?? (await fetchSheetWithClient(spreadsheetId, sheetTitle))
+    const spreadsheetInfo = await fetchSpreadsheetInfo(spreadsheetId)
+    const sheetTitleFromId = spreadsheetInfo.sheets.find((sheet) => sheetId === sheet.properties.sheetId)?.properties.title
+    const sheetIdFromTitle = spreadsheetInfo.sheets.find((sheet) => sheetTitle === sheet.properties.title)?.properties.sheetId
+
+    const sheet = fetchedSheet ?? (await fetchSheetWithClient(spreadsheetId, sheetTitleFromId ?? sheetTitle))
```
alecmev, line 395: It has to be fetched via title, right? Because their API doesn't accept id?

https://github.com/framer/plugins/blob/85026c400997d879e12dab97294a1af783db9720/plugins/google-sheets/src/sheets.ts#L147

anthonyec: There may be another API to just get a whole sheets values not using a range, but I don't wanna introduce too much change this late

plugins/google-sheets/src/App.tsx
```diff
@@ -47,9 +47,13 @@ export function AuthenticatedApp({ pluginContext, setContext }: AuthenticatedApp
     const [spreadsheetId, setSpreadsheetId] = useState<string | null>(
         pluginContext.type === "update" ? pluginContext.spreadsheetId : null
     )
+    const [sheetId, setSheetId] = useState<number | null>(
+        pluginContext.type === "update" ? pluginContext.sheetId : null
+    )
     const [sheetTitle, setSheetTitle] = useState<string | null>(
         pluginContext.type === "update" ? pluginContext.sheetTitle : null
     )
+
```
alecmev, line 56: Need to do here the same thing as in `syncSheet` and use `sheetTitleFromId` by default, and fall back to `sheetTitle` only if that one is `null`. This is the source of the bug where "Manage" doesn't work on an "upgraded" collection with a renamed sheet.

plugins/google-sheets/src/pages/SelectSheet.tsx
alecmev: For future readers, we reverted it because Google does return these as numbers, so it's fine.

alecmev:
Superseded by https://github.com/framer/plugins/pull/102.

PR #102, opened 2024-11-07
Fix Google Sheets not picking up sheet renames

alecmev:
Simplified version of https://github.com/framer/plugins/pull/101, same QA steps.

### QA

- [ ] Create a collection using the stable plugin.
- [ ] Confirm that renaming the sheet breaks "Sync".
- [ ] Confirm that renaming the sheet back fixes "Sync".
- [ ] Add `plugin=5i8ra6xs08ydg7wbx4nhndbn1&pluginVersion=jlwb6k2pqovdmb4cclqu5554k3s5sqfh` to the query.
- [ ] Dismiss the "create collection" dialog.
- [ ] Rename the sheet again.
- [ ] Observe that the error dialog now offers some advice.
- [ ] Follow the advice.
- [ ] Observe that it fixes "Sync".
- [ ] Rename the sheet again and observe that "Sync" still works.
- [ ] Delete the collection and go through the same steps for "Manage" instead of "Sync".


plugins/google-sheets/src/pages/Problem.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { framer } from "framer-plugin"
+import { Button } from "../components/Button"
+
+interface Props extends PropsWithChildren {
+    height: number
+    spreadsheetId: string
+    setContext: (newContext: PluginContext) => void
+}
+
+export function Problem({ height, spreadsheetId, setContext, children }: Props) {
```
anthonyec, line 12: Great component name

plugins/google-sheets/src/sheets.ts
```diff
@@ -435,9 +450,10 @@ export async function getPluginContext(): Promise<PluginContext> {
     const isAuthenticated = !!tokens
 
     const spreadsheetId = await collection.getPluginData(PLUGIN_SPREADSHEET_ID_KEY)
-    const sheetTitle = await collection.getPluginData(PLUGIN_SHEET_TITLE_KEY)
+    const storedSheetId = await collection.getPluginData(PLUGIN_SHEET_ID_KEY)
+    const storedSheetTitle = await collection.getPluginData(DO_NOT_USE_ME_PLUGIN_SHEET_TITLE_KEY)
 
-    if (!spreadsheetId || !sheetTitle || !isAuthenticated) {
+    if (!spreadsheetId || (storedSheetTitle === null && storedSheetId === null) || !isAuthenticated) {
```
anthonyec, line 456: In the [other PR](https://github.com/framer/plugins/pull/101/files#diff-cdf12cd4b95b9d54071b04babfa33c04c4fb8c29de9cdf475d8065858adcdf59R455-R458), it only showed the new sheet modal when the title was missing but not the ID. This is because the ID might not exist for people who haven't run the new version of the plugin. I may have missed something though

plugins/google-sheets/src/pages/Problem.tsx
alecmev: Thanks 😄 Didn't want to overthink it.

plugins/google-sheets/src/sheets.ts
alecmev: Yep, this is on purpose! Possible states:

1. Both are unset: That means there's no sheet to look up, no choice but ask for a new one.
2. Only ID is set: It's ok, that means we already migrated this.
3. Only title is set: Also ok, that means it's yet to be migrated.
4. Both are set: Shouldn't happen, as we clear the title, but it's fine, we just ignore it if ID is set.

Do you see a potential problem here?

anthonyec: Nah, that sounds good!

anthonyec (approved):

PR #103, opened 2024-11-12
Fix Google Sheets link import

triozer:
### Description

This PR fixes an issue with Google Sheets link import functionality.

### Testing

- [ ] Create a test spreadsheet containing:
  - Column for ID
  - Column for Links (URLs)
- [ ] Import the spreadsheet using the plugin
- [ ] Verify that:
  - Links are properly detected and imported
  - Error handling works as expected for invalid URLs

### DIFF
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -68,16 +68,18 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
     if (typeof cellValue === "number") return "number"
 
     if (typeof cellValue === "string") {
-        if (/^\d{1,2}\/\d{1,2}\/\d{4}/.test(cellValue)) return "date"
+        const cellValueLowered = cellValue.toLowerCase()
 
-        if (/^#[0-9A-Fa-f]{6}$/.test(cellValue)) return "color"
+        if (/^\d{1,2}\/\d{1,2}\/\d{4}/.test(cellValueLowered)) return "date"
 
-        if (/<[a-z][\s\S]*>/i.test(cellValue)) return "formattedText"
+        if (/^#[0-9a-f]{6}$/.test(cellValueLowered)) return "color"
+
+        if (/<[a-z][\s\S]*>/i.test(cellValueLowered)) return "formattedText"
 
         try {
-            new URL(cellValue)
+            new URL(cellValueLowered)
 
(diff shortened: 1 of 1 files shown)


plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -72,11 +72,17 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
 
         if (/^#[0-9A-Fa-f]{6}$/.test(cellValue)) return "color"
 
-        if (/\.(jpeg|jpg|gif|png)$/.test(cellValue)) return "image"
+        if (/<[a-z][\s\S]*>/i.test(cellValue)) return "formattedText"
 
-        if (/^(https?:\/\/)?([a-zA-Z0-9.-]+)\.([a-zA-Z]{2,6})([/\w .-]*)*\/?$/.test(cellValue)) return "link"
+        try {
+            new URL(cellValue)
```
alecmev, line 78: If we're okay with Baseline 2023 features, then maybe [`URL.canParse`](https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static)?

alecmev: Safari 16.6 doesn't support it, unfortunately.

plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -72,11 +72,17 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
...
 
-        if (/<[a-z][\s\S]*>/i.test(cellValue)) return "formattedText"
+            if (/\.(jpeg|jpg|gif|png)$/.test(cellValue)) return "image"
```
alecmev, line 80: Can you, please, update this to also include all supported file types:

```
["gif", "image/gif"],
["jpg", "image/jpeg"],
["jpeg", "image/jpeg"],
["png", "image/png"],
["apng", "image/png"],
["svg", "image/svg+xml"],
["webp", "image/webp"],
```

And make it case-insensitive? Right now it rejects `foo.JPG`, for example. Might be cleaner to just do a `const cellValueLower = cellValue.toLowerCase()` at the beginning of the block and update the regular expressions to use that, without any upper ranges like `A-F` or `i` flags.

triozer: Yes, that's indeed cleaner! Thanks for flagging 😁

alecmev (approved):
Let's go! 🎉

PR #104, opened 2024-11-12
CMS Export: select active collection automatically

jonastreub:
### Description

When opening CMS Export we now automatically select the active collection.

Also updated all plugins to the latest version. And the lock file has lost a lot of "unnecessary" lines. Please check if that's okay.

### Testing

- [ ] CMS Export should automatically select the active collection
  - [ ] Add the sample collection
  - [ ] Open the CMS Export plugin
  - [ ] The active collection should be visible
  - [ ] Go to the canvas
  - [ ] The plugin should open without selecting any collections

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5455,6 +5455,15 @@
                 }
             }
         },
+        "node_modules/framer-plugin": {
+            "version": "2.0.2",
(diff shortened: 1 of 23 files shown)


anthonyec (approved):

anthonyec:
Tested it and looks good!

PR #109, opened 2024-11-13
Improve Google Sheets header syncing

triozer:
### Description

Update how we map Google Sheets header to the plugin. Previously we were using number-based Indexes and that would lead to shift when adding / moving / removing columns in the spreadsheet and then syncing it in Framer. 

### Testing

- [ ] Should support column operations
  - [ ] Import a spreadsheet
  - [ ] Edit/add/remove/move a column anywhere in the spreadsheet
  - [ ] Sync
  - [ ] You should be prompted to re-configure the collection
- [ ] Should be backward compatible
  - [ ] Checkout main
  - [ ] Import a spreadsheet and change the slug, the field name and ignore some fields
  - [ ] Checkout `fix/sheets-stable-header`
  - [ ] Click on either sync or manage
  - [ ] You should be prompted to re-configure the collection with pre-filled slug, field name and ignored field

### DIFF
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -2,19 +2,12 @@ import { Fragment, useMemo, useState } from "react"
 import { CollectionField, ManagedCollectionField } from "framer-plugin"
 import { useInView } from "react-intersection-observer"
 import cx from "classnames"
-import {
-    CellValue,
-    CollectionFieldType,
-    hasFieldConfigurationChanged,
-    HeaderRow,
-    PluginContext,
-    Row,
-    SyncMutationOptions,
-} from "../sheets"
+import { CellValue, CollectionFieldType, HeaderRow, PluginContext, Row, SyncMutationOptions } from "../sheets"
 
 import { IconChevron } from "../components/Icons"
 import { Button } from "../components/Button"
 import { CheckboxTextfield } from "../components/CheckboxTextField"
+import { generateUniqueNames } from "../utils"
 
 interface CollectionFieldConfig {
     field: ManagedCollectionField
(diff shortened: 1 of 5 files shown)


plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -148,28 +146,50 @@ export function MapSheetFieldsPage({
 }: Props) {
     const { ref: scrollRef, inView: isAtBottom } = useInView({ threshold: 1 })
 
+    const uniqueColumnNames = useMemo(() => generateUniqueNames(headerRow), [headerRow])
```
triozer, line 149: Maybe this could be stored in `pluginContext`?

plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -90,26 +79,35 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
...
+    return headerRow.map((headerName, columnIndex) => {
+        const sanitizedName = uniqueColumnNames[columnIndex]
+
+        return {
+            field: {
+                id: sanitizedName,
+                name: sanitizedName,
+                type: getFieldType(context, sanitizedName, row?.[columnIndex]),
+            } as ManagedCollectionField,
+            originalFieldName: headerName,
```
triozer, line 108: This is not used anymore.

plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -148,28 +146,50 @@ export function MapSheetFieldsPage({
...
+                // If the disabled column is the slug column, we need to update it to the next
+                // possible slug field
+                if (id === slugColumn) {
+                    const nextSlugField = getPossibleSlugFields(fieldConfig).find(
+                        field => field.id !== id && !nextSet.has(field.id)
+                    )
+                    if (nextSlugField) {
+                        setSlugColumn(nextSlugField.id)
+                    }
+                }
```
triozer, line 192: Quick UX win

plugins/google-sheets/src/sheets.ts
```diff
@@ -515,19 +534,56 @@ export async function getPluginContext(): Promise<PluginContext> {
     // turns out to be false too often - switch to batchGetByDataFilter +
     // GridRange and drop titles altogether.
 
+    const sheetHeaderRow = sheet.values[0]
+    const currentSheetHeaderRowHash = generateHashId(sheetHeaderRow.join(","))
```
alecmev, line 538: Hmm, re-ordering columns shouldn't trigger a re-sync. To avoid this, I think this needs to be sorted before hashing?

Also, maybe use a spicier delimiter, that's unlikely to occur in source data. Could just do a constant random string, e.g., `ErSW0wuIm7KM90LF`, since we aren't length-limited.

`generateHashId` is also a poor checksum hash, but it's out of scope of this PR 😄 Just noting it for awareness.

plugins/google-sheets/src/pages/MapSheetFields.tsx
alecmev: Hmm, why was it there in the first place? Can drop it, probably.

plugins/google-sheets/src/pages/MapSheetFields.tsx
alecmev: Nah, can't, due to https://github.com/framer/company/issues/30923. This PR actually fixes that.

plugins/google-sheets/src/sheets.ts
alecmev: > re-ordering columns shouldn't trigger a re-sync

Or, maybe it's fine! It's not a huge price to pay, and column re-ordering doesn't happen every day.

alecmev (approved):
LGTM! 👌

huntercaron:
also LGTM!

PR #112, opened 2024-11-15
Switch assertNever for shouldBeNever

tom-james-watson:
We don't want to throw if an unexpected type is found here - we only want to show a typescript warning. Prevents errors when the plugin API changes to add a new field, for example.

Depends on https://github.com/framer/FramerStudio/pull/19842.

### DIFF
--- a/plugins/cms-export/src/assert.ts
+++ b/plugins/cms-export/src/assert.ts
@@ -1,3 +1,8 @@
-export function assertNever(x: never, error?: unknown): never {
-  throw error || new Error((x as unknown) ? `Unexpected value: ${x}` : "Application entered invalid state")
-}
+/**
+ * A utility function that does nothing but makes TypeScript check for the never type.
+ *
+ * For example, sometimes something that should never happen is expected to
+ * happen, like during a rollback. To prevent unwanted crashers use
+ * `shouldBeNever` instead of `assertNever`.
+ */
+export function shouldBeNever(_: never) {}
--- a/plugins/cms-export/src/csv.ts
+++ b/plugins/cms-export/src/csv.ts
@@ -1,7 +1,7 @@
 import type { Collection, CollectionField, CollectionItem } from "framer-plugin"
 
-import { isFileAsset, isImageAsset, isColorStyle } from "framer-plugin"
-import { assertNever } from "./assert"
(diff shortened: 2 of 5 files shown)


plugins/cms-export/src/csv.ts
```diff
@@ -110,7 +110,7 @@ export function getDataForCSV(fields: CollectionField[], items: CollectionItem[]
                 }
 
                 default: {
-                    assertNever(field)
+                    shouldBeNever(field)
```
jonastreub, line 113: Should we do `columns.push("")` here? It seems like we have still added this to the header.

jonastreub:
Would be great to land this change. Needs a couple more tweaks

plugins/notion/package.json
```diff
@@ -14,7 +14,7 @@
         "@notionhq/client": "^2.2.15",
         "@tanstack/react-query": "^5.29.2",
         "classnames": "^2.5.1",
-        "framer-plugin": "^2.0.2",
+        "framer-plugin": "^2.0.4",
```
tom-james-watson, line 17: We need to `npm i` once this is published to update the lockfile.

tom-james-watson:
Sorry about the delay here! I've refactored this to avoid processing unsupported/unknown fields entirely.

plugins/cms-export/src/assert.ts
```diff
@@ -1,3 +1,8 @@
-export function assertNever(x: never, error?: unknown): never {
-  throw error || new Error((x as unknown) ? `Unexpected value: ${x}` : "Application entered invalid state")
-}
+/**
+ * A utility function that does nothing but makes TypeScript check for the never type.
+ *
+ * For example, sometimes something that should never happen is expected to
+ * happen, like during a rollback. To prevent unwanted crashers use
+ * `shouldBeNever` instead of `assertNever`.
```
jonastreub, line 6: In the context of this plugin its not about rollbacks but more about being forward compatible when union types are being extended

jonastreub (approved):
LGTM 👏

plugins/cms-export/src/assert.ts
tom-james-watson: Yeah, that message just got copy-pasted. I'll improve it 👌

PR #115, opened 2024-11-21
Prepare Airtable plugin for Release

triozer:
Polish and update the Airtable plugin.

### Todo

- [x] Auto-select first table
- [x] CMS Field
  - [x] Add support for collection reference
  - [x] Add support for single collection reference
  - [x] Add support for multiple collection references
- [x] Better data storage
  - [x] Only store ignored fields hash

### DIFF
--- a/plugins/airtable/src/airtable.ts
+++ b/plugins/airtable/src/airtable.ts
@@ -534,11 +534,15 @@ function getIgnoredFieldIds(rawIgnoredFieldIds: string | null) {
         return []
     }
 
-    const parsed = JSON.parse(rawIgnoredFieldIds)
-    if (!Array.isArray(parsed)) return []
-    if (!parsed.every(val => typeof val === "string")) return []
+    try {
+        const parsed = JSON.parse(rawIgnoredFieldIds)
+        if (!Array.isArray(parsed)) return []
+        if (!parsed.every(val => typeof val === "string")) return []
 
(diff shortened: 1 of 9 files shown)


triozer:
Beta version was released at this point.

plugins/airtable/src/airtable.ts
```diff
@@ -412,71 +529,36 @@ export async function syncTable({
...
-
-    for (const fieldSchema of tableSchema.fields) {
-        if (ignoredFieldIds.includes(fieldSchema.id)) continue
-
-        const field = getCollectionForAirtableField(fieldSchema)
-        if (field) {
-            fields.push(field)
-        }
-    }
+    const parsed = JSON.parse(rawIgnoredFieldIds)
```
anthonyec, line 537: Should try/catch any JSON parsing, because if some dodgy data somehow gets into `rawIgnoredFieldIds`, it will be stuck in a fail loop

anthonyec (approved):
Left one comment about catching JSON parse errors. Apart from that though, looks good to me!

plugins/airtable/src/airtable.ts
triozer: Nice catch!

david-genger:
I am having an issue after this merge. 
When trying to import/sync from Airtable I get this error message. 
image

PR #116, opened 2024-11-22
Notion: import YouTube videos and blockquotes

madebyisaacr:
### Description

This pull request adds support for importing YouTube videos and blockquotes from Notion in formatted text.

Blockquotes are added with the `<blockquote>` HTML element.

YouTube videos are added with the `<template>` element and the YouTube component's identifier, which is the same way the CMS Export plugin exports YouTube videos to HTML. Having a component identifier hardcoded in the plugin may not be ideal, but as far as I can tell it works without any issues.

Other Notion video types (Vimeo and file upload) are not imported because Framer does not support those in formatted text yet.

### Testing

- [ ] Create a Notion page in a database.
- [ ] Add a blockquote to the page content.
- [ ] Add a YouTube video embed to the page content.

### DIFF
--- a/plugins/notion/src/blocksToHTML.ts
+++ b/plugins/notion/src/blocksToHTML.ts
@@ -1,6 +1,8 @@
 import { BlockObjectResponse, RichTextItemResponse } from "@notionhq/client/build/src/api-endpoints"
 import { assert } from "./utils"
 
+const YOUTUBE_ID_REGEX = /(?:youtu\.be\/|youtube\.com\/(?:watch\?v=|embed\/))(?<videoId>[^?&]+)/u
+
 export function richTextToHTML(texts: RichTextItemResponse[]) {
     return texts
         .map(({ plain_text, annotations, href }) => {
@@ -92,9 +94,15 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
                 break
             }
             case "code":
-                htmlContent += `<pre data-language="${
-                    CODE_LANGUAGE_MAPPING[block.code.language] || "JavaScript"
-                }"><code>${richTextToHTML(block.code.rich_text)}</code></pre>`
+                const language = CODE_LANGUAGE_MAPPING[block.code.language]
+
+                if (language) {
+                    htmlContent += `<pre data-language="${language}"><code>${richTextToHTML(
+                        block.code.rich_text
+                    )}</code></pre>`
+                } else {
+                    htmlContent += `<pre><code>${richTextToHTML(block.code.rich_text)}</code></pre>`
+                }
                 break
             case "quote":
                 htmlContent += `<blockquote>${richTextToHTML(block.quote.rich_text)}</blockquote>`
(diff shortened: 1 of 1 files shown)


plugins/notion/src/blocksToHTML.ts
```diff
@@ -94,6 +94,24 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
             case "code":
                 htmlContent += `<pre><code class="language-${block.code.language.replace(" ", "-")}">${richTextToHTML(block.code.rich_text)}</code></pre>`
                 break
```
triozer, line 96: We now pass the language in the dataset

```html
<pre data-language="<language>"><code>...</code></pre>
```

plugins/notion/src/blocksToHTML.ts
```diff
@@ -94,6 +94,24 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
...
+                            url: { type: "string", value: url },
+                            play: { type: "enum", value: "Off" },
+                            shouldMute: { type: "boolean", value: true },
+                        }
+                        htmlContent += `<template data-module-identifier="module:NEd4VmDdsxM3StIUbddO/9rhBPUZttCbLCWqJEL42/YouTube.js:Youtube" data-module-props='${JSON.stringify(
+                            videoProps
+                        )}'></template>`
+                    }
+                }
+                break
```
triozer, line 114: For external videos (Youtube and Vimeo), we use iframes directly

- **YouTube**:
```html
<iframe src="https://www.youtube.com/embed/<videoId>?controls=0&autoplay=1&loop=0&mute=1"></iframe>
```
- **Vimeo**:
```html
<iframe src="https://player.vimeo.com/video/<videoId>?controls=0&autoplay=1&loop=0&muted=1"></iframe>
```

triozer (commented):
Thanks for the PR!

We recently updated our CMS export, could you please update the implementation to match the newest formats?

plugins/notion/src/blocksToHTML.ts
madebyisaacr: I updated it to use the data-language attribute. This change required creating a mapping of Notion code languages to Framer code languages. Unsupported languages import as JavaScript since there's no way to have no code formatting on a code block.

plugins/notion/src/blocksToHTML.ts
madebyisaacr: Updated to use an iframe with the /embed URL for YouTube videos. Vimeo videos aren't supported in formatted text in Framer.

plugins/notion/src/blocksToHTML.ts
```diff
@@ -92,7 +92,38 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
                 break
             }
             case "code":
-                htmlContent += `<pre><code class="language-${block.code.language.replace(" ", "-")}">${richTextToHTML(block.code.rich_text)}</code></pre>`
+                htmlContent += `<pre data-language="${
+                    CODE_LANGUAGE_MAPPING[block.code.language] || "JavaScript"
+                }"><code>${richTextToHTML(block.code.rich_text)}</code></pre>`
```
triozer, line 97:
```suggestion
                const language = CODE_LANGUAGE_MAPPING[block.code.language]

                if (language) {
                    htmlContent += `<pre data-language="${language}"><code>${richTextToHTML(block.code.rich_text)}</code></pre>`
                } else {
                    htmlContent += `<pre><code>${richTextToHTML(block.code.rich_text)}</code></pre>`
                }
```

For this to work, you will need to:
- encapsulate the case in curly brackets
- type `CODE_LANGUAGE_MAPPING` to `Record<string, string | null>`.

plugins/notion/src/blocksToHTML.ts
triozer: We do have support for Vimeo, it's not visible in the UI, but if you paste a Vimeo link it will embed it.

plugins/notion/src/blocksToHTML.ts
```diff
@@ -92,7 +92,38 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
...
+
+                            if (videoId && /^[a-zA-Z0-9_-]{11}$/.test(videoId)) {
+                                const embedUrl = `https://www.youtube.com/embed/${encodeURIComponent(
+                                    videoId
+                                )}?controls=0&autoplay=0&loop=0&mute=1`
+                                htmlContent += `<iframe src="${embedUrl}"></iframe>`
+                            }
+                        } catch (error) {
+                            console.warn("Invalid YouTube URL:", url)
+                        }
```
triozer, line 124: We could potentially rely on RegEx to simplify this logic:

```tsx
const VIMEO_ID_REGEX = /vimeo\.com\/(?<videoId>[^?]+)/u
const YOUTUBE_ID_REGEX = /(?:youtu\.be\/|youtube\.com\/(?:watch\?v=|embed\/))(?<videoId>[^?&]+)/u
```
I'll let you adjust the embed src for Vimeo, or you can use Framer's default by not passing any query variables.
```suggestion
 const url = block.video.external.url

                    const vimeoId = url.match(VIMEO_ID_REGEX)?.groups?.videoId
                    const youtubeId = url.match(YOUTUBE_ID_REGEX)?.groups?.videoId

                    if (vimeoId) {
                        htmlContent += `<iframe src="https://player.vimeo.com/video/${vimeoId}"></iframe>`
                        break
                    } else if (youtubeId) {
                        htmlContent += `<iframe src="https://www.youtube.com/embed/${youtubeId}"></iframe>`
                        break
                    }

                    console.warn("Unsupported video URL:", block.video.external.url)
```

madebyisaacr: Great, this is much simpler. I updated it to use your code for YouTube videos, but left Vimeo videos out because it shows this in the formatted text when importing videos from Vimeo:

image

Since Vimeo videos aren't supported in the formatted text editor yet, we can leave it out for now and add it later when it becomes officially supported.

PR #118, opened 2024-11-25
Transform value with line breaks to formatted text

triozer:
### Description

If a cell contains a line break, then we should make the field type a Formatted Text.

### Testing

- [ ] Import a Google Sheets with a cell having a long text with paragraphs
- [ ] The type should be infered to `Formatted Text`

### DIFF
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -55,7 +55,7 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
         const cellValueLowered = cellValue.trim().toLowerCase()
 
         // If the cell value contains a newline, it's probably a formatted text field
-        if (/\n/.test(cellValueLowered)) return "formattedText"
+        if (cellValueLowered.includes("\n")) return "formattedText"
         if (/^\d{1,2}\/\d{1,2}\/\d{4}/.test(cellValueLowered)) return "date"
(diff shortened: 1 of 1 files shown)


plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -52,13 +52,13 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
     if (typeof cellValue === "number") return "number"
 
     if (typeof cellValue === "string") {
-        const cellValueLowered = cellValue.toLowerCase()
+        const cellValueLowered = cellValue.trim().toLowerCase()
 
+        // If the cell value contains a newline, it's probably a formatted text field
+        if (/\n/.test(cellValueLowered)) return "formattedText"
```
alecmev, line 58: Any reason not to combine with the other regex? For sake of an early return?

triozer: > For sake of an early return?

Yes, that's correct. It's a small optimization win!

alecmev: Makes sense! Speaking of, maybe this?

```suggestion
        // If the cell value contains a newline, it's probably a formatted text field
        if (cellValueLowered.includes("\n")) return "formattedText"
```

triozer: Nice one!

PR #119, opened 2024-11-25
When syncing only check for non-ignored fields

triozer:
### Description

This change will prevent opening the management view when an ignored column is changed, this will allow seamless syncing.

### Testing

- [ ] It should synchronize a spreadsheet seamlessly
  - [ ] Import a spreadsheet and ignore some columns
  - [ ] Update the ignored columns in Google Sheets
  - [ ] Sync


alecmev (commented):
LGTM, but I'd extract the hashing into a shared function at this point, I think 😛

alecmev (approved):

PR #120, opened 2024-11-25
Add current database / collection ID mapping

triozer:
### Description

Previously, when importing a Notion database with self-referencing relationships (records that reference other records within the same database), we had to follow a two-step process:
1.	First import the database
2.	Then perform a sync to populate the reference fields

With this PR, we're adding the current collection to the table mapping system during the import process itself, which allows us to handle circular references directly during the import. This eliminates the need for the additional sync step and ensures that self-referencing relationships are preserved correctly in a single operation.

This depends on https://github.com/framer/FramerStudio/pull/19792, which adds support for circular references when inserting items in a managed collection.

### Testing

- [ ] Importing a table with self-referencing values
  - [ ] Create a Notion database that contains records with relations to other records within the same database
  - [ ] Import this database into Framer
  - [ ] Verify that all self-referencing relationships are preserved correctly in the Framer collection

### DIFF
--- a/plugins/notion/src/App.tsx
+++ b/plugins/notion/src/App.tsx
@@ -1,7 +1,7 @@
 import { framer } from "framer-plugin"
 import { useEffect, useState } from "react"
 import "./App.css"
-import { PluginContext, useSynchronizeDatabaseMutation } from "./notion"
+import { PluginContext, PluginContextNew, PluginContextUpdate, useSynchronizeDatabaseMutation } from "./notion"
 
 import { GetDatabaseResponse } from "@notionhq/client/build/src/api-endpoints"
 import { SelectDatabase } from "./SelectDatabase"
@@ -13,7 +13,11 @@ interface AppProps {
     context: PluginContext
 }
 
-export function AuthenticatedApp({ context }: AppProps) {
+interface AuthenticatedAppProps {
+    context: PluginContextNew | PluginContextUpdate
+}
+
+export function AuthenticatedApp({ context }: AuthenticatedAppProps) {
     const [databaseConfig, setDatabaseConfig] = useState<GetDatabaseResponse | null>(
         context.type === "update" ? context.database : null
     )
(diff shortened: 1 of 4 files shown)


tom-james-watson:
What's the context of this change?

tom-james-watson:
Looking at this code again, it seems like I unnecessarily added this `const databaseIdKey = "notionDatabaseId"` when we already have `const pluginDatabaseIdKey = "notionPluginDatabaseId"` which seems to store the same thing. I think we can just consolidate these to use `pluginDatabaseIdKey` only?

tom-james-watson:
Was the problem that `framer.getCollections` doesn't contain the current collection and that it's therefore missing from `databaseIdMap`?

plugins/notion/src/App.tsx
```diff
@@ -32,6 +36,12 @@ export function AuthenticatedApp({ context }: AppProps) {
         })
     }, [databaseConfig])
 
+    useEffect(() => {
+        if (!databaseConfig) return
+
+        context.databaseIdMap.set(databaseConfig.id, context.collection.id)
+    }, [context, databaseConfig])
```
tom-james-watson, line 43: Why is this needed?

tom-james-watson (commented):
I'm still a bit confused about what this PR does. Why is this needed? Ideally there would be an associated issue that describes what the bug or task is. Mind adding a bit more of an explanation about what was not working before and why this is needed? 🙏

plugins/notion/src/App.tsx
triozer: image

It maps the selected Notion database to the current collection id. Could be better if moved to the `onSubmit`.

tom-james-watson:
That description now is perfect :)

plugins/notion/src/App.tsx
tom-james-watson: Doesn't this

https://github.com/framer/plugins/pull/120/files/ffde46269b1e791e784d2afcf9650aeece4748b4#diff-0b64b774d3bc268fbeb7b15c667dac368a9c45ee00413a89778ec0efa2d4ef1fR784-R786

also do that, though?

triozer: Ah, this would work only if `databaseId` is set which is not the case when importing a new one (importing vs syncing).

But you are right in the sense that we should never enter this if statement because when we're importing a new database, `databaseId` will be `null`, and when we're syncing an existing database, the ID will already be in the map from the collection iteration. So this is effectively dead code that can never be reached in either scenario.

tom-james-watson: OK - best to remove then, it sounds like

PR #130, opened 2024-12-03
Add initial version of Localization Import & Export

tom-james-watson:
### Description

Adds an initial version of a "Localization Import & Export" plugin, which uses the new unstable localization APIs that are under development.

Depends on https://github.com/framer/FramerStudio/pull/19930.

Closes https://github.com/framer/company/issues/31112.

### Testing

- [ ] Download and install https://poedit.net/
- [ ] Use an alpha channel project
- [ ] Create a project and add some pieces of rich text to the canvas
- [ ] Set default locale and add two locales
- [ ] Translate some items and leave others untranslated
- [ ] Go to localization view and press Plugins
- [ ] Open plugin from URL: https://plugins-l2p4yzcb8-framer-app.vercel.app/localization-import-export
- [ ] Confirm that you can select each locale and export an xliff file
- [ ] Open the file with Poedit, make some changes to translations and save the file
- [ ] Open the plugin again and import the updated file
- [ ] The updated translations should be reflected in Framers localizations


plugins/localization-import-export/package-lock.json
triozer, line 1: I don't think we need to commit this lockfile, we can rely on NPM Workspaces to handle it.

tom-james-watson: Right yeah, that was copied over from having first developed it outside of this repo 👍

plugins/localization-import-export/.gitignore
alecmev, line 1: Also drop?

plugins/localization-import-export/src/App.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export function App() {
+    const [selectedLocaleId, setSelectedLocaleId] = useState<string>("")
+    const [locales, setLocales] = useState<readonly Locale[]>([])
+    const [defaultLocale, setDefaultLocale] = useState<Locale | null>(null)
+
+    useEffect(() => {
+        async function loadLocales() {
+            const allLocales = await framer.unstable_getLocales()
+            const defaultLocale = await framer.unstable_getDefaultLocale()
```
alecmev, line 57: Nit: shadowing.

plugins/localization-import-export/src/App.tsx
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        loadLocales()
+    }, [])
+
+    async function handleExport() {
+        if (!selectedLocaleId || !defaultLocale) return
+
+        const targetLocale = locales.find(locale => locale.id === selectedLocaleId)
+        if (!targetLocale) return
```
alecmev, line 73: Nit: assert/throw?

plugins/localization-import-export/src/xliff.ts
alecmev, line 1: I can see the upsides of doing our own thing, but, for posterity, why not [`xliff`](https://github.com/locize/xliff)?

plugins/localization-import-export/src/xliff.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export function createLocalizationsUpdateFromXliff(
+    xliffDocument: Document,
+    targetLocale: Locale
+): Record<LocalizationSourceId, LocalizedValuesUpdate> {
+    const update: Record<LocalizationSourceId, LocalizedValuesUpdate> = {}
+
+    const units = xliffDocument.querySelectorAll("unit")
+    units.forEach(unit => {
```
alecmev, line 68: Nit: `for-of`?

plugins/localization-import-export/src/xliff.ts
```diff
@@ -0,0 +1,78 @@
...
+        const id = unit.getAttribute("id")
+        const target = unit.querySelector("target")?.textContent
+        if (!id || !target) return
+
+        const targetValue = target ? target : null
```
alecmev, line 73: Nit:

```suggestion
        const targetValue = target ?? null
```

tom-james-watson: That's not quite the same - if `target` is `""` then we want it to be `null`

alecmev: Ah, pardon, good point, I'm so used to [`strict-boolean-expressions`](https://typescript-eslint.io/rules/strict-boolean-expressions/)...

alecmev: Ah, it can't be falsey here though, due to the return above.

alecmev (approved):
LGTM! 👌 Leaving QA to Hunter though 😉

PR #140, opened 2024-12-17
increase modal size to 600x600

marco-framer:
### Description

The Google Sheets plugin UI is very small which makes it hard to read longer field names. Suggest increasing the modal size to 600x600 pixels. 
*Old:*
CleanShot 2024-12-13 at 14 35 22@2x

*New:*
CleanShot 2024-12-13 at 14 38 31@2x

### DIFF
--- a/plugins/google-sheets/src/App.tsx
+++ b/plugins/google-sheets/src/App.tsx
@@ -88,8 +88,8 @@ export function AuthenticatedApp({ pluginContext, setContext }: AuthenticatedApp
 
     useLayoutEffect(() => {
         framer.showUI({
-            width: sheetTitle !== null ? 340 : 320,
-            height: sheetTitle !== null ? 425 : 345,
+            width: sheetTitle !== null ? 600 : 320,
+            height: sheetTitle !== null ? 600 : 345,
         })
     }, [sheetTitle])
 


plugins/google-sheets/src/App.tsx
```diff
@@ -88,8 +88,8 @@ export function AuthenticatedApp({ pluginContext, setContext }: AuthenticatedApp
 
     useLayoutEffect(() => {
         framer.showUI({
-            width: sheetTitle !== null ? 340 : 320,
-            height: sheetTitle !== null ? 425 : 345,
+            width: sheetTitle !== null ? 600 : 320,
+            height: sheetTitle !== null ? 600 : 345,
```
tom-james-watson, line 92: I'll leave this one for @huntercaron, but looks like at least the falsey case needs handling here

triozer (commented):
For the Airtable and Google Sheet plugins, we allow to resize the modal. We can also increase a bit the width from 340 to 360.

https://github.com/framer/plugins/blob/fe474f2481b6b5ec4ae9a102f206e20cc15b00de/plugins/airtable/src/App.tsx#L80-L93

huntercaron:
Yeah I think making this resizable like @triozer mentioned is the right move here

triozer:
Supersetted by https://github.com/framer/plugins/pull/140. Thanks @marco-framer! :)

PR #146, opened 2024-12-20
Fix Airtable 100 records limit

triozer:
### Description

The Airtable API has a 100 records per request limitation, they provide a `offset` field on the response body when there is more data to fetch.

https://airtable.com/developers/web/api/list-records

### Fixes

- https://www.framer.community/c/plugins/new-airtable-plugin-from-framer-only-imports-100-records

### Testing

- [ ] Import a base with more than 100 items
  - [ ] Open https://airtable.com/appsSUBLDCC9PvjXt/shrhmouTrRQK4saLh/tblzuKOcc3gjwX7Gd/viws9tt34yvdmT8QE and copy base to your workspace (for Framer employees there is a shared account for Airbase, ping me on Slack)
  - [ ] Import the base
  - [ ] You should have 1000 items

### DIFF
--- a/plugins/airtable/src/api.ts
+++ b/plugins/airtable/src/api.ts
@@ -438,17 +438,32 @@ export const fetchBases = (offset?: string): Promise<BasesResponse> => {
  */
 export const fetchRecords = async (args: FetchRecordsParams): Promise<AirtableRecord[]> => {
     const { baseId, tableId } = args
+    const records: AirtableRecord[] = []
+    let offset = ""
 
-    const data = await request({
-        path: `/${baseId}/${tableId}`,
-        method: "get",
-        query: {
-            returnFieldsByFieldId: "true",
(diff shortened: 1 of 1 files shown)


alecmev (commented):
Works great, but, having fixed the error logging, it turns out that this hits Airtable's rate limit of [5 requests per second per base](https://airtable.com/developers/web/api/rate-limits), need to throttle. As they mention in the linked page, their own SDK does this out of the box, maybe we could just use that instead? We get free types too 😛

triozer:
I wanted to do that in a follow-up PR 🙄 but let's do it now!

alecmev:
Can merge this and do it in a separate PR, no problem at all! This still works somehow, despite the rate-limiting 😉

alecmev (approved):

triozer:
Yeah, then let's do that as using their official SDK will add a bunch of noise

PR #151, opened 2024-12-23
Allow Google Sheets modal to be resizable

triozer:
### Description

See https://github.com/framer/plugins/pull/140.

### Testing

- [ ] Open the plugin in a managed collection
  - [ ] You should be able to resize the window

### DIFF
--- a/plugins/google-sheets/src/App.tsx
+++ b/plugins/google-sheets/src/App.tsx
@@ -87,9 +87,17 @@ export function AuthenticatedApp({ pluginContext, setContext }: AuthenticatedApp
     }, [isUserInfoError, isSelectSheetError, setContext])
 
     useLayoutEffect(() => {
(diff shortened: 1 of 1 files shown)


plugins/google-sheets/src/App.tsx
```diff
@@ -87,9 +87,17 @@ export function AuthenticatedApp({ pluginContext, setContext }: AuthenticatedApp
     }, [isUserInfoError, isSelectSheetError, setContext])
 
     useLayoutEffect(() => {
+        const width = sheetTitle ? 360 : 320
+        const height = sheetTitle ? 425 : 345
```
alecmev, line 91: No `!== null` on purpose? If not, I'd compare to `null` explicitly in `resizable` too.

alecmev (approved):
Didn't test, but looks simple enough!

PR #161, opened 2025-01-27
Fix Google Sheet date

triozer:
### Description

This pull request fixes how we import Google Sheets dates. This 

### Testing

On the sheet `With Date` there are some dates (inserted using Google Sheets date tool): https://docs.google.com/spreadsheets/d/1o1RaseNzmT1ug5kYQr-LWrHbCWCglOISJpv8jGdxC28/edit?usp=sharing.

- [ ] It imports date into the CMS
  - [ ] Start the OAuth Worker locally
  - [ ] Start the Plugin locally
  - [ ] Launch the Plugin in Framer
  - [ ] Import the `With Date` sheet from → https://docs.google.com/spreadsheets/d/1o1RaseNzmT1ug5kYQr-LWrHbCWCglOISJpv8jGdxC28/edit?usp=sharing.
  - [ ] All columns with date values should be recognized as Date field
  - [ ] It imports the correct date

### DIFF
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -56,7 +56,8 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
 
         // If the cell value contains a newline, it's probably a formatted text field
         if (cellValueLowered.includes("\n")) return "formattedText"
-        if (/^\d{1,2}-\d{1,2}-\d{4}/.test(cellValueLowered)) return "date"
+        const maybeDate = new Date(cellValueLowered)
+        if (!Number.isNaN(maybeDate.getTime())) return "date"
         if (/^#[0-9a-f]{6}$/.test(cellValueLowered)) return "color"
         if (/<[a-z][\s\S]*>/.test(cellValueLowered)) return "formattedText"
 
--- a/plugins/google-sheets/src/sheets.ts
+++ b/plugins/google-sheets/src/sheets.ts
@@ -283,7 +283,7 @@ function getFieldValue(fieldType: CollectionFieldType, cellValue: CellValue) {
     switch (fieldType) {
         case "number": {
             const num = Number(cellValue)
-            if (isNaN(num)) {
+            if (Number.isNaN(num)) {
                 return null
             }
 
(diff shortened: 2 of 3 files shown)


plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -56,7 +56,8 @@ const inferFieldType = (cellValue: CellValue): CollectionFieldType => {
 
         // If the cell value contains a newline, it's probably a formatted text field
         if (cellValueLowered.includes("\n")) return "formattedText"
-        if (/^\d{1,2}-\d{1,2}-\d{4}/.test(cellValueLowered)) return "date"
+        const maybeDate = new Date(cellValueLowered)
+        if (!isNaN(maybeDate.getTime())) return "date"
```
jonastreub, line 60: With Sheets we don't allow users to pick the field type? And Google doesn't give us any types either?

plugins/google-sheets/src/sheets.ts
```diff
@@ -294,9 +294,12 @@ function getFieldValue(fieldType: CollectionFieldType, cellValue: CellValue) {
         }
         case "date": {
             if (typeof cellValue !== "string") return null
-            const [month, day, year] = cellValue.split(/[-/]/).map(Number)
-            const date = new Date(Date.UTC(year, month - 1, day, 12, 0, 0, 0))
-            return date.toISOString()
+        try {
+                const date = new Date(Date.parse(cellValue))
+                return date.toISOString()
+            } catch (e) {
```
jonastreub, line 300: Nit. We try not to use abbreviations. But e seems unused here so you can completely remove the variable I think

plugins/google-sheets/src/pages/MapSheetFields.tsx
triozer: Sheets allows the user to select, but we try to give the best suggestion. And no, the API only returns the cell values without much information.

alecmev: Can be done, in theory, with [`effectiveFormat`](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/cells).

alecmev, line 60: Nit: Better version of `isNaN`, but make sure it behaves how we want:

```suggestion
        if (!Number.isNaN(maybeDate.getTime())) return "date"
```

triozer: @alecmev this is gold! Let's use this option when refactoring the plugin (to use the starter)? Wdyt?

alecmev: Absolutely! It's definitely a refactor 😌

jonastreub (approved):
LGTM 👏

PR #168, opened 2025-02-12
Explicitly set undefined column value to null

triozer:
### Description

This pull request ensures that when a field does not have a defined value it's set to null. Allowing to clear it.

### Context

https://framer-team.slack.com/archives/C06L5H5ADK2/p1739362411251979

### Testing

- [ ] Set values to null when re-syncing a spreadsheet
  - [ ] Import a spreadsheet
  - [ ] Remove any value (not the slug)
  - [ ] Re-sync the spreadsheet
  - [ ] Check that the field value is either unset or set to the default value (in case of a number)

### DIFF
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -222,11 +222,17 @@ export function MapSheetFieldsPage({
                 return field
             })
 
+        const colFieldTypes: Record<string, CollectionFieldType> = {}
+
+        for (const field of allFields) {
+            colFieldTypes[field.id] = field.type
+        }
+
         onSubmit({
             fields: allFields,
             spreadsheetId,
             sheetTitle,
-            colFieldTypes: fieldConfig.map(field => field.type ?? "string"),
+            colFieldTypes,
             ignoredColumns: Array.from(disabledColumns),
             slugColumn,
             lastSyncedTime: getLastSyncedTime(pluginContext, slugColumn),
--- a/plugins/google-sheets/src/sheets.ts
+++ b/plugins/google-sheets/src/sheets.ts
@@ -392,6 +392,12 @@ function processSheetRow({
         return null
     }
 
+    for (const headerRowName of uniqueHeaderRowNames) {
+        if (!(headerRowName in fieldData)) {
+            fieldData[headerRowName] = null
+        }
+    }
+
     return {
         id: itemId,
         slug: slugValue,
--- a/plugins/google-sheets/src/sheets.ts
+++ b/plugins/google-sheets/src/sheets.ts
@@ -258,7 +258,7 @@ export interface SyncResult extends SyncStatus {
 }
 
 interface ProcessSheetRowParams {
-    fieldTypes: CollectionFieldType[]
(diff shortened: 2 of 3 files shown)


tom-james-watson:
This won't work right - because we don't accept `null` at the moment for anything other than collection references.

triozer:
> This won't work right - because we don't accept `null` at the moment for anything other than collection references.

This got merged, so this works now 🕺🏽

tom-james-watson (changes requested):
If you try and import this sheet, it will fail: https://docs.google.com/spreadsheets/d/1TcedbEOK3L9a4BCArLRkjHa5hgXtBZ93dzAL54eGApM/edit?gid=0#gid=0

plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -226,7 +226,13 @@ export function MapSheetFieldsPage({
             fields: allFields,
             spreadsheetId,
             sheetTitle,
-            colFieldTypes: fieldConfig.map(field => field.type ?? "string"),
+            colFieldTypes: fieldConfig.reduce(
```
tom-james-watson, line 229: More readable if this were something like this

```ts
        const colFieldTypes: Record<string, CollectionFieldType> = {}

        // Determine if the field type is already configured, otherwise default to "string"
        for (const colName of headerRow) {
            const field = fields.find(field => field?.name === colName)
            colFieldTypes[colName] = field?.type ?? "string"
        }

        syncSheet({
            ignoredColumns,
            slugColumn,
            fetchedSheet: sheet,
            lastSyncedTime,
            spreadsheetId,
            sheetTitle,
            fields,
            colFieldTypes,
        }).then(() => framer.closePlugin())
```

plugins/google-sheets/src/sheets.ts
```diff
@@ -392,6 +385,16 @@ function processSheetRow({
...
 
+    for (const headerRowName of uniqueHeaderRowNames) {
+        if (!(headerRowName in fieldData)) {
+            if (["string", "formattedText"].includes(fieldTypes[headerRowName])) {
+                fieldData[headerRowName] = ""
+            } else {
+                fieldData[headerRowName] = null
+            }
+        }
+    }
```
tom-james-watson, line 396: Wait why are we not just omitting these values? It's not required that you pass values for all fields.

triozer: Doing so would keep the previously set value

plugins/google-sheets/src/pages/MapSheetFields.tsx
triozer: Yes!

plugins/google-sheets/src/sheets.ts
triozer: So yeah, this won't unset the value but just ignore it

```ts
const managedCollection = await framer.getActiveManagedCollection()

await managedCollection.addItems([
  {
    id: "test",
    slug: "test",
    fieldData: {
      "First Name": {
         type: "string",
         value: "John",
      },
    },
  },
])

await managedCollection.addItems([{ id: "test", slug: "test", fieldData: {} }])
```

PR #169, opened 2025-02-14
Fix logic for selecting initial airtable slug

tom-james-watson:
### Description

Fixes the selection of the initial slug field for airtable. Previously it was just using the primary field, which may or may not be a valid slug field, and didn't then match the select.

### Testing

- [x] Copy https://airtable.com/appD9FwLtlJmCHmlh/shrROIklEvIzesb0O to your account
- [x] Run this plugin locally and import it
- [x] Confirm that without changing any field mappings that it will actually use the selected slug field


triozer (approved):
Looks good!

PR #170, opened 2025-02-18
WIP: Use CMS starter for Airtable Plugin

triozer:
### Description

This pull request …

### Testing

  - [ ] Step 1
  - [ ] Step 2
  - [ ] Step 3
  - [ ] Step 1
  - [ ] Step 2
  - [ ] Step 3


triozer:
Clear tokens if no OAuth worker

tom-james-watson:
Do you have a kitchen sink airtable base I can use?

tom-james-watson:
Maybe needs fixing in the starter itself, but the disabled checkbox is less visible now:

https://github.com/user-attachments/assets/531f9a77-6288-4f3e-8c9a-e9a1267ac52c

tom-james-watson (commented):
Why are we keeping the old one around as `airtable-legacy`? Can't we just delete it? It would make the diff easier to read.

triozer:
image

We can remove.

triozer:
Maybe we should extract and merge [c17123a](https://github.com/framer/plugins/pull/170/commits/c17123ab0ec2d0e4674900ad77306ca6b10ea774)

tom-james-watson:
I think once this is merged then it should be good to publish as a new version anyway, so I don't think there's any need peresonally

triozer:
Removed it!

tom-james-watson (approved):
Looks good to me. I think we should get a couple more people to have a good QA of it on Monday before merging, though. Probably good to get Andreu to take a look! Also Alec is back Monday, too.

triozer:
I will build it and publish a new beta for it! It will be easier to QA.

plugins/airtable/src/data.ts
```diff
@@ -47,14 +49,16 @@ export async function getTables(baseId: string, signal: AbortSignal): Promise<Ai
     )
 }
 
+type AllowedTypes = ManagedCollectionField["type"] | "unsupported"
```
alecmev, line 52: Nit: `AllowedType`?

plugins/airtable/src/data.ts
```diff
@@ -272,102 +291,151 @@ export async function inferFields(collection: ManagedCollection, table: Airtable
...
 
             const cellValue = item.fields[fieldSchema.id]
             if (cellValue !== undefined) {
                 const field = fieldsById.get(fieldSchema.id)
                 if (!field) continue
 
-                itemData[fieldSchema.id] = getItemValueForField(fieldSchema, cellValue)
-            }
-        }
+                const fieldData = getItemValueForField(fieldSchema, cellValue)
```
alecmev, line 409: Nit: `fieldDataEntry`

plugins/airtable/src/data.ts
```diff
@@ -272,102 +291,151 @@ export async function inferFields(collection: ManagedCollection, table: Airtable
...
 
 export async function getItems(dataSource: DataSource, slugFieldId: string) {
     const items = await fetchRecords(dataSource.baseId, dataSource.tableId)
     const fieldsById = new Map(dataSource.fields.map(field => [field.id, field]))
-    const itemsData: ({ id: string } & Record<string, unknown>)[] = []
+    const itemsData: { id: string; slugValue: string; fieldData: FieldDataInput }[] = []
 
     for (const item of items) {
-        const itemData: Record<string, unknown> = {}
+        const itemData: FieldDataInput = {}
```
alecmev, line 401: Nit: If `fieldData` below becomes `fieldDataEntry`, then this could be just `fieldData` maybe?

plugins/airtable/src/data.ts
```diff
@@ -435,6 +433,46 @@ export async function getItems(dataSource: DataSource, slugFieldId: string) {
             slugField = fieldData
         }
 
+        // Check for missing fields and set default values.
+        // TODO: In Plugin 4.0, unset fields will be removed, this will no longer be needed.
+        for (const field of dataSource.fields) {
```
alecmev, line 438: Nit: Why not do this in the first loop? To retain the easy `continue`?

alecmev (approved):
Looking good! 👌

plugins/airtable/src/data.ts
triozer: Yeah the separate approach makes the code more maintainable, even if slightly less efficient

PR #176, opened 2025-02-25
Notion Plugin: Add block quote support

NiklasPor:
### Description

Notion supports blockquotes and Framer is supporting blockquotes. We were missing the blockquotes from notion so I added them.

### Testing

- [x] Add a blockquote in a notion doc
  - [x] Sync the notion doc to your database
  - [x] Check if it appears correctly
 
 
image


mintplo:
I appreciate your efforts, and I also needed this feature.

Based on [this PR](https://github.com/framer/plugins/pull/192), I think it would be best to assign code reviewers so that they can review your code.

CleanShot 2025-03-17 at 21 46 23

hacksman:
Add pls 🙌

PR #177, opened 2025-02-25
Notion Plugin: Add youtube embed support

NiklasPor:
### Description

Notion & Framer both support YouTube embeds. This syncs them to framer.

This pull request …

### Testing

- [x] Add youtube video as embed to notion
  - [x] Sync framer cms collection
  - [x] Check if youtube embed is rendered correctly

In Notion:
image

In Framer:
image

### DIFF
--- a/plugins/notion/src/blocksToHTML.ts
+++ b/plugins/notion/src/blocksToHTML.ts
@@ -38,9 +38,7 @@ export function richTextToHTML(texts: RichTextItemResponse[]) {
         .join("")
 }
 
-const YOUTUBE_ID_REGEX =
-    /(?:youtube\.com\/(?:[^\/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)(?<videoId>[^"&?\/\s]{11})/i
-
+const YOUTUBE_ID_REGEX = /(?:youtu\.be\/|youtube\.com\/(?:watch\?v=|embed\/))(?<videoId>[^?&]+)/iu
 export function blocksToHtml(blocks: BlockObjectResponse[]) {
     let htmlContent = ""
 


hacksman:
+1 pls

plugins/notion/src/blocksToHTML.ts
```diff
@@ -94,6 +94,24 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
             case "code":
                 htmlContent += `<pre><code class="language-${block.code.language.replace(" ", "-")}">${richTextToHTML(block.code.rich_text)}</code></pre>`
                 break
+            case "video":
+                if (block.video.type !== "external") {
+                    break
```
RayJason, line 103: Is it possible to support uploaded local videos?

NiklasPor: Feel free to implement it in a separate PR, if you find out it's possible. This one is about support for the youtube embeds.

plugins/notion/src/blocksToHTML.ts
```diff
@@ -94,6 +94,24 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
...
+                }
+
+                const videoUrl = block.video.external.url
+                const youtubeMatch = videoUrl.match(
+                    /(?:youtube\.com\/(?:[^\/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?\/\s]{11})/i
```
triozer: Can you move this regex outside the function?

`const YOUTUBE_ID_REGEX = …`

plugins/notion/src/blocksToHTML.ts
```diff
@@ -94,6 +94,24 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
...
+                )
+
+                const videoId = youtubeMatch?.[1]
+                if (!videoId) {
+                    break
+                }
+
+                // Framer styles and modifies the YouTube iframe automatically
+                htmlContent += `<iframe src="https://www.youtube.com/embed/${videoId}"></iframe>`
+                break
```
triozer: Would be easier to read like this, wdyt?

```suggestion
                const youtubeId = videoUrl.match(YOUTUBE_ID_REGEX)?.groups?.videoId
                if (youtubeId) {
                    htmlContent += `<iframe src="https://www.youtube.com/embed/${youtubeId}"></iframe>`
                }
                break
```

plugins/notion/src/blocksToHTML.ts
NiklasPor: Done

plugins/notion/src/blocksToHTML.ts
NiklasPor: I updated the implementation to use a named regex group instead.

plugins/notion/src/blocksToHTML.ts
```diff
@@ -38,6 +38,9 @@ export function richTextToHTML(texts: RichTextItemResponse[]) {
         .join("")
 }
 
+const YOUTUBE_ID_REGEX =
+    /(?:youtube\.com\/(?:[^\/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)(?<videoId>[^"&?\/\s]{11})/i
+
```
triozer: This regex seems a bit complex to me, while it's more permissive, I think we can simplify it to

```suggestion
const YOUTUBE_ID_REGEX = /(?:youtu\.be\/|youtube\.com\/(?:watch\?v=|embed\/))(?<videoId>[^?&]+)/iu
```

Wdyt?

NiklasPor: TBH I don't have any strong opinion about this, seems to work fine with regular and embed links. Also easier to read 👍

PR #179, opened 2025-02-25
Add Greenhouse Plugin

andreugordillovazquez:
### Description

This PR adds the source code for the Greenhouse plugin to the repository. The plugin enables users to import data from Greenhouse into their Framer collections.

### Testing

**Basic Import Flow**

- [ ] Open the plugin in development mode
- [ ] Add any valid company handle (e.g., `perplexityai`, `cabify`)
- [ ] Select the collection that needs to be imported (e.g., Jobs) and select all fields available
- [ ] Verify successful import
- [ ] Ensure multi-reference fields work as expected
  - [ ] Create a new collection and import a different Greenhouse collection (e.g., Departments)
  - [ ] Select the Jobs collection and enable the Departments field by clicking on "Manage"
  - [ ] Re-sync the Jobs collection and ensure the referenced items appear as expected

**Error Handling**

- [ ] Test different cases where errors are handled
  - [ ] Company handle is empty → "Connect" button should be disabled
  - [ ] Company handle is incorrect → An error toast should appear
- [ ] Verify appropriate error messages are shown


andreugordillovazquez:
@huntercaron @tom-james-watson

plugins/greenhouse/src/App.tsx
```diff
@@ -0,0 +1,188 @@
...
+        return () => clearTimeout(timeout)
+    }, [])
+
+    // Initialize Greenhouse API when board token changes
+    useLayoutEffect(() => {
+        if (boardToken) {
+            initGreenhouse(boardToken)
+                .then(() => {
+                    setIsGreenhouseInitialized(true)
+                    framer.setPluginData("greenhouse", boardToken)
```
tom-james-watson, line 39: This would be better called "board-token" or something like that

plugins/greenhouse/src/components/auth.tsx
```diff
@@ -0,0 +1,108 @@
...
+                src={Logo}
+                alt="Greenhouse Hero"
+                className="object-contain w-full rounded-[10px] h-[180px] bg-greenhouse-green bg-opacity-10"
+            />
+            
+            {/* Token Input */}
+            <div className="flex flex-col gap-[10px] text-secondary">
+                <div className="row justify-between items-center">
+                    <label htmlFor="boardToken" className="ml-[15px]">
+                        Board Token
```
tom-james-watson, line 61: As a first time user of this, I don't really know what this means. Is this a term that greenhouse uses?

tom-james-watson (commented):
Ideally this should be updated to the latest `3.0.0-beta.1`. `3.0.0` itself will be released today too. It reworks a bunch of the APIs around adding items.

We also spent quite a lot of time reworking the [CMS Starter](https://github.com/framer/plugins/tree/main/starters/cms) to provide a much cleaner and more consistent starting place for these plugins. It's a shame but I guess this was worked on before that was finished / communicated. One of the main motivations of the starter is to avoid bugs by providing a well thought out base from which to build these other plugins. So, ideally, it also would be written with that as a base.

I'm not sure how best to proceed, because potentially both of those would take quite a while 😕. We might need to talk through this in CMS sync. 🙏

andreugordillovazquez:
Closed to refactor the plugin with the CMS starter and using the latest APIs.

PR #182, opened 2025-02-28
Improve CMS Starter project setup

alecmev:
### Description

TODO

### QA

- [ ] TODO


starters/cms/vite.config.ts
huntercaron, line 1: Why not ts?

starters/cms/package.json
```diff
@@ -1,34 +1,39 @@
...
+        "eslint-plugin-react-hooks": "^5.2.0",
+        "eslint-plugin-react-refresh": "^0.4.19",
+        "framer-plugin-tools": "^1.0.0",
+        "globals": "^16.0.0",
+        "prettier": "^3.5.2",
+        "typescript": "^5.7.3",
+        "typescript-eslint": "^8.25.0",
+        "vite": "^6.2.0",
+        "vite-plugin-framer": "^1.0.6",
+        "vite-plugin-mkcert": "^1.17.7"
```
huntercaron, line 37: single numbers!! look at nextjs for example

starters/cms/vite.config.ts
alecmev: Ah, it's mentioned in the commit message:

> Moved to a JS config because the current monorepo setup isn't perfect and causes some challenging-to-resolve red squiggly lines here, and a config this simple just doesn't really need type-checking.

huntercaron: hmm I think its nice to have auto-complete and type checking in-case we make updates to the vite plugin for example. We don't want to degrade the UX of using the starter because our repo setup sucks

starters/cms/package.json
alecmev: Are you talking about this? 😏 https://github.com/vercel/next.js/commit/4ec443931aaf53bd831fa57ac5733b97d7dac550 So, they do that for templates, which don't even have their own `package.json`, but I think our starters are more like their examples, which **do** use specific versions, as they're full-fledged projects:

https://github.com/vercel/next.js/blob/canary/examples/cms-contentful/package.json

My rationale is mentioned in the commit message:

> Use full versions for dependencies as those are the minimum versions that the starter is actually confirmed to work with.

starters/cms/vite.config.ts
alecmev: That stuff and code intelligence in general is still there!

Screenshot 2025-03-06 at 12 28 01

alecmev: Moved to TS 😌

starters/cms/package.json
alecmev: To expand a little, imagine, we base this starter on `framer-plugin@3.1.0`, which has a `framer.doFoo()` method, added in `3.1.0`. But then somebody uses this starter in their Framer plugin monorepo with other plugins which are on `framer-plugin@3.0.0`. If the version in this `package.json` is `^3`, their package managed will just keep `framer-plugin` on `3.0.0` and the starter will break, since there's no `framer.doFoo()`. If the version in `package.json` is `^3.1.0`, the package manager will ensure that at least the starter gets `framer-plugin@3.1.0`, if not the whole repository (depends on the manager and its settings).

huntercaron: Ok good points! Lets go for it as you have it then

starters/cms/vite.config.ts
```diff
@@ -1,15 +1,6 @@
...
-import mkcert from "vite-plugin-mkcert"
+import { defineConfig } from "vite"
 import framer from "vite-plugin-framer"
+import mkcert from "vite-plugin-mkcert"
 
-// https://vitejs.dev/config/
-export default defineConfig({
-    plugins: [react(), mkcert(), framer()],
-    build: {
-        target: "ES2022",
```
huntercaron, line 10: We don't need this?

alecmev: Nope, set it `vite-plugin-framer` 😌

huntercaron: big brain move, thats perfect

huntercaron:
QA LGTM! ty!

PR #185, opened 2025-03-07
Fix MUST CMS Starter nits

alecmev:
### Description

- Switched from `getManagedCollection` to `getActiveManagedCollection`.
- Dropped premature `memo` around `FieldMappingRow`.
- ~Added `eslint-plugin-simple-import-sort` for import sorting.~
- ~Added `eslint-plugin-import-x` for import deduplication and other goodies.~
- Applied the import fixes.

### Testing

- [ ] `npm run check` passes without errors.
- [ ] Field-mapping still works.


huntercaron:
I don't think we need import sorting, its a bit too opinionated I think

huntercaron:
I want to be careful of how many deps we are adding, I think it was fine and we just needed to clean up the one import IMO. People should be free to make these decisions themselves

huntercaron (changes requested):
Yeah re: earlier comments, this is too much for a starter. Lets keep it simple and just fix the import manually.

huntercaron (approved):
NICE

huntercaron:
also QA LGTM

PR #187, opened 2025-03-07
getDataSources -> getDataSourceHeaders

alecmev:
### Description

Made it async, which is more realistic, and used idiomatic state-management techniques.

### Testing

- [ ] Data source selection still works.
- [ ] Field-mapping still works.

### DIFF
--- a/starters/cms/src/data.ts
+++ b/starters/cms/src/data.ts
@@ -5,11 +5,17 @@ import {
     type ManagedCollection,
     type ManagedCollectionItemInput,
 } from "framer-plugin"
+import { useEffect, useState } from "react"
 
 export const PLUGIN_KEYS = {
     DATA_SOURCE_ID: "dataSourceId",
     SLUG_FIELD_ID: "slugFieldId",
 } as const
+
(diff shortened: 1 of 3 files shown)


starters/cms/src/data.ts
```diff
@@ -5,23 +5,52 @@ import {
...
+                setDataSourceHeaders(newDataSources)
+            } catch (error) {
+                console.error(error)
+                framer.notify(`Failed to load data source headers. Check the logs for more details.`, {
+                    variant: "error",
+                })
+            } finally {
+                setAreDataSourceHeadersLoading(false)
+            }
+        })()
```
tom-james-watson, line 50: Nit:

```suggestion
        async function fetch() {
            try {
                const newDataSources = await getDataSourceHeaders()
                setDataSourceHeaders(newDataSources)
            } catch (error) {
                console.error(error)
                framer.notify(`Failed to load data source headers. Check the logs for more details.`, {
                    variant: "error",
                })
            } finally {
                setAreDataSourceHeadersLoading(false)
            }
        }
        void fetch()
```

PR #195, opened 2025-03-18
Allow importing CSV without the slug column

triozer:
### Description

This pull request allows to import CSV file that does not have a slug column/item without slug. It will try to derive it from the "Based on" value configured from Framer UI.

### Testing

Start the Plugin in dev-mode

- Add a sample blog Collection
- Export the collection using the CMS Export
- Import successfully a CSV without the slug column
  - Add an item without specifying the slug in the CSV file
  - Use this Plugin to import the CSV file
  - [ ] The import should have succeeded (the toast should show items updated/skipped + 1 item added)
  - Remove all items from the Collection
  - Remove the Slug column from the CSV file
  - Re-import the CSV file
  - [ ] The import should have succeeded (the toast should show the right number of items added)
- Fails when there is no field to derive the slug from
  - Remove the "Slug" column and the "Title" from the CSV file
  - Import the CSV file
  - [ ] You should see an error saying "Import failed. Ensure your CSV has a field named “Title”"


plugins/csv-import/src/csv.ts
```diff
@@ -167,37 +210,68 @@ function getRecordValueForField(
...
 
-    if (typeof slugFieldIndex !== "number") {
+    // Find the base field
+    const basedOnField = fields.find(field => field.id === slugField.basedOn)
+    if (!basedOnField || basedOnField.type !== "string") {
         throw new ImportError(
             "error",
-            `Import failed. Ensure your CSV has a Slug field named “${collection.slugFieldName}”`
+            `Import failed. Slug field “${slugField.name}” must be based on another string field.`
         )
```
tom-james-watson, line 255: Is this actually possible?

plugins/csv-import/src/csv.ts
```diff
@@ -167,37 +210,68 @@ function getRecordValueForField(
...
     }
 
+    // Try to find the base field in CSV headers
+    index = getFirstMatchingIndex(csvHeader, basedOnField.name)
+    if (index === -1) {
+        throw new ImportError("error", `Import failed. Ensure your CSV has a field named “${basedOnField.name}”`)
```
tom-james-watson, line 261: I think this error is maybe not what should be shown in the case the slug and the based on are both missing. In this case, I think it's best to just show the `Ensure your CSV has a field named “${slugField.name}` again.

plugins/csv-import/src/csv.ts
```diff
@@ -167,37 +210,68 @@ function getRecordValueForField(
...
+    }
+
+    return index
+}
+
+/** Importer for "records": string based values with named keys */
+export async function processRecords(collection: Collection, records: CSVRecord[]) {
+    if (!collection.slugFieldName) {
+        throw new ImportError("error", "Import failed. Ensure your CMS Collection only has one Slug field.")
+    }
```
tom-james-watson, line 271: This error doesn't seem to match what's happening? What does it actually mean if a collection doesn't have a `slugFieldName`?

plugins/csv-import/src/csv.ts
```diff
@@ -97,66 +108,98 @@ const findRecordValue = (record: CSVRecord, key: string) => {
 const collator = new Intl.Collator("en", { sensitivity: "base" })
 const BOOLEAN_TRUTHY_VALUES = /1|y(?:es)?|true/iu
 
-function getRecordValueForField(
-    field: CollectionField,
+function getFieldInputForField(
```
tom-james-watson, line 111: Ideally the upgrade would have been its own PR 🙏

tom-james-watson: To QA this properly we need a collection with all of the field types I think

plugins/csv-import/src/csv.ts
triozer: Theoretically it can happens on very old CMS collection

triozer: Theoretically yes! I don't think we enforce that on our side

plugins/csv-import/src/csv.ts
triozer: This is also only possible theoretically, on old version of Framers' CMS slug were not enforced!

plugins/csv-import/src/csv.ts
triozer: I added an action in the DevKit to be able to to that easily

plugins/csv-import/src/csv.ts
tom-james-watson: Let's make the error message match what's going wrong then - something about how no slug field was found.

tom-james-watson:
Can you add QA steps for checking fieldData upgrade too, or ideally extract that into its own PR? 🙏

triozer:
> Is this QA step right?

Actually not, I just saw that using the plugin from the Vercel link is broken :/ so using it locally seems to be the best way.

> It's being skipped

Let's remove this use case, it's a weird use-case after all, mixing update/new items. It should be fast enough to generate two CSVs for each case.

tom-james-watson:
Not sure what you mean - the existing use cases being tested in the QA make sense, but the actual checks don't seem to match what is happening

triozer:
> should be able to import when slug value is missing if title is present
Will be done in a follow up PR

> show the actual error in the toast when slug and title both missing (say that slug field is missing):
Code updated. Instead of asking the user to open te console, we directly show the error message 😄

triozer:
> should be able to import when slug value is missing if title is present

Done here https://github.com/framer/plugins/pull/205!

tom-james-watson (commented):
Can we just merge https://github.com/framer/plugins/pull/205 into this PR and merge the QA? It's effectively all the same functionality, so it's easier to review/QA if it's all in one place in this case.

tom-james-watson (approved):

PR #199, opened 2025-03-19
Allow String/Formatted Text to be interchangeable

triozer:
### Description

This pull request give more "room" to work when mapping text from Airtable. We now allow a "Single line text" to be set as a Formatted Text and a "Long text" as a String.

Fixes: https://github.com/framer/company/issues/32188.

### Testing

Start the Plugin locally or use the pending version from the Marketplace

- Select "FRAMER TEST TABLE"
- You should be able to change the mapping type to be a "String"
- [x] Importing should work


plugins/airtable/src/data.ts
```diff
@@ -284,7 +284,11 @@ export async function inferFields(collection: ManagedCollection, table: Airtable
                         break
                     case "richText":
                     case "multilineText":
-                        fields.push({ ...fieldMetadata, type: "formattedText", allowedTypes: ["formattedText"] })
+                        fields.push({
+                            ...fieldMetadata,
+                            type: "formattedText",
+                            allowedTypes: ["formattedText"],
+                        })
```
alecmev, line 291: No change?

plugins/airtable/src/data.ts
```diff
@@ -141,7 +141,7 @@ export async function inferFields(collection: ManagedCollection, table: Airtable
                     ...fieldMetadata,
                     airtableType: fieldSchema.type,
                     type: "string",
-                    allowedTypes: ["string"],
+                    allowedTypes: ["formattedText", "string"],
```
alecmev, line 144: Does the order matter? Maybe `"string"` should be first if it does?

triozer: Yes it does! Updated!

plugins/airtable/src/data.ts
```diff
@@ -284,7 +284,11 @@ export async function inferFields(collection: ManagedCollection, table: Airtable
                         break
                     case "richText":
                     case "multilineText":
-                        fields.push({ ...fieldMetadata, type: "formattedText", allowedTypes: ["formattedText"] })
+                        fields.push({
+                            ...fieldMetadata,
+                            type: "formattedText",
+                            allowedTypes: ["formattedText", "string"],
```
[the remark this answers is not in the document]
triozer: Ideally, we should extract that to some functions and call them here too. As a formula, can return a formula 🤪. Did not ticket yet but we also may add support for Lookup/Rollup fields. More of a feature than a QoL update

PR #202, opened 2025-03-20
Update Locale Sync for 3.1.0

tom-james-watson:
### Testing

- [x] Can still export locale
- Update a couple of localizations in the xliff file
- [x] Can still import locale

### DIFF
--- a/plugins/locale-sync/src/xliff.ts
+++ b/plugins/locale-sync/src/xliff.ts
@@ -13,7 +13,7 @@ function escapeXml(unsafe: string): string {
 
 /** See http://docs.oasis-open.org/xliff/xliff-core/v2.0/os/xliff-core-v2.0-os.html#state */
 type XliffState = "initial" | "translated" | "reviewed" | "final"
-type FramerStatus = LocalizationSource["locales"][Locale["id"]]["status"]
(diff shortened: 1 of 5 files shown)


plugins/locale-sync/src/xliff.ts
```diff
@@ -13,7 +13,7 @@ function escapeXml(unsafe: string): string {
 
 /** See http://docs.oasis-open.org/xliff/xliff-core/v2.0/os/xliff-core-v2.0-os.html#state */
 type XliffState = "initial" | "translated" | "reviewed" | "final"
-type FramerStatus = LocalizationSource["locales"][Locale["id"]]["status"]
+type FramerStatus = LocalizationSource["valueByLocale"][Locale["id"]]["status"]
```
triozer, line 16: Can we re-use `LocalizedValueStatus`?

triozer (approved):
Looks good, QAing!

triozer:
QA done! It's a bit hard to really notice what changed (and if the changes have been taken into account) maybe we can add the count in the toast?

triozer:
Also, can we fix?

image

tom-james-watson:
The count is maybe not trivial, depending on how you count. I think it's best to revisit, because I also think that maybe we should be clearing/ignoring too, but that requires more changes.

PR #203, opened 2025-03-20
Add the current collection id on import

triozer:
### Description

This pull request fixes a bug where we could not use the active collection when importing.

### Testing

- Import a table which references itself (Employee directory)
- [x] You should be able to set the field to the current active collection

### DIFF
--- a/plugins/airtable/src/fields.ts
+++ b/plugins/airtable/src/fields.ts
@@ -163,15 +163,18 @@ function inferDateField(
 
 async function inferRecordLinksField(
     fieldSchema: AirtableFieldSchema & { type: "multipleRecordLinks" },
-    collection: ManagedCollection
+    collection: ManagedCollection,
+    tableIdBeingLinkedTo: string
 ): Promise<PossibleField> {
     const foundCollections: { id: string; name: string }[] = []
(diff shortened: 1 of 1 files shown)


tom-james-watson (approved):

plugins/airtable/src/fields.ts
```diff
@@ -163,15 +163,18 @@ function inferDateField(
...
     const type = fieldSchema.options.prefersSingleRecordLink ? "collectionReference" : "multiCollectionReference"
 
     if (fieldSchema.options.linkedTableId) {
-        // Check if the linked table is the current collection
         const tableId = await collection.getPluginData(PLUGIN_KEYS.TABLE_ID)
-        if (tableId === fieldSchema.options.linkedTableId) {
+
+        if (tableIdBeingLinkedTo === fieldSchema.options.linkedTableId) {
+            foundCollections.push({ id: collection.id, name: "This Collection" })
+        } else if (tableId === fieldSchema.options.linkedTableId) {
```
alecmev, line 177: Why not a single condition? 🤔

triozer: Could! But I feel like it's harder to forget like this

PR #204, opened 2025-03-24
Update framer-plugin to latest and upgrade field data

triozer:
### Description

This pull request updates the version of framer-plugin used from 2 from 3. And also uses the new FieldData APIs.

### Testing

- Start this plugin locally
- Use the DevKit to add all supported fields (Select a non-managed Collection then open the devkit → Actions → Add all possible fields)
- Add an item
- Use CMS Export (or DevKit export) to export the collection as a CSV
- Remove the item
- Import the CSV
- [x] The import should succeed

### DIFF
--- a/plugins/csv-import/src/csv.ts
+++ b/plugins/csv-import/src/csv.ts
@@ -1,4 +1,12 @@
-import { Collection, CollectionField, CollectionItem, CollectionItemInput, framer } from "framer-plugin"
+import {
+    Collection,
+    Field,
+    CollectionItem,
+    CollectionItemInput,
+    framer,
+    FieldDataInput,
+    FieldDataEntryInput,
+} from "framer-plugin"
 
 type CSVRecord = Record<string, string>
 
@@ -78,7 +86,10 @@ class ImportError extends Error {
      * @param variant Notification variant to show the user
      * @param message Message to show the user
      */
-    constructor(readonly variant?: "error" | "warning", message?: string) {
+    constructor(
+        readonly variant?: "error" | "warning",
+        message?: string
+    ) {
         super(message)
     }
 }
@@ -86,7 +97,7 @@ class ImportError extends Error {
 /** Used to indicated a value conversion failed, used by `RecordImporter` and `setValueForVariable` */
 class ConversionError extends Error {}
 
-const findRecordValue = (record: CSVRecord, key: string) => {
+function findRecordValue(record: CSVRecord, key: string) {
     const value = Object.entries(record).find(([k]) => collator.compare(k, key) === 0)?.[1]
     if (!value) {
         return null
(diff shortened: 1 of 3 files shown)


plugins/csv-import/src/csv.ts
```diff
@@ -97,66 +108,98 @@ const findRecordValue = (record: CSVRecord, key: string) => {
...
+
+        case "link":
+            return { type: "link", value }
+
         case "file":
+            return { type: "file", value }
+
         case "image":
-            return value.trim()
+            return { type: "image", value }
```
tom-james-watson, line 139: Should probably keep the `trim()`

```suggestion
            return { type: "image", value: value ? value.trim() : null }
```

plugins/csv-import/src/csv.ts
```diff
@@ -97,66 +108,98 @@ const findRecordValue = (record: CSVRecord, key: string) => {
...
+            return { type: "number", value: number }
         }
 
-        case "boolean":
-            return BOOLEAN_TRUTHY_VALUES.test(value)
+        case "boolean": {
+            if (value === null) {
+                return { type: "boolean", value: false }
+            }
+            return { type: "boolean", value: BOOLEAN_TRUTHY_VALUES.test(value) }
```
tom-james-watson, line 156: `BOOLEAN_TRUTHY_VALUES.test(null)` is `false`

```suggestion
            return { type: "boolean", value: BOOLEAN_TRUTHY_VALUES.test(value) }
```

plugins/csv-import/src/csv.ts
```diff
@@ -97,66 +108,98 @@ const findRecordValue = (record: CSVRecord, key: string) => {
...
                 return new ConversionError(`Invalid value for field “${field.name}” expected a valid date`)
+            }
             const isoDate = date.toISOString().split("T")[0]
-            return new Date(isoDate).toJSON()
+            return { type: "date", value: new Date(isoDate).toJSON() }
         }
 
         case "enum": {
+            if (value === null) {
+                return { type: "enum", value: field.cases[0].id }
```
tom-james-watson, line 173: Should probably check for the case existing (and throw an error if it doesn't)

plugins/csv-import/src/csv.ts
```diff
@@ -97,66 +108,98 @@ const findRecordValue = (record: CSVRecord, key: string) => {
 const collator = new Intl.Collator("en", { sensitivity: "base" })
 const BOOLEAN_TRUTHY_VALUES = /1|y(?:es)?|true/iu
 
-function getRecordValueForField(
-    field: CollectionField,
+function getFieldInputForField(
```
tom-james-watson, line 111:
```suggestion
function getFieldDataEntryInputForField(
```

plugins/csv-import/src/csv.ts
triozer: RegExp#test only allows a string :( TypeScript not

triozer:
```suggestion
            return { type: "boolean", value: value ? BOOLEAN_TRUTHY_VALUES.test(value) : false }
```

plugins/csv-import/src/csv.ts
triozer: I don't this would be possible right? I'll add a check for it anyway

plugins/csv-import/src/csv.ts
tom-james-watson: Ah ok - yeh looks good

plugins/csv-import/src/csv.ts
tom-james-watson: I think it shouldn't, but really we should turn on unchecked indexed access lints anyway

triozer: I'd say we should probably add biome globally, but IIRC @huntercaron had some opinions on this

tom-james-watson: Yeah I think biome for this might be a bit confusing for people forking. But we can turn it on in eslint.

tom-james-watson (approved):

PR #205, opened 2025-03-25
Allow missing slug value when a based on field column is present

triozer:
### Description

This pull request allows a mix of defined and undefined slug value when there is a based on field present.

### Testing

- Add a sample blog Collection
- Export the collection using the CMS Export
- Import successfully a CSV without the slug column
  - Add an item without specifying the slug in the CSV file
  - Use this Plugin to import the CSV file
  - [ ] The import should have succeeded (the toast should show items updated/skipped + 1 item added)
  - Remove the "Title" field
  - Re-import the CSV file
  - [ ] The import should have succeeded (the toast should show items added + 1 item skipped because of missing slug field)

### DIFF
--- a/plugins/csv-import/src/csv.ts
+++ b/plugins/csv-import/src/csv.ts
@@ -224,21 +224,20 @@ function findSlugFieldIndex(
     csvHeader: string[],
     slugField: { name: string; basedOn?: string | null },
     fields: Field[]
-): number {
+): { slugIndex: number; basedOnIndex: number } {
     // Try direct match first
-    let index = getFirstMatchingIndex(csvHeader, slugField.name)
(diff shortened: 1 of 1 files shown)


tom-james-watson:
I still can't import either of these

[Articles-no-slug-field.csv](https://github.com/user-attachments/files/19449910/Articles-no-slug-field.csv)
[Articles-missing-slug.csv](https://github.com/user-attachments/files/19449911/Articles-missing-slug.csv)

tom-james-watson:
For no slug field, it errors saying it needs a "Slug" field, even though there's a "Title" field.

For missing slug field value, it skips the value even though there's a "Title" field value.

PR #210, opened 2025-04-07
Add paste handler

triozer:
### Description

This pull request adds support for pasting CSV to import when the Plugin window is focused.

### Testing

[Articles.csv](https://github.com/user-attachments/files/19633491/Articles.csv)

- Copy the CSV content
- Open this Plugin via ``
- [ ] When pasting content the Plugin should run the import

### DIFF
--- a/plugins/csv-import/src/App.tsx
+++ b/plugins/csv-import/src/App.tsx
@@ -181,14 +181,24 @@ export function App({ collection }: { collection: Collection }) {
         }
     }, [])
 
-    const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
-        event.preventDefault()
+    useEffect(() => {
+        const handlePaste = async (event: ClipboardEvent) => {
+            if (!event.clipboardData) return
 
-        const formData = new FormData(form.current!)
-        const file = formData.get("file") as File
+            const csv = event.clipboardData.getData("text/plain")
+            if (!csv) return
 
-        const csv = await file.text()
+            await processAndImport(csv)
+        }
 
+        window.addEventListener("paste", handlePaste)
+
+        return () => {
+            window.removeEventListener("paste", handlePaste)
+        }
+    }, [])
+    
+    const processAndImport = async (csv: string) => {
         try {
             const csvRecords = await parseCSV(csv)
             if (csvRecords.length === 0) {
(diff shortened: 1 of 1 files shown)


plugins/csv-import/src/App.tsx
```diff
@@ -181,14 +181,24 @@ export function App({ collection }: { collection: Collection }) {
...
-        const csv = await file.text()
+            await processAndImport(csv)
+        }
+
+        window.addEventListener("paste", handlePaste)
 
+        return () => {
+            window.removeEventListener("paste", handlePaste)
+        }
+    }, [])
```
kaloyanvi, line 199: Should we add a comment here that `processAndImport` is intentionally not added to the dependency array?

plugins/csv-import/src/App.tsx
```diff
@@ -219,6 +229,17 @@ export function App({ collection }: { collection: Collection }) {
         }
     }
 
+    const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+        event.preventDefault()
+
+        const formData = new FormData(form.current!)
```
kaloyanvi, line 235:
```suggestion
        if (!form.current) return
        const formData = new FormData(form.current)
```

tom-james-watson (commented):
Strange that you need to click on the plugin window to focus it. Ideally as soon as we open a plugin it should be focused.

kaloyanvi:
> Strange that you need to click on the plugin window to focus it. Ideally as soon as we open a plugin it should be focused.

Indeed, I was just testing and this does feel strange / broken.

plugins/csv-import/src/App.tsx
tom-james-watson: Why don't we just use `useCallback` for `processAndImport`?

plugins/csv-import/src/App.tsx
```diff
@@ -181,14 +181,24 @@ export function App({ collection }: { collection: Collection }) {
...
-    const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
-        event.preventDefault()
+    useEffect(() => {
+        const handlePaste = async (event: ClipboardEvent) => {
+            if (!event.clipboardData) return
 
-        const formData = new FormData(form.current!)
-        const file = formData.get("file") as File
+            const csv = event.clipboardData.getData("text/plain")
+            if (!csv) return
```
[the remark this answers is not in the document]
triozer: Added it as a sanity check but tested with Safari 18.4, Zen Browser (Firefox 137.0) and Arc (Chrome 135.0.7049.42) and all 3 of them worked.

tom-james-watson (approved):

kaloyanvi (approved):

PR #211, opened 2025-04-09
Support permissions in CMS Starter

alecmev:
### Description

**Warning:** To be merged when permissions are out of beta!

- In sync mode: Exits on success. Shows UI on failure or if missing permissions.
- In management mode: Shows UI regardless.
- Disables the UI if missing permissions, puts an explanation `title` on the import button.

Fixes: https://github.com/framer/company/issues/32805

### Testing

- [x] Steps:

1. Follow the account setup steps from this PR: https://github.com/framer/FramerStudio/pull/21714
2. Try relevant permutations of permissions in both plugin modes (including while open).

### DIFF
--- a/starters/cms/src/FieldMapping.tsx
+++ b/starters/cms/src/FieldMapping.tsx
@@ -1,6 +1,6 @@
-import { type ManagedCollectionFieldInput, framer, type ManagedCollection } from "framer-plugin"
+import { type ManagedCollectionFieldInput, framer, type ManagedCollection, useIsAllowedTo } from "framer-plugin"
 import { useEffect, useState } from "react"
-import { type DataSource, dataSourceOptions, mergeFieldsWithExistingFields, syncCollection } from "./data"
+import { type DataSource, dataSourceOptions, mergeFieldsWithExistingFields, syncCollection, syncMethods } from "./data"
 
 interface FieldMappingRowProps {
     field: ManagedCollectionFieldInput
@@ -8,6 +8,7 @@ interface FieldMappingRowProps {
     isIgnored: boolean
     onToggleDisabled: (fieldId: string) => void
     onNameChange: (fieldId: string, name: string) => void
+    disabled: boolean
 }
 
 function FieldMappingRow({
@@ -16,6 +17,7 @@ function FieldMappingRow({
     isIgnored,
     onToggleDisabled,
     onNameChange,
+    disabled,
 }: FieldMappingRowProps) {
     return (
         <>
@@ -23,6 +25,7 @@ function FieldMappingRow({
                 type="button"
                 className={`source-field ${isIgnored ? "ignored" : ""}`}
                 onClick={() => onToggleDisabled(field.id)}
+                disabled={disabled}
             >
                 <input type="checkbox" checked={!isIgnored} tabIndex={-1} readOnly />
                 <span>{originalFieldName ?? field.id}</span>
@@ -40,7 +43,7 @@ function FieldMappingRow({
             </svg>
             <input
                 type="text"
-                disabled={isIgnored}
+                disabled={isIgnored || disabled}
                 placeholder={field.id}
                 value={field.name}
                 onChange={event => onNameChange(field.id, event.target.value)}
@@ -136,6 +139,8 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
         })
     }
 
+    const isAllowedToManage = useIsAllowedTo("ManagedCollection.setFields", ...syncMethods)
+
     const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault()
 
@@ -150,8 +155,11 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
         try {
             setStatus("syncing-collection")
 
-            const fieldsToSync = fields.filter(field => !ignoredFieldIds.has(field.id))
+            const fieldsToSync = fields
+                .filter(field => !ignoredFieldIds.has(field.id))
+                .map(field => ({ ...field, name: field.name.trim() || field.id }))
 
+            await collection.setFields(fieldsToSync)
             await syncCollection(collection, dataSource, fieldsToSync, selectedSlugField)
             await framer.closePlugin("Synchronization successful", { variant: "success" })
         } catch (error) {
@@ -189,6 +197,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                             if (!selectedField) return
                             setSelectedSlugField(selectedField)
                         }}
+                        disabled={!isAllowedToManage}
                     >
                         {possibleSlugFields.map(possibleSlugField => {
                             return (
@@ -211,13 +220,18 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                             isIgnored={ignoredFieldIds.has(field.id)}
                             onToggleDisabled={toggleFieldDisabledState}
                             onNameChange={changeFieldName}
+                            disabled={!isAllowedToManage}
                         />
                     ))}
                 </div>
 
                 <footer>
                     <hr />
-                    <button type="submit" disabled={isSyncing}>
+                    <button
+                        type="submit"
+                        disabled={isSyncing || !isAllowedToManage}
+                        title={isAllowedToManage ? undefined : "Insufficient permissions"}
+                    >
                         {isSyncing ? <div className="framer-spinner" /> : `Import ${dataSourceName}`}
                     </button>
                 </footer>
--- a/starters/cms/src/data.ts
+++ b/starters/cms/src/data.ts
@@ -4,6 +4,7 @@ import {
     framer,
     type ManagedCollection,
     type ManagedCollectionItemInput,
+    type ProtectedMethod,
 } from "framer-plugin"
 
 export const PLUGIN_KEYS = {
@@ -101,11 +102,6 @@ export async function syncCollection(
     fields: readonly ManagedCollectionFieldInput[],
     slugField: ManagedCollectionFieldInput
 ) {
-    const sanitizedFields = fields.map(field => ({
-        ...field,
-        name: field.name.trim() || field.id,
-    }))
-
     const items: ManagedCollectionItemInput[] = []
     const unsyncedItems = new Set(await collection.getItemIds())
 
@@ -123,7 +119,7 @@ export async function syncCollection(
 
         const fieldData: FieldDataInput = {}
         for (const [fieldName, value] of Object.entries(item)) {
-            const field = sanitizedFields.find(field => field.id === fieldName)
+            const field = fields.find(field => field.id === fieldName)
 
             // Field is in the data but skipped based on selected fields.
             if (!field) continue
@@ -141,13 +137,18 @@ export async function syncCollection(
         })
     }
 
-    await collection.setFields(sanitizedFields)
     await collection.removeItems(Array.from(unsyncedItems))
     await collection.addItems(items)
 
     await collection.setPluginData(PLUGIN_KEYS.DATA_SOURCE_ID, dataSource.id)
     await collection.setPluginData(PLUGIN_KEYS.SLUG_FIELD_ID, slugField.id)
 }
+
+export const syncMethods = [
+    "ManagedCollection.removeItems",
+    "ManagedCollection.addItems",
+    "ManagedCollection.setPluginData",
+] as const satisfies ProtectedMethod[]
 
 export async function syncExistingCollection(
     collection: ManagedCollection,
@@ -159,6 +160,10 @@ export async function syncExistingCollection(
     }
 
     if (framer.mode !== "syncManagedCollection" || !previousSlugFieldId) {
+        return { didSync: false }
+    }
+
+    if (!framer.isAllowedTo(...syncMethods)) {
         return { didSync: false }
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -15209,7 +15209,7 @@
             "name": "cms-starter",
             "version": "0.0.0",
             "dependencies": {
-                "framer-plugin": "^3.3.0",
+                "framer-plugin": "^3.3.1",
                 "react": "^18.3.1",
                 "react-dom": "^18.3.1"
             },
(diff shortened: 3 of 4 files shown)


starters/cms/src/FieldMapping.tsx
```diff
@@ -146,8 +154,11 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
         try {
             setStatus("syncing-collection")
 
-            const fieldsToSync = fields.filter(field => !ignoredFieldIds.has(field.id))
+            const fieldsToSync = fields
+                .filter(field => !ignoredFieldIds.has(field.id))
+                .map(field => ({ ...field, name: field.name.trim() || field.id }))
 
+            await collection.setFields(fieldsToSync)
```
triozer, line 161: This should probably move into `syncCollection`

triozer: For the starter, we don't really need to do that, but yeah, good catch.

In case we have update the choices of an enum, we want to add them when syncing (vs opening going to the management ui).

starters/cms/src/data.ts
```diff
@@ -141,14 +136,21 @@ export async function syncCollection(
...
     await collection.setPluginData(PLUGIN_KEYS.SLUG_FIELD_ID, slugField.id)
 }
 
+export const syncMethods = [
+    "ManagedCollection.getFields",
+    "ManagedCollection.getItemIds",
+    "ManagedCollection.removeItems",
+    "ManagedCollection.addItems",
+    "ManagedCollection.setPluginData",
+] as const
```
jonastreub, line 152: Can plugin authors get help from the library types in managing this?

starters/cms/src/main.tsx
```diff
@@ -4,20 +4,18 @@ import { framer } from "framer-plugin"
...
 import { createRoot } from "react-dom/client"
 import { App } from "./App.tsx"
-import { PLUGIN_KEYS, syncExistingCollection } from "./data"
+import { PLUGIN_KEYS, syncExistingCollection, syncMethods } from "./data"
 
-const activeCollection = await framer.getActiveManagedCollection()
+if (!framer.isAllowedTo("getActiveManagedCollection", "ManagedCollection.getPluginData")) {
+    await framer.closePlugin("Insufficient permissions", { variant: "error" })
+    throw new Error("Insufficient permissions for getActiveManagedCollection and ManagedCollection.getPluginData")
+}
```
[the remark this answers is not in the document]
jonastreub: We should make the life of plugin developers as easy as possible. So only have them care about certain permissions if it adds value to their plugin. This code looks like unnecessary complexity to me

starters/cms/src/FieldMapping.tsx
```diff
@@ -132,9 +132,17 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
+    const isAllowedToManage = useIsAllowedTo("ManagedCollection.setFields", ...syncMethods)
+
     const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault()
 
+        if (!isAllowedToManage) {
+            console.error("Insufficient permissions for management")
+            framer.notify("Insufficient permissions for management", { variant: "error" })
+            return
+        }
```
jonastreub, line 144: I would make the form disabled, otherwise you might think you can make changes while you can't. That's too late because it's a frustrating experience

starters/cms/src/FieldMapping.tsx
jonastreub, line 1: I think all of the checkboxes should be disabled when you only have viewing rights

starters/cms/src/FieldMapping.tsx
```diff
@@ -212,14 +223,16 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
+                    <button disabled={isSyncing || !isAllowedToManage} tabIndex={0}>
                         {isSyncing ? (
                             <div className="framer-spinner" />
-                        ) : (
+                        ) : isAllowedToManage ? (
                             <span>
                                 Import <span style={{ textTransform: "capitalize" }}>{dataSourceName}</span>
                             </span>
+                        ) : (
+                            "Insufficient permissions"
```
jonastreub, line 235: In the Framer app when you don't have sufficient permission we show exactly the same UI, the only difference is that it's disabled. Do you think without changing the text here it's a confusing experience?

jonastreub:
Looking at the screenshot you shared I would expect all inputs to be disabled. And I think the _Insufficient Permissions_ text is really ugly. Just disabling the button should be clear enough. Maybe we could add that text in the title tooltip of the button if you think that is helpful

starters/cms/src/data.ts
alecmev: Had this thought briefly, but think `as const` is fine, as ultimately this gets fed into `isAllowedTo`, which will not let you provide invalid methods. This is why I didn't export `NamespacedMethod` from `framer-plugin`.

Thoughts?

starters/cms/src/main.tsx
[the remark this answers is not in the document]
alecmev: I had the same thought as you, "this feels silly", but you and me are enjoying the access to the internals of Framer, so we know that these two methods are always allowed (for now). But this project is here to demo "best practices", and our recommendation is "don't rely on any method being available", as that's the simplest policy. I don't want people looking at this and thinking "ok, method A is checked, but B isn't, why, the docs don't say anything about this, what is going on".

starters/cms/src/FieldMapping.tsx
alecmev: I looked into it when making this PR, there isn't really a way to "disable" a form sans disabling all the submission triggers (like the submit button, which is done). This is a safeguard against some "submit on enter press" behavior of some browser. I didn't turn it into just a `return` because you'd still want to know what happened if you rely on that behavior of your browser.

starters/cms/src/data.ts
jonastreub: I think the DUX is pretty bad if you get a type error somewhere completely different from where you are typing

starters/cms/src/main.tsx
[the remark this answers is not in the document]
jonastreub: I disagree. A plugin should only be launched if it can do at least something. Plugins should never have to write logic to immediately close themselves. That's going to be very bad UX for the user of the plugin

starters/cms/src/FieldMapping.tsx
alecmev: Thought about that too 😄 Didn't do it because I think I'd prefer plugin authors to err on the side of "disable as little as possible", can't predict peoples' workflows. Just as a use case, imagine, you open a CMS plugin, see the "Import" button saying that you don't have permissions, you ask your admin to give you those permissions, and while they're at it, you're already configuring the collection, just waiting for the "Import" button to become active.

starters/cms/src/FieldMapping.tsx
jonastreub: So I would just return

starters/cms/src/FieldMapping.tsx
alecmev: Disabled submit buttons traditionally mean "something is missing in your input", overloading that is a no-go, IMO. I considered adding a pill above the disabled button, explaining what's wrong, but chose not to as to not get bogged down by now having to design a whole new thing 😄 I completely agree that ideally buttons shouldn't change their text. I also agree that we should inspire people with good UI out of the box, so will look into improving this!

starters/cms/src/main.tsx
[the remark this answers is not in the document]
alecmev: We don't know what plugins want to do though. It's a little bit more clear in the case of a CMS plugin - they'll want to do CMS things - but imagine a general-purpose canvas plugin. What if in the future we have more granular permissions and there's an ability to limit read access to CMS? And then a plugin wants to execute `framer.getCollections`, which right now is allowed for anybody.

However, I agree that it should never result in an early exit. Instead, plugins should just have disabled UI for things that they can't do, including reading, not only writing. But that would be a pretty major refactor of this starter to get to that. Is it worth it?

starters/cms/src/FieldMapping.tsx
alecmev: I feel like it's a bad move to rely on just disabling things. Disabled inputs can still be focusable. Can also forget to disable something. Turning this into a `return` would be trading robustness for brevity, is that worth it? Also, see my comment about not wanting to disable the rest of the UI: https://github.com/framer/plugins/pull/211#discussion_r2037029630

starters/cms/src/FieldMapping.tsx
jonastreub: Okay. Let's then not use a form but make it a regular disabled button

jonastreub: Or just change the type of the button to button in case its disabled

alecmev: Oh, but it's still semantically a submit button. Didn't mean it in a "let's be pedantic about HTML spec" way 😄

starters/cms/src/data.ts
```diff
@@ -141,14 +136,20 @@ export async function syncCollection(
...
 
-    await collection.setFields(sanitizedFields)
     await collection.removeItems(Array.from(unsyncedItems))
     await collection.addItems(items)
 
     await collection.setPluginData(PLUGIN_KEYS.DATA_SOURCE_ID, dataSource.id)
     await collection.setPluginData(PLUGIN_KEYS.SLUG_FIELD_ID, slugField.id)
 }
 
+// TODO: ProtectedMethod[]
```
alecmev, line 146: Waiting for `beta.3`.

starters/cms/src/main.tsx
```diff
@@ -31,3 +24,18 @@ if (didSync) {
...
+if (framer.mode === "syncManagedCollection") {
+    if (framer.isAllowedTo(...syncMethods)) {
+        const { didSync } = await syncExistingCollection(activeCollection, previousDataSourceId, previousSlugFieldId)
+        if (didSync) {
+            await framer.closePlugin("Synchronization successful", { variant: "success" })
+        } else {
+            launch()
+        }
+    } else {
+        launch()
```
alecmev, line 37: Launching UI if the permissions are off, which will look like the screenshot. Not ideal, as then you can't just sync from there, but you'll probably just close the plugin and try again later.

starters/cms/src/App.css
```diff
@@ -24,8 +24,17 @@ form {
...
-.field-input {
+select {
     width: 100%;
+}
+
+select:disabled,
+input[type="text"]:disabled,
+.ignored {
+    opacity: 0.5;
+}
```
jonastreub, line 35: I wonder if we could add this to our plugin package CSS defaults

starters/cms/src/App.css
```diff
@@ -178,3 +183,7 @@ form {
     background: linear-gradient(to bottom, transparent, var(--framer-color-bg));
     pointer-events: none;
 }
+
+.mapping footer > button {
+    text-transform: capitalize;
```
jonastreub, line 188: Why this?

starters/cms/src/main.tsx
```diff
@@ -31,3 +24,18 @@ if (didSync) {
...
     )
 }
+
+if (framer.mode === "syncManagedCollection") {
+    if (framer.isAllowedTo(...syncMethods)) {
+        const { didSync } = await syncExistingCollection(activeCollection, previousDataSourceId, previousSlugFieldId)
+        if (didSync) {
+            await framer.closePlugin("Synchronization successful", { variant: "success" })
+        } else {
+            launch()
```
jonastreub, line 34: This conditional tree is pretty hard to read. How about we check for permissions within `syncExistingCollection`?

starters/cms/src/main.tsx
jonastreub: `launch` isn't very descriptive by itself. How about naming it something like `showUI`?

starters/cms/src/data.ts
```diff
@@ -141,14 +137,19 @@ export async function syncCollection(
...
 
     await collection.setPluginData(PLUGIN_KEYS.DATA_SOURCE_ID, dataSource.id)
     await collection.setPluginData(PLUGIN_KEYS.SLUG_FIELD_ID, slugField.id)
 }
 
+export const syncMethods = [
+    "ManagedCollection.removeItems",
+    "ManagedCollection.addItems",
+    "ManagedCollection.setPluginData",
+] as const satisfies ProtectedMethod[]
```
jonastreub, line 151: Is the term _Protected_ often associated with requiring permissions? It's new to me. We could also go for something more descriptive? e.g. `MethodRequiringPermission`. Wdyt?

jonastreub: Btw. Thanks for moving the type to here. Much better ❤️

starters/cms/src/FieldMapping.tsx
```diff
@@ -204,23 +217,22 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
-                            <div className="framer-spinner" />
-                        ) : (
-                            <span>
-                                Import <span style={{ textTransform: "capitalize" }}>{dataSourceName}</span>
-                            </span>
-                        )}
+                    <hr />
+                    <button
+                        disabled={isSyncing || !isAllowedToManage}
+                        tabIndex={0}
```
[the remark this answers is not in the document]
alecmev: Yeah, it's the default, dropped! From the other button too.

starters/cms/src/data.ts
alecmev: Resolved is Slack, keeping as-is 😌

starters/cms/src/main.tsx
alecmev: Moved to `syncExistingCollection`!

starters/cms/src/App.css
alecmev: It was there as `style` before, just moved to CSS, I suppose for collections that have lowercase words in them.

starters/cms/src/App.css
alecmev: Will make an issue for that, there might be a reason for not having them.

jonastreub (approved):
Great work 👏

triozer (approved):

PR #220, opened 2025-05-03
Revamp Notion plugin

madebyisaacr:
## Description

This pull request adds many improvements and bug fixes to the Notion plugin.

### General:
- [x] Update to `framer-plugin` 3.1.0

### Design:
- [x] Increase default plugin UI size to 600 x 500 to prevent field names from being cut off.
- [x] Improve design of unsupported fields.
- [x] All field type labels now match the name of fields in Framer (e.g. Toggle instead of Boolean)
- [x] Made “Next” a primary button on the database selection page.
- [x] Added the title of the Notion database to the top of the config UI, which links to the database on Notion.
- [x] When only 1 field type is available, the dropdown is disabled. This makes it easy to know which fields can have their type changed.

### Features:
- [x] The page cover image can now be imported as an image field.
- [x] Added support for phone number, unique ID, people, created by and last edited by Notion properties. People fields are imported as the user's name.
- [x] The title field is now always first in the list.
- [x] Added support for blockquotes and YouTube videos in page content.
- [x] Improved the detection of code block languages in page content.

### Bug Fixes:
- [x] Fix email, image and number fields causing an error if they are empty. Empty number fields are imported as 0.
- [x] Fixed status fields. Status fields now work the same way as select fields.
- [x] Non-image files (such as txt files) are now skipped when importing as images.
- [x] Removed the option to import page content as string instead of formatted text. It let you do this before, but it didn't really work because it still converted the content to HTML.

## To-Do:
- [ ] Add support for formula and rollup fields.
- [ ] Improve error messages to be easily understandable and not get cut off by the notification width limit.
- [ ] Fix duplicate slugs causing an error. When a duplicate slug exists, it should increment the value with a number suffix (e.g. blog, blog-2, blog-3)

image

There's a lot of changes in this PR, but I decided it would be better to submit them all as a single PR instead of splitting into multiple, just to make it easier to review. Each feature/change is in a separate commit.

Once this is approved, I will do the same updates for the Airtable and Google Sheets plugins.

Please leave any feedback and comments you have!

PR #221, opened 2025-05-05
Convert Notion plugin to use CMS Starter

madebyisaacr:
Changes:

* Convert the Notion plugin to use the [CMS Starter](https://www.framer.com/updates/cms-starter-plugin) for shared UI and logic code.
* Update `framer-plugin` to 3.1.0
* Increase default plugin size to 600 x 500 to prevent text getting cut off.

Another PR: https://github.com/framer/plugins/pull/220 includes over 20 improvements, new field types, and bug fixes to the Notion plugin. @triozer suggested that Notion be switched to use the CMS starter first before merging that PR.

### DIFF
--- a/plugins/notion/src/App.tsx
+++ b/plugins/notion/src/App.tsx
@@ -23,8 +23,8 @@ export function AuthenticatedApp({ context }: AuthenticatedAppProps) {
     )
 
     useEffect(() => {
-        const width = databaseConfig ? 360 : 325
-        const height = databaseConfig ? 425 : 370
+        const width = databaseConfig ? 600 : 325
+        const height = databaseConfig ? 500 : 370
         framer.showUI({
             width,
             height,


plugins/notion/src/auth.ts
```diff
@@ -0,0 +1,105 @@
...
+        //     const res = await fetch(`${API_BASE_URL}/auth/authorize`, {
+        //         method: "POST",
+        //     })
+        //     console.log("res", res)
+        //     return res.status === 400
+        // } catch (error) {
+        //     console.error("Failed to connect to OAuth worker:", error)
+        //     return false
+        // }
+    }
```
madebyisaacr: This file is copied from https://github.com/framer/plugins/blob/main/plugins/airtable/src/auth.ts
The endpoint used to check `isWorkerAlive` in the Airtable plugin doesn't appear to be implemented in the Notion worker, so I made it always return true as a placeholder.

madebyisaacr:
There’s still some TypeScript errors in the code. Let me know if that’s important to fix before reviewing.

plugins/notion/tsconfig.json
alecmev, line 1: Was good before! 😉

plugins/notion/package.json
```diff
@@ -1,40 +1,41 @@
...
     "dependencies": {
         "@notionhq/client": "^2.2.15",
-        "@tanstack/react-query": "^5.29.2",
-        "classnames": "^2.5.1",
-        "framer-plugin": "^3.3.2",
-        "react": "^18",
-        "react-dom": "^18",
-        "react-error-boundary": "^4.0.13",
-        "vite-plugin-mkcert": "^1"
+        "framer-plugin": "^3.2.0",
```
alecmev:
```suggestion
        "framer-plugin": "^3.3.2",
```

alecmev (commented):
Just a couple of drive-by nits, leaving the full review to @triozer 😌

plugins/notion/tsconfig.json
madebyisaacr: Thanks, good catch! I reverted this change.

yarn.lock
```diff
@@ -7486,28 +7486,27 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "notion@workspace:plugins/notion"
   dependencies:
+    "@eslint/js": "npm:^9.21.0"
     "@notionhq/client": "npm:^2.2.15"
```
alecmev: Can you also update to version 3 of this package, please? Or is the migration challenging?

madebyisaacr: Updated to @notionhq/client version 3.1.3

plugins/notion/package.json
```diff
@@ -1,31 +1,41 @@
...
-        "preview": "run g:preview",
-        "check-typescript": "run g:check-typescript"
+        "build": "vite build",
+        "check": "npm run typecheck && npm run lint",
+        "dev": "vite",
+        "format": "prettier . --write",
+        "lint": "eslint . --max-warnings 0",
+        "pack": "framer-plugin-tools pack",
+        "preview": "vite preview",
+        "typecheck": "tsc"
```
alecmev: Make sure to revert this, just in case 😌 (except for `pack`, that one is good)

plugins/notion/package.json
```diff
@@ -1,31 +1,41 @@
...
+        "eslint-plugin-react-refresh": "^0.4.19",
+        "framer-plugin-tools": "^1.0.0",
+        "globals": "^16.0.0",
+        "jiti": "^2.4.2",
+        "prettier": "^3.5.2",
+        "typescript": "^5.7.3",
+        "typescript-eslint": "^8.25.0",
+        "vite": "^6.2.6",
+        "vite-plugin-framer": "^1.0.7",
+        "vite-plugin-mkcert": "^1.17.7"
```
alecmev: 😛 

```suggestion
        "@types/react": "^18.3.18",
        "@types/react-dom": "^18.3.5",
        "framer-plugin-tools": "^1.0.0"
```

plugins/notion/vite.config.ts
alecmev, line 1: Should be deleted, the Vite config is shared now 😌

plugins/notion/src/auth.ts
```diff
@@ -0,0 +1,105 @@
...
+        // try {
+        //     const res = await fetch(`${API_BASE_URL}/auth/authorize`, {
+        //         method: "POST",
+        //     })
+        //     console.log("res", res)
+        //     return res.status === 400
+        // } catch (error) {
+        //     console.error("Failed to connect to OAuth worker:", error)
+        //     return false
+        // }
```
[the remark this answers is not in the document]
tom-james-watson: If we don't have a relevant backend API for this then let's just remove the check entirely.

plugins/notion/src/auth.ts
tom-james-watson: If we don't have a relevant backend API for this then let's just remove the check entirely.

plugins/notion/src/SelectDataSource.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    return (
+        <main className="framer-hide-scrollbar setup">
+            <div className="intro">
+                <img src="/notion-connect.png" className="select-database-img" />
+                <p>
+                    To manually connect a database, open it in Notion, click on the three dots icon in the top right
+                    corner, then select Connections and connect to Framer.
+                </p>
```
tom-james-watson, line 70: image


This looks better aligned left IMO

plugins/notion/src/FieldMapping.tsx
```diff
@@ -0,0 +1,323 @@
...
+            <hr className="sticky-divider" />
+            <form onSubmit={handleSubmit}>
+                <label className="slug-field" htmlFor="slugField">
+                    <div className="heading-row">
+                        <span>Slug Field</span>
+                        {database?.url && (
+                            <a href={database.url} target="_blank" className="heading-link">
+                                View in Notion
+                            </a>
+                        )}
```
tom-james-watson, line 280: This is a nice idea, but it needs a better design, IMO.

It would be best to add an issue, because we should try and keep this PR as focused as possible on using the CMS Starter structure 🙏 

image

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,339 @@
...
+                assertFieldTypeMatchesPropertyType(property.type, fieldType)
+                fields.push({
+                    type: "file",
+                    id: fieldInfo.id,
+                    name: fieldInfo.name,
+                    allowedFileTypes: [],
+                    userEditable: false,
+                })
+                break
+            case "multiCollectionReference":
```
tom-james-watson: Databases that reference themselves are broken. You can test using this page:

https://www.notion.so/framer/Plugin-Testing-With-Self-References-14aadf6e8c96800eb165e7e7b447f107?source=copy_link

tom-james-watson (changes requested):
Nice work so far! Do you mind addressing the failing lints/types? I can take another pass once those are ✅

plugins/notion/src/FieldMapping.tsx
madebyisaacr: @tom-james-watson this link was also added to the Airtable plugin. We discussed the design on Slack last week, originally it looked different but we settled on this design.
https://github.com/framer/plugins/pull/249

I figured it makes sense to do this here to keep all plugins consistent.

plugins/notion/src/data.ts
madebyisaacr: Added support for self-referencing relations.

plugins/notion/src/FieldMapping.tsx
tom-james-watson: OK, let's just leave it in for now then - we can take another pass at it some time 🙏

plugins/notion/src/api.ts
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    const results = await collectPaginatedAPI(notion!.search, {
+        filter: {
+            property: "object",
+            value: "database",
+        },
+    })
+
+    return results.filter(isFullDatabase) as DatabaseObjectResponse[]
```
alecmev:
```suggestion
    return results.filter(isFullDatabase)
```

plugins/notion/src/api.ts
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    assert(notion, "Notion client is not initialized")
+    const database = await notion.databases.retrieve({ database_id: databaseId })
+
+    if (!isFullDatabase(database)) {
+        throw new Error(`Database ${databaseId} is not a full database`)
+    }
+
+    return database as DatabaseObjectResponse
```
alecmev:
```suggestion
    return database
```

plugins/notion/src/auth.ts
```diff
@@ -0,0 +1,91 @@
...
+        get: () => {
+            if (this.storedTokens) return this.storedTokens
+
+            const bearerToken = localStorage.getItem(PLUGIN_KEYS.BEARER_TOKEN)
+            if (!bearerToken) return null
+
+            const storedTokens = { bearer_token: bearerToken } as Tokens
+            this.storedTokens = storedTokens
+
+            return storedTokens
```
alecmev:
```suggestion
            this.storedTokens = { bearer_token: bearerToken }

            return this.storedTokens
```

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,361 @@
...
+                })
+                skipContent = true
+            }
+
+            let slugValue: null | string = null
+            const fieldData: FieldDataInput = {}
+
+            for (const property of Object.values(item.properties)) {
+                if (property.id === slugField.id) {
+                    const resolvedSlug = getPropertyValue(property, { type: "string" } as ManagedCollectionFieldInput)
```
alecmev:
```suggestion
                    const resolvedSlug = getPropertyValue(property, slugField)
```

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,361 @@
...
+                if (!field) continue
+
+                const fieldValue = getPropertyValue(property, field)
+                if (fieldValue === null || fieldValue === undefined) {
+                    console.warn(
+                        `Skipping item at index ${index} because it doesn't have a valid value for field ${field.name}`
+                    )
+                }
+
+                fieldData[field.id] = { type: field.type, value: fieldValue } as FieldDataEntryInput
```
alecmev: What about making `getPropertyValue` return `FieldDataEntryInput`, so that this cast isn't necessary?

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,361 @@
...
+                id: item.id,
+                slug: slugValue,
+                draft: false,
+                fieldData,
+            }
+        })
+    )
+
+    const result = await Promise.all(promises)
+    const items = result.filter(Boolean) as ManagedCollectionItemInput[]
```
alecmev: (need to add `isNotNull` to `utils.ts`)

```suggestion
    const items = result.filter(isNotNull)
```

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,361 @@
...
+        if (!slugField) {
+            framer.notify(`No field matches the slug field id “${previousSlugFieldId}”. Sync will not be performed.`, {
+                variant: "error",
+            })
+            return { didSync: false }
+        }
+
+        const ignoredFieldIds = (
+            previousIgnoredFieldIds ? new Set(JSON.parse(previousIgnoredFieldIds)) : new Set()
+        ) as Set<string>
```
alecmev:
```suggestion
        const ignoredFieldIds: Set<string> = (
            previousIgnoredFieldIds ? new Set(JSON.parse(previousIgnoredFieldIds)) : new Set()
        )
```

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,361 @@
...
+                    }
+                }
+                break
+            }
+            default:
+                throw new Error(`Unsupported field type: ${fieldType}`)
+        }
+    }
+
+    return fields as ManagedCollectionFieldInput[]
```
alecmev: This is unfortunate, that we don't have an easy-to-access type for `cases`, need to address that.

```suggestion
    const fields: ManagedCollectionFieldInput[] = []

    for (const fieldInfo of fieldsInfo) {
        const property = fieldInfo.notionProperty
        const fieldType = fieldInfo.type

        if (fieldInfo.id === pageContentProperty.id) {
            fields.push({
                type: "formattedText",
                id: fieldInfo.id,
                name: fieldInfo.name,
                userEditable: false,
            })
            continue
        }

        if (!property || !fieldType) continue

        switch (fieldType) {
            case "boolean":
            case "date":
            case "number":
            case "string":
            case "formattedText":
            case "link":
            case "image": {
                assertFieldTypeMatchesPropertyType(property.type, fieldType)
                fields.push({
                    type: fieldType,
                    id: fieldInfo.id,
                    name: fieldInfo.name,
                    userEditable: false,
                })
                break
            }
            case "enum": {
                assertFieldTypeMatchesPropertyType(property.type, fieldType)

                let cases: Extract<ManagedCollectionFieldInput, { type: "enum" }>["cases"] | null = null
                switch (property?.type) {
                    case "select":
                        cases = property.select.options.map(option => ({
                            id: option.id,
                            name: option.name,
                        }))
                        break
                    case "status":
                        cases = property.status.options.map(option => ({
                            id: option.id,
                            name: option.name,
                        }))
                        break
                }

                if (cases) {
                    fields.push({
                        type: "enum",
                        id: fieldInfo.id,
                        name: fieldInfo.name,
                        cases,
                        userEditable: false,
                    })
                }

                break
            }
            case "file": {
                assertFieldTypeMatchesPropertyType(property.type, fieldType)
                fields.push({
                    type: "file",
                    id: fieldInfo.id,
                    name: fieldInfo.name,
                    allowedFileTypes: [],
                    userEditable: false,
                })
                break
            }
            case "multiCollectionReference": {
                assertFieldTypeMatchesPropertyType(property.type, fieldType)

                if (property.type === "relation") {
                    const databaseId = property.relation?.database_id
                    if (databaseId && databaseIdMap) {
                        const collectionId = databaseIdMap.get(databaseId)
                        if (collectionId) {
                            fields.push({
                                type: "multiCollectionReference",
                                id: fieldInfo.id,
                                name: fieldInfo.name,
                                collectionId,
                                userEditable: false,
                            })
                        }
                    }
                }
                break
            }
            default:
                throw new Error(`Unsupported field type: ${fieldType}`)
        }
    }

    return fields
```

plugins/notion/src/FieldMapping.tsx
```diff
@@ -0,0 +1,338 @@
...
+            </svg>
+            {isNotSupported ? (
+                <div className="unsupported-field">{unsupported ? "Unsupported Field" : "Missing Collection"}</div>
+            ) : (
+                <>
+                    <select
+                        className="field-type"
+                        disabled={disabled || allowedTypes.length <= 1}
+                        value={type ?? ""}
+                        onChange={event => onFieldTypeChange(id, event.target.value as FieldType)}
```
alecmev: I'd add an assert here, just to make sure `event.target.value` is what we expect it to be. Also, it's `ManagedCollectionField["type"]`, not `FieldType`, right?

plugins/notion/src/FieldMapping.tsx
```diff
@@ -0,0 +1,338 @@
...
+    const isAllowedToManage = useIsAllowedTo(...syncMethods)
+
+    const [status, setStatus] = useState<"mapping-fields" | "loading-fields" | "syncing-collection">(
+        initialSlugFieldId ? "loading-fields" : "mapping-fields"
+    )
+    const isSyncing = status === "syncing-collection"
+    const isLoadingFields = status === "loading-fields"
+
+    const dataSourceName = dataSource.name
+    const database = dataSource.database as DatabaseObjectResponse
```
alecmev:
```suggestion
    const database = dataSource.database
```

alecmev (commented):
Went through all instances of ` as `. Just in case, please, use it only when it's absolutely unavoidable (like control-flow-analysis limitations).

plugins/notion/src/FieldMapping.tsx
madebyisaacr: `FieldType` is an alias for `ManagedCollectionField["type"]`
```
type FieldType = ManagedCollectionField["type"]
```

alecmev: Ah, then can you replace all instances of `ManagedCollectionField["type"]` with `FieldType` elsewhere in the plugin, please?

alecmev: Or, wait, I see, it's a custom alias in this plugin. I'd just use `ManagedCollectionField["type"]` directly, and if it's a common pattern, then we can maybe think about exporting it from `framer-plugin` later.

madebyisaacr: Done

plugins/notion/src/data.ts
madebyisaacr: It can't use `slugField` here because the type must be "string", and `slugField`'s type is not always string.

Using this solution instead. `type` is the only part of the object that gets used, but it fulfills the type requirements.

```tsx
const SLUG_FIELD: ManagedCollectionFieldInput = {
    type: "string",
    id: "slug",
    name: "Slug",
    userEditable: false,
}
```
```tsx
const resolvedSlug = getPropertyValue(property, SLUG_FIELD)
```

alecmev: I see. Taking a closer look at this, it could be a dedicated function maybe? Just for slugs? Which does a `switch` and returns the most appropriate value (or throws, for example, if `property.type` is `"checkbox"` and such).

plugins/notion/src/data.ts
madebyisaacr: Done

plugins/notion/src/data.ts
madebyisaacr: Done. Added a `getSlugValue` function for this.

plugins/notion/src/api.ts
```diff
@@ -281,77 +293,95 @@ export async function getPageBlocksAsRichText(pageId: string) {
...
+                return { type: "formattedText", value: richTextToHTML(property.rich_text) }
             }
 
-            return richTextToPlainText(property.rich_text)
+            return { type: "string", value: richTextToPlainText(property.rich_text) }
         }
         case "select": {
-            if (!property.select) {
-                return field.type === "enum" ? (field.cases?.[0]?.id ?? null) : null
+            if (field.type === "enum") {
```
alecmev: Nit:

```suggestion
            if (field.type !== "enum") return null
```

plugins/notion/src/api.ts
```diff
@@ -281,77 +293,95 @@ export async function getPageBlocksAsRichText(pageId: string) {
...
+                return { type: "enum", value: property.select.id }
             }
 
-            return property.select.id
+            return null
         }
         case "status": {
-            if (!property.status) {
-                return field.type === "enum" ? (field.cases?.[0]?.id ?? null) : null
+            if (field.type === "enum") {
```
alecmev: Same nit as above.

plugins/notion/src/api.ts
```diff
@@ -281,77 +293,95 @@ export async function getPageBlocksAsRichText(pageId: string) {
...
-            const firstFile = property.files[0]
-
-            switch (firstFile?.type) {
-                case "external":
-                    return firstFile.external.url
-                case "file":
-                    return firstFile.file.url
-                default:
-                    return null
+            if (field.type === "file" || field.type === "image") {
```
alecmev: Same nit as above, early return.

plugins/notion/src/api.ts
```diff
@@ -281,77 +293,95 @@ export async function getPageBlocksAsRichText(pageId: string) {
...
+                    default:
+                        return { type: field.type, value: null }
+                }
             }
+
+            return null
         }
         case "email": {
-            return property.email ?? ""
+            if (field.type === "formattedText" || field.type === "string") {
```
alecmev: Same nit.

plugins/notion/src/data.ts
```diff
@@ -123,26 +116,26 @@ export async function syncCollection(
 
             for (const property of Object.values(item.properties)) {
                 if (property.id === slugField.id) {
-                    const resolvedSlug = getPropertyValue(property, { type: "string" } as ManagedCollectionFieldInput)
+                    const slug = getSlugValue(property)
 
-                    if (!resolvedSlug || typeof resolvedSlug !== "string") {
+                    if (!slug) {
                         break
                     }
```
alecmev: Nit:

```suggestion
                    if (!slug) break
```

plugins/notion/src/data.ts
```diff
@@ -123,26 +116,26 @@ export async function syncCollection(
...
-                const fieldValue = getPropertyValue(property, field)
-                if (fieldValue === null || fieldValue === undefined) {
+                const fieldEntry = getFieldDataEntryForProperty(property, field)
+                if (fieldEntry === null || fieldEntry === undefined) {
                     console.warn(
                         `Skipping item at index ${index} because it doesn't have a valid value for field ${field.name}`
                     )
+                } else {
+                    fieldData[field.id] = fieldEntry
                 }
```
alecmev: Nit (no avoidable negation):

```suggestion
                if (fieldEntry) {
                    fieldData[field.id] = fieldEntry
                } else {
                    console.warn(
                        `Skipping item at index ${index} because it doesn't have a valid value for field ${field.name}`
                    )
                }
```

alecmev:
> How are you checking ESLint warnings? `yarn check-eslint` returns nothing for me.

My bad! I was looking at the warnings reported by GitHub below a commit, and didn't pay attention to the paths, they come from other plugins, all clear on that coast 👌 Reviewing!

plugins/notion/package.json
```diff
@@ -1,31 +1,26 @@
...
-        "pack": "npx framer-plugin-tools@latest pack",
         "preview": "run g:preview",
+        "pack": "framer-plugin-tools pack",
         "check-typescript": "run g:check-typescript"
     },
     "dependencies": {
-        "@notionhq/client": "^2.3.0",
-        "@tanstack/react-query": "^5.81.2",
-        "classnames": "^2.5.1",
+        "@notionhq/client": "3.1.3",
```
alecmev: Nit:

```suggestion
        "@notionhq/client": "^3.1.3",
```

yarn.lock
alecmev, line 1: Please, go through this diff and revert all changes that make versions lower. I assume it happened because of a rebase (when rebasing, the easiest way to do "the right thing" is to take the lockfile from `main` as-is and then run `yarn`).

plugins/notion/src/vite-env.d.ts
alecmev, line 1: This has been re-evaluated 😛 Please, keep the file.

plugins/notion/src/utils.ts
```diff
@@ -74,18 +60,15 @@ export function generateRandomId() {
     window.crypto.getRandomValues(array)
 
     let id = ""
-    for (const value of array) {
-        id += value.toString(16).padStart(2, "0")
+    for (let i = 0; i < array.length; i++) {
+        id += array[i]!.toString(16).padStart(2, "0")
```
alecmev: I'd keep the original, beats the `!`.

plugins/notion/src/utils.ts
```diff
@@ -74,18 +60,15 @@ export function generateRandomId() {
...
 
     return id
 }
 
-export function assertNever(x: never, error?: unknown): never {
-    throw error || new Error((x as unknown) ? `Unexpected value: ${x}` : "Application entered invalid state")
-}
-
 export const syncMethods = [
+    "ManagedCollection.setFields",
```
alecmev: This shouldn't be needed, the idea is that people without Canvas permissions should still be able to sync. See [the Airtable permission PR](https://github.com/framer/plugins/pull/237) for inspiration.

plugins/notion/src/main.tsx
```diff
@@ -1,94 +1,82 @@
...
+        activeCollection,
+        previousDatabaseId,
+        previousSlugFieldId,
+        previousIgnoredFieldIds,
+        previousLastSynced,
+        previousDatabaseName,
+        databaseIdMap
+    )
+    didSync = didSyncResult
+}
```
alecmev: Can you move `isAllowedToSync` into `syncExistingCollection`, please? Then this can just become `const { didSync } = ...`, [like in Airtable](https://github.com/framer/plugins/blob/18b8653816b22a9ab24fb4afd16eab5ddcab64cd/plugins/airtable/src/main.tsx#L43-L49).

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,352 @@
...
+    const databaseIdMap: DatabaseIdMap = new Map()
+
+    for (const collection of await framer.getCollections()) {
+        const collectionDatabaseId = await collection.getPluginData(PLUGIN_KEYS.DATABASE_ID)
+        if (!collectionDatabaseId) continue
+
+        databaseIdMap.set(collectionDatabaseId, collection.id)
+    }
+
+    return databaseIdMap
```
alecmev: To parallelize:

```suggestion
    const databaseIdMap: DatabaseIdMap = new Map()
    const promises: Promise<void>[] = []

    for (const collection of await framer.getCollections()) {
        const task = async () => {
            const collectionDatabaseId = await collection.getPluginData(PLUGIN_KEYS.DATABASE_ID)
            if (!collectionDatabaseId) return
            databaseIdMap.set(collectionDatabaseId, collection.id)
        }
        
        promises.push(task())
    }

    await Promise.all(promises)
    return databaseIdMap
```

plugins/notion/src/App.tsx
```diff
@@ -1,88 +1,136 @@
...
-    const [databaseConfig, setDatabaseConfig] = useState<GetDatabaseResponse | null>(
-        context.type === "update" ? context.database : null
-    )
+    // Support self-referencing databases by allowing the current collection to be referenced.
+    const finalDatabaseIdMap = useMemo(() => {
+        if (dataSource?.database.id) {
+            return new Map(databaseIdMap).set(dataSource.database.id, collection.id)
+        }
+        return databaseIdMap
+    }, [databaseIdMap, dataSource?.database.id, collection.id])
```
alecmev: How is it possible that `getDatabaseIdMap` doesn't populate this? Doesn't `previousDataSourceId` ultimately come from the same plugin data key?

plugins/notion/src/App.tsx
```diff
@@ -1,88 +1,136 @@
...
+import { useEffect, useLayoutEffect, useMemo, useState } from "react"
+import { type DatabaseIdMap, type DataSource, getDataSource } from "./data"
+import { FieldMapping } from "./FieldMapping"
+import { NoTableAccess } from "./NoAccess"
+import { SelectDataSource } from "./SelectDataSource"
 
 interface AppProps {
-    context: PluginContext
+    collection: ManagedCollection
+    previousDataSourceId: string | null
```
alecmev: Maybe, for consistency?

```suggestion
    previousDatabaseId: string | null
```

plugins/notion/src/FieldMapping.tsx
```diff
@@ -0,0 +1,339 @@
...
+    ignored,
+    isAllowedToManage,
+    unsupported,
+    missingCollection,
+    onToggleIgnored,
+    onNameChange,
+    onFieldTypeChange,
+}: FieldMappingRowProps) {
+    const { id, name, originalName, type, allowedTypes } = fieldInfo
+    const isNotSupported = unsupported || missingCollection
```
alecmev: Confusing name, since it's pretty much a synonym of `unsupported`.

plugins/notion/src/FieldMapping.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                unsupported={
+                                    !Array.isArray(fieldInfo.allowedTypes) || fieldInfo.allowedTypes.length === 0
+                                }
+                                missingCollection={isMissingCollection(fieldInfo, databaseIdMap)}
+                                onToggleIgnored={toggleFieldIgnoredState}
+                                onNameChange={changeFieldName}
+                                onFieldTypeChange={changeFieldType}
+                            />
+                        )
+                    })}
```
alecmev: Nit:

```suggestion
                    {fieldsInfo.map(fieldInfo => 
                        (
                            <FieldMappingRow
                                key={`field-${fieldInfo.id}`}
                                fieldInfo={fieldInfo}
                                ignored={ignoredFieldIds.has(fieldInfo.id)}
                                isAllowedToManage={isAllowedToManage}
                                unsupported={
                                    !Array.isArray(fieldInfo.allowedTypes) || fieldInfo.allowedTypes.length === 0
                                }
                                missingCollection={isMissingCollection(fieldInfo, databaseIdMap)}
                                onToggleIgnored={toggleFieldIgnoredState}
                                onNameChange={changeFieldName}
                                onFieldTypeChange={changeFieldType}
                            />
                        )
                    )}
```

plugins/notion/src/Login.tsx
```diff
@@ -0,0 +1,86 @@
...
+import auth from "./auth"
+
+interface AuthenticationProps {
+    onAuthenticated: () => void
+}
+
+export function Authenticate({ onAuthenticated }: AuthenticationProps) {
+    const [isLoading, setIsLoading] = useState(false)
+    const pollInterval = useRef<number | ReturnType<typeof setInterval>>()
+    const pollTimeout = useRef<number | ReturnType<typeof setTimeout>>()
```
alecmev: No?

```suggestion
    const pollInterval = useRef<ReturnType<typeof setInterval>>()
    const pollTimeout = useRef<ReturnType<typeof setTimeout>>()
```

plugins/notion/src/Login.tsx
```diff
@@ -0,0 +1,86 @@
...
+            const fetchTokens = () => {
+                auth.fetchTokens(readKey).then(tokens => {
+                    clearInterval(pollInterval.current)
+                    clearTimeout(pollTimeout.current)
+                    resolve(tokens)
+                })
+            }
+
+            // Initial 10 second delay
+            pollTimeout.current = setTimeout(() => {
```
alecmev: I guess it makes sense, but is it worth it, just to avoid potentially 3 unsuccessful requests?

plugins/notion/src/NoAccess.tsx
```diff
@@ -0,0 +1,42 @@
...
+import auth from "./auth"
+
+export function NoTableAccess({ previousDatabaseId }: { previousDatabaseId: string | null }) {
+    const [isRetrying, setIsRetrying] = useState(false)
+
+    const handleRetryClick = () => {
+        setIsRetrying(true)
+        auth.authorize()
+            .then(() => {
+                setIsRetrying(false)
```
alecmev: Nit: Can be in `.finally` perhaps?

alecmev (commented):
Submitting the comments I have so far, will review the few remaining files in 1-2 hours.

plugins/notion/src/FieldMapping.tsx
madebyisaacr: Changed it to `isFieldUnavailable`, is that better?

plugins/notion/src/utils.ts
madebyisaacr: Done

plugins/notion/src/App.tsx
madebyisaacr: This is for when you're creating a new collection. The plugin data key doesn't get set until after you sync the collection, so without this code self-referencing relation fields don't work until after the first time you sync.

So this code adds the current database to the `databaseIdMap` so it can be used in the initial sync.

plugins/notion/src/FieldMapping.tsx
alecmev: Yeah! 👌

yarn.lock
madebyisaacr: Done. There's still a lot of changes in yarn.lock but I think most of them are removing TailwindCSS and related dependencies.

alecmev: Awesome! Yeah, the removals are expected and encouraged 😛

plugins/notion/src/blocksToHTML.ts
```diff
@@ -39,7 +39,7 @@ export function richTextToHTML(texts: RichTextItemResponse[]) {
 }
 
 const YOUTUBE_ID_REGEX = /(?:youtu\.be\/|youtube\.com\/(?:watch\?v=|embed\/))(?<videoId>[^?&]+)/iu
-export function blocksToHtml(blocks: BlockObjectResponse[]) {
+export function blocksToHTML(blocks: BlockObjectResponse[]) {
```
alecmev: Nit: There's no official naming convention, but, [I'd treat abbreviations/acronyms as words](https://www.approxion.com/capital-offenses-how-to-handle-abbreviations-in-camelcase/) 😛 And rename the file to `blocksToHtml` instead.

plugins/notion/src/Login.tsx
madebyisaacr: The delay was originally meant to help fix an issue with Cloudflare caching API call results. But it doesn't seem to be happening anymore, so I'll remove it.

plugins/notion/src/App.tsx
alecmev: Okay, makes sense! Then one small nit: Rename `databaseIdMap` to something like `existingCollectionDatabaseIdMap` and then `finalDatabaseIdMap` just to `databaseIdMap`, maybe? So that the shorter version is the clear choice for further usage.

plugins/notion/src/api.ts
```diff
@@ -0,0 +1,464 @@
...
+            }
+        },
+        auth: token,
+    })
+}
+
+export async function getNotionDatabases() {
+    if (!notion) {
+        initNotionClient()
+    }
```
alecmev: Turn `initNotionClient` into `getNotionClient` that just caches it in a `notionClientSingleton` variable or something like that, so that you don't need `notion!` and `assert` anymore.

plugins/notion/src/api.ts
```diff
@@ -0,0 +1,464 @@
...
+            missingCollection.push(fieldInfo)
+            continue
+        }
+
+        const isUnsupported = !Array.isArray(allowedTypes) || allowedTypes.length === 0
+        if (isUnsupported) {
+            unsupported.push(fieldInfo)
+        } else {
+            supported.push(fieldInfo)
+        }
```
alecmev: Nit: Flip the logic maybe? The fewer negations, the easier to read usually (no "un" in "unsupported").

alecmev (commented):
More comments, but still not finished, will leave more in a bit, great progress! 👌

plugins/notion/src/Login.tsx
alecmev: Oh, yeah, that kind of issues definitely should be solved on the worker side 🤔

tom-james-watson:
QA looked good until I just pulled the most recent changes. See the failing typescript check 😬

madebyisaacr:
@tom-james-watson can you invite me to the Kitchen Sink page in Notion? isaac@framer.com

plugins/notion/src/FieldMapping.tsx
```diff
@@ -0,0 +1,334 @@
...
+        collection
+            .getFields()
+            .then(collectionFields => {
+                if (abortController.signal.aborted) return
+
+                setFieldsInfo(mergeFieldsInfoWithExistingFields(initialFieldsInfo, collectionFields))
+                setStatus("mapping-fields")
+            })
+            .catch(error => {
+                if (!abortController.signal.aborted) {
```
alecmev: Nit: Early return.

plugins/notion/src/Login.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        return new Promise(resolve => {
+            const fetchTokens = () => {
+                auth.fetchTokens(readKey).then(tokens => {
+                    clearInterval(pollInterval.current)
+                    clearTimeout(pollTimeout.current)
+                    resolve(tokens)
+                })
+            }
```
alecmev: Low-hanging nit 😛 

```suggestion
            const fetchTokens = async () => {
                const tokens = await auth.fetchTokens(readKey)
                clearInterval(pollInterval.current)
                clearTimeout(pollTimeout.current)
                resolve(tokens)                
            }
```

plugins/notion/src/auth.ts
alecmev, line 1: Don't act on this comment (because I know that this is based on Airtable), but just wanted to mention that this could use a refactor 🥲 Opinions about single-use classes aside, why can't it store `readKey` as state?...

plugins/notion/src/api.ts
```diff
@@ -0,0 +1,461 @@
...
+
+/**
+ * Copied from:
+ * https://github.com/makenotion/notion-sdk-js/blob/7950edc034d3007b0612b80d3f424baef89746d9/src/helpers.ts#L47
+ * Notion has a bug where pagination returns the same page cursor when fetching
+ * another page in some rare cases. This results in the same pages being fetched
+ * over and over, resulting in infinite loop. This function is modified to keep
+ * track of which page cursors we've seen and bail out early in case the same
+ * cursor is seen twice
+ */
```
alecmev, line 334: I wonder if it's still the case in v3 🤔

plugins/notion/src/api.ts
```diff
@@ -0,0 +1,461 @@
...
+        }
+        case "email": {
+            if (field.type !== "formattedText" && field.type !== "string") return null
+
+            return { type: field.type, value: property.email ?? "" }
+        }
+    }
+
+    return null
+}
```
alecmev: Nit: Could it perhaps make more sense for this to be in `data.ts`? Did you put it here on purpose?

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,354 @@
...
+            id: database.id,
+            name: richTextToPlainText(database.title) || "Untitled Database",
+            database,
+        }
+    })
+}
+
+export function getDataSourceFieldsInfo(database: DatabaseObjectResponse, databaseIdMap: DatabaseIdMap): FieldInfo[] {
+    return getDatabaseFieldsInfo(database, databaseIdMap)
+}
```
alecmev: Does this need to exist? 😛

plugins/notion/src/data.ts
```diff
@@ -0,0 +1,354 @@
...
+                if (fieldEntry) {
+                    fieldData[field.id] = fieldEntry
+                } else {
+                    console.warn(
+                        `Skipping item at index ${index} because it doesn't have a valid value for field ${field.name}`
+                    )
+                }
+            }
+
+            if (!slugValue || typeof slugValue !== "string") {
```
alecmev: Nit: Can it be not a string if it's not `null`?

plugins/notion/src/main.tsx
```diff
@@ -1,94 +1,74 @@
...
 
-    if (!pluginContext.database) return false
-    if (!pluginContext.slugFieldId) return false
-    if (pluginContext.hasChangedFields) return false
+const root = document.getElementById("root")
+if (!root) throw new Error("Root element not found")
 
-    return true
+if (!tokens) {
+    await new Promise<void>(resolve => {
```
alecmev, line 23: Also non-actionable, comment for the future: This feels off 🤔 Having to mount twice.

alecmev (commented):
Scraping the bottom of the barrel here 😛 Pretty much just nits left. Last changes + QA, and should be good to merge, thanks!

plugins/notion/src/api.ts
madebyisaacr: Perplexity couldn't find anything saying it's been fixed, so lets keep it to be safe.
https://www.perplexity.ai/search/i-found-this-in-a-codebase-tha-Qbsj9iwHRp.AdPSYT1T6cw

plugins/notion/src/api.ts
madebyisaacr: No specific reason. This function was originally in `notion.ts` which is now `api.ts`. I moved it to `data.ts`.

plugins/notion/src/data.ts
madebyisaacr: No 🤣 it was a leftover from the CMS starter. Fixed.

plugins/notion/src/data.ts
madebyisaacr: No, slugValue is only a string or null. Fixed.

plugins/notion/src/main.tsx
alecmev: Can you do this, please? I'm doing it in Airtable.

```ts
if (!tokens) {
    const reactRoot = ReactDOM.createRoot(root)

    await new Promise<void>(resolve => {
        reactRoot.render(
            <React.StrictMode>
                <Authenticate onAuthenticated={resolve} />
            </React.StrictMode>
        )
    })

    reactRoot.unmount()
}
```

madebyisaacr:
Let's update to v4 in another PR. I'll open an issue.

alecmev (approved):
Thank you, let's merge! 😎

PR #222, opened 2025-05-05
Better CMS Starter

clementroche:
This pull request introduces several enhancements to the CMS starter, focusing on improved data modeling, field mapping, and synchronization functionality. Key changes include adding support for `collectionReference` and `multiCollectionReference` fields, introducing a more robust ID and slug handling mechanism, and making the `FieldMapping` component more flexible and user-friendly.

### Field mapping enhancements:
* Updated `FieldMapping.tsx` to include a `style` prop for customization and improved handling of missing references in `FieldMappingRow`. Missing references are now displayed as "Missing Collection" and visually disabled.
* Replaced `stroke-linecap`, `stroke-linejoin`, and `stroke-width` with their camelCase equivalents (`strokeLinecap`, `strokeLinejoin`, `strokeWidth`) for better React compatibility.

### Data synchronization improvements:
* Added `idField` and `slugField` to the `DataSource` type to standardize ID and slug handling. Introduced a `slugify` function to ensure slugs are unique and URL-safe.
* Enhanced the `syncCollection` function to handle `collectionReference` and `multiCollectionReference` fields by mapping them to the correct collection IDs, skipping items without valid IDs, and generating slugs dynamically.

### Codebase refactoring:
* Introduced the `ExtendedEditableManagedCollectionField` type to extend field capabilities, including support for `dataSourceId` and `isMissingReference`.

### DIFF
--- a/starters/cms/src/data.ts
+++ b/starters/cms/src/data.ts
@@ -11,15 +11,22 @@ export const PLUGIN_KEYS = {
     SLUG_FIELD_ID: "slugFieldId",
 } as const
 
+export type ExtendedEditableManagedCollectionField = EditableManagedCollectionField & {
+    dataSourceId?: string
+    isMissingReference?: boolean // reference to a collection that doesn't exist
+}
+
 export interface DataSource {
     id: string
-    fields: readonly EditableManagedCollectionField[]
+    fields: readonly ExtendedEditableManagedCollectionField[]
     items: FieldDataInput[]
+    idField: EditableManagedCollectionField | null // to be used as id field
+    slugField: EditableManagedCollectionField | null // to be used as slug field
 }
 
 export const dataSourceOptions = [
-    { id: "articles", name: "Articles" },
-    { id: "categories", name: "Categories" },
+    { id: "articles", name: "Articles", idFieldId: "Id", slugFieldId: "Title" },
+    { id: "categories", name: "Categories", idFieldId: "Id", slugFieldId: "Title" },
 ] as const
 
 /**
@@ -38,14 +45,72 @@ export const dataSourceOptions = [
  *   ]
  * }
  */
+
+type ExtendedManagedCollection = ManagedCollection & {
+    dataSourceId: string | null
+}
+
+let collections: ManagedCollection[] = []
+let collectionsWithDataSourceId: ExtendedManagedCollection[] = []
+
+if (framer.mode === "syncManagedCollection" || framer.mode === "configureManagedCollection") {
+    collections = await framer.getManagedCollections()
+    collectionsWithDataSourceId = await Promise.all(
+        collections.map(async collection => {
+            const dataSourceId = await collection.getPluginData(PLUGIN_KEYS.DATA_SOURCE_ID)
+            return { ...collection, dataSourceId } as ExtendedManagedCollection
+        })
+    )
+}
+
+const slugs = new Map<string, number>()
+
+function slugify(text: string) {
+    text = text.trim()
+    text = text.slice(0, 60) // limit to 60 characters
+
+    if (slugs.has(text)) {
+        const count = slugs.get(text) ?? 0
+        slugs.set(text, count + 1)
+        text = `${text} ${count + 1}`
+    } else {
+        slugs.set(text, 0)
+    }
+
+    const slug = text
+        .replace(/^\s+|\s+$/g, "")
+        .toLowerCase()
+        .replace(/[^a-z0-9 -]/g, "")
+        .replace(/\s+/g, "-")
+        .replace(/-+/g, "-")
+        .replace(/-+/g, "-")
+    return slug
+}
+
 export async function getDataSource(dataSourceId: string, abortSignal?: AbortSignal): Promise<DataSource> {
     // Fetch from your data source
     const dataSourceResponse = await fetch(`/data/${dataSourceId}.json`, { signal: abortSignal })
     const dataSource = await dataSourceResponse.json()
 
     // Map your source fields to supported field types in Framer
-    const fields: EditableManagedCollectionField[] = []
+    const fields: ExtendedEditableManagedCollectionField[] = []
     for (const field of dataSource.fields) {
+        if (field.type === "multiCollectionReference" || field.type === "collectionReference") {
+            if (!field.dataSourceId) {
+                console.warn(`No data source id found for collection reference field"${field.name}".`)
+                field.isMissingReference = true
+            } else {
+                const collection = collectionsWithDataSourceId.find(
+                    collection => collection.dataSourceId === field.dataSourceId
+                )
+
+                if (!collection) {
+                    console.warn(`No collection found for data source "${field.dataSourceId}".`)
+                    field.isMissingReference = true
+                }
+            }
+        }
+
         switch (field.type) {
             case "string":
             case "number":
@@ -54,17 +119,19 @@ export async function getDataSource(dataSourceId: string, abortSignal?: AbortSig
             case "formattedText":
             case "date":
             case "link":
+            case "collectionReference":
+            case "multiCollectionReference":
                 fields.push({
                     id: field.name,
                     name: field.name,
                     type: field.type,
+                    ...(field.dataSourceId && { dataSourceId: field.dataSourceId }),
+                    ...(field.isMissingReference && { isMissingReference: true }),
                 })
                 break
             case "image":
             case "file":
             case "enum":
-            case "collectionReference":
-            case "multiCollectionReference":
                 console.warn(`Support for field type "${field.type}" is not implemented in this Plugin.`)
                 break
             default: {
@@ -75,8 +142,15 @@ export async function getDataSource(dataSourceId: string, abortSignal?: AbortSig
 
     const items = dataSource.items as FieldDataInput[]
 
+    const dataSourceOption = dataSourceOptions.find(option => option.id === dataSourceId)
+
+    const idField = fields.find(field => field.id === dataSourceOption?.idFieldId) ?? null
+    const slugField = fields.find(field => field.id === dataSourceOption?.slugFieldId) ?? null
+
     return {
         id: dataSource.id,
+        idField,
+        slugField,
         fields,
         items,
     }
@@ -98,13 +172,32 @@ export function mergeFieldsWithExistingFields(
 export async function syncCollection(
     collection: ManagedCollection,
     dataSource: DataSource,
-    fields: readonly EditableManagedCollectionField[],
+    fields: readonly ExtendedEditableManagedCollectionField[],
     slugField: EditableManagedCollectionField
 ) {
-    const sanitizedFields = fields.map(field => ({
-        ...field,
-        name: field.name.trim() || field.id,
-    }))
+    const sanitizedFields = fields
+        .map(field => {
+            if (
+                (field.type === "multiCollectionReference" || field.type === "collectionReference") &&
+                field.dataSourceId
+            ) {
+                const collectionId = collectionsWithDataSourceId.find(
+                    collection => collection.dataSourceId === field.dataSourceId
+                )?.id
+                if (collectionId) {
+                    field.collectionId = collectionId
+                } else {
+                    console.warn(`No collection found for data source "${field.dataSourceId}".`)
+                    return null
+                }
+            }
+
+            return {
+                ...field,
+                name: field.name.trim() || field.id,
+            }
+        })
+        .filter(field => field !== null)
 
     const items: ManagedCollectionItemInput[] = []
     const unsyncedItems = new Set(await collection.getItemIds())
@@ -116,6 +209,12 @@ export async function syncCollection(
         const slugValue = item[slugField.id]
         if (!slugValue || typeof slugValue.value !== "string") {
             console.warn(`Skipping item at index ${i} because it doesn't have a valid slug`)
+            continue
+        }
+
+        const idValue = item[dataSource.idField?.id ?? ""]
+        if (!idValue || typeof idValue.value !== "string") {
+            console.warn(`Skipping item at index ${i} because it doesn't have a valid id`)
             continue
         }
 
@@ -134,8 +233,8 @@ export async function syncCollection(
         }
 
         items.push({
-            id: slugValue.value,
-            slug: slugValue.value,
+            id: idValue.value,
+            slug: slugify(slugValue.value),
             draft: false,
             fieldData,
         })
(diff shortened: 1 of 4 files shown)


starters/cms/src/data.ts
```diff
@@ -174,7 +248,12 @@ export async function syncExistingCollection(
             return { didSync: false }
         }
 
-        await syncCollection(collection, dataSource, existingFields, slugField)
+        await syncCollection(
+            collection,
+            dataSource,
+            existingFields as ManagedCollectionFieldInput[],
+            slugField as ManagedCollectionFieldInput
```
jonastreub, line 255: Why do you need to cast the types here? Casts are dangerous because you are overriding something which might not be true

starters/cms/src/data.ts
```diff
@@ -38,14 +40,72 @@ export const dataSourceOptions = [
...
+
+    const slug = text
+        .replace(/^\s+|\s+$/g, "")
+        .toLowerCase()
+        .replace(/[^a-z0-9 -]/g, "")
+        .replace(/\s+/g, "-")
+        .replace(/-+/g, "-")
+        .replace(/-+/g, "-")
+    return slug
+}
```
jonastreub, line 83: A `slugify` function should not have side effects like making something unique. If someone starts reusing this function for something different it could easily lead to bugs

starters/cms/src/data.ts
```diff
@@ -38,14 +40,72 @@ export const dataSourceOptions = [
...
+
+let collections: ManagedCollection[] = []
+let collectionsWithDataSourceId: ExtendedManagedCollection[] = []
+
+if (framer.mode === "syncManagedCollection" || framer.mode === "configureManagedCollection") {
+    collections = await framer.getManagedCollections()
+    collectionsWithDataSourceId = await Promise.all(
+        collections.map(async collection => {
+            const dataSourceId = await collection.getPluginData(PLUGIN_KEYS.DATA_SOURCE_ID)
+            return { ...collection, dataSourceId } as ExtendedManagedCollection
```
jonastreub, line 56: This cast is dangerous. `dataSourceId` should be type checked

starters/cms/src/data.ts
```diff
@@ -38,14 +40,72 @@ export const dataSourceOptions = [
  *   ]
  * }
  */
+
+type ExtendedManagedCollection = ManagedCollection & {
+    dataSourceId: string | null
```
jonastreub, line 45: Isn't `dataSourceId` the same as the `id` of the collection?

clementroche: `id` is generated by framer-plugin itself meanwhile `dataSourceId` is relative to data.

`const dataSourceId = await collection.getPluginData(PLUGIN_KEYS.DATA_SOURCE_ID)`

starters/cms/src/data.ts
clementroche: I removed the typecasting

starters/cms/src/data.ts
clementroche: I've fixed it by copying the `text` variable to another one to make it safer.

starters/cms/src/data.ts
clementroche: I've removed the typecasting for `slugField`. But I struggle with `existingFields`, I might need some help since those types are not documented.

starters/cms/src/data.ts
jonastreub: Should we change the name to something like `externalDatasourceID`?

starters/cms/src/data.ts
jonastreub: Strings are value types in JS. So they can't be mutated. That's not the issue. The issue is that a `slugify` function should not add suffixes

jonastreub: This is our implementation btw:

```ts
// Note: We don't use the "slugify" package here because we want a very specific
// behavior for our CMS and web pages. Make sure to pick the right slugify
// function for your use case!

// Match everything except for letters, numbers and parentheses.
const nonSlugCharactersRegExp = /[^\p{Letter}\p{Number}()]+/gu
// Match leading/trailing dashes, for trimming purposes.
const trimSlugRegExp = /^-+|-+$/gu

/**
 * Takes a freeform string and removes all characters except letters, numbers,
 * and parentheses. Also makes it lower case, and separates words by dashes.
 * This makes the value URL safe.
 */
    return value.toLowerCase().replace(nonSlugCharactersRegExp, "-").replace(trimSlugRegExp, "")
}
```

starters/cms/src/data.ts
clementroche: I think that not an issue anymore since i've refactored how to retrieve collection

starters/cms/src/data.ts
clementroche: We need to make slugs unique at some point in case users select a field that is not unique.

jonastreub: It's very easy to use the function in a wrong way. For example when someone starts building an actual plugin and every time they fetch all items and make sure all slugs are valid using slugify. Every sync all slugs will change

jonastreub: Even if you name it `uniqueSlugify`, it will be very easy to use it wrong

jonastreub: And that mostly means the function is too magic — as in having unexpected side effects.

clementroche: Valid point, `slugs` map should be reset after batched usage then?

jonastreub: Indeed. I would create two functions `slufigy` and `createUniqueSlug`. `slugify` only does one job. And `createUniqueSlug` takes both the raw string value and a Set with taken slugs.

starters/cms/src/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (existingSlugs.has(text)) {
+        const count = existingSlugs.get(text) ?? 0
+        existingSlugs.set(text, count + 1)
+        text = `${text} ${count + 1}`
+    } else {
+        existingSlugs.set(text, 0)
+    }
+
+    return slugify(text)
+}
```
jonastreub, line 22: This can still result in duplicate slugs, because we check the input instead of the generated slug

jonastreub:
Why would you want to show the id? This is typically hidden from users

starters/cms/src/data.ts
```diff
@@ -168,17 +206,17 @@ export async function syncExistingCollection(
...
         if (!slugField) {
-            framer.notify(`No field matches the slug field id “${previousSlugFieldId}”. Sync will not be performed.`, {
+            framer.notify(`No field matches the slug field id "${previousSlugFieldId}". Sync will not be performed.`, {
                 variant: "error",
             })
             return { didSync: false }
         }
 
-        await syncCollection(collection, dataSource, existingFields, slugField)
+        await syncCollection(collection, dataSource, existingFields as ManagedCollectionFieldInput[], slugField)
```
jonastreub, line 215: This `as ManagedCollectionFieldInput[]` cast is a bad habit, so should definitely not be in the starter

starters/cms/src/data.ts
```diff
@@ -168,17 +206,17 @@ export async function syncExistingCollection(
 
         const slugField = dataSource.fields.find(field => field.id === previousSlugFieldId)
         if (!slugField) {
-            framer.notify(`No field matches the slug field id “${previousSlugFieldId}”. Sync will not be performed.`, {
+            framer.notify(`No field matches the slug field id "${previousSlugFieldId}". Sync will not be performed.`, {
```
jonastreub, line 209: Are these quotes changed on purpose?

starters/cms/src/data.ts
clementroche: This is required since `collection.getField()` returns `ManagedCollectionField[]` meawnhile `collection.setFields` expects `ManagedCollectionFieldInput[]`

jonastreub: Casts are never the solution though. It silences an actual issue.

jonastreub: Of course you can use casts in your private code but this is an example where the code quality should be really high

jonastreub: Btw, the fact that it currently errors is really bad… but adding a cast only hides the actual issue that we should fix

clementroche: I fixed it:

``` js
const fields: ManagedCollectionFieldInput[] = []
        for (const field of existingFields) {
            if (field.type === "multiCollectionReference" || field.type === "collectionReference") {
                fields.push({
                    id: field.id,
                    name: field.name,
                    type: field.type,
                    collectionId: field.collectionId,
                })
            } else if (field.type === "enum") {
                fields.push({
                    id: field.id,
                    name: field.name,
                    type: field.type,
                    cases: [],
                })
            } else if (field.type === "file" || field.type === "image") {
                fields.push({
                    id: field.id,
                    name: field.name,
                    type: field.type,
                    allowedFileTypes: [],
                })
            } else {
                fields.push({
                    id: field.id,
                    name: field.name,
                    type: field.type,
                })
            }
        }
  ```

File: starters/cms/src/utils.ts (hunk 5 above)
Comment (clementroche) [reply starters/cms/src/utils.ts]: True, i've made it more robust now. First, input get slugified then function makes it unique.
https://github.com/framer/plugins/pull/222/commits/a3e3b8c8e4b6ff9511f5bda6152ba3a8e0641bc5#diff-e1867fda29cfd75d20d909c3895b73c2878e32c5c28bc70912a893251891c2a1

**Comment by clementroche** [comment]
> Why would you want to show the id? This is typically hidden from users

I needed a way to define what field will be the ID since it's necessary to keep it consistent for mapping `multiCollectionReference` and `collectionReference` correctly.  Currently the data cms starter is using is not realistic, maybe that's why it's not obvious to you. All the stuff i'm bringing here comes from my experience developing CMS Plugins at the same time.

File: starters/cms/src/utils.ts (hunk 5 above)
Comment (jonastreub) [reply starters/cms/src/utils.ts]: The function still has bugs. And I'm wondering if automatically making it unique is better than telling the user about the duplicate slug so they can fix it themselves. Having a duplicate slug is probably a mistake and if we don't warn them you will publish a URL that you will probably need to change once you see the mistake.

Comment (jonastreub) [reply starters/cms/src/utils.ts]: The bug is that we don't check the final generated slug for uniqueness. So if my data source has the following slugs it will result in duplicates:
1. `my-slug`
2. `my-slug-0`
3. `my-slug` → transformed into `my-slug-0` which is a duplicate

File: starters/cms/src/data.ts [hunk 8]
```diff
@@ -174,7 +212,39 @@ export async function syncExistingCollection(
...
+                    name: field.name,
+                    type: field.type,
+                    cases: [],
+                })
+            } else if (field.type === "file" || field.type === "image") {
+                fields.push({
+                    id: field.id,
+                    name: field.name,
+                    type: field.type,
+                    allowedFileTypes: [],
```
jonastreub, line 236: The empty arrays are incorrect

jonastreub:
> ids can change depending which API you're fetching

Can you be more specific?

The ids are supposed to be coming from your external data source. So in theory you would not need to read anything from Framer data

starters/cms/src/data.ts
clementroche: Yo, sorry it's a bit hard to solve this since i can't find documentation, but here is [the commit](https://github.com/framer/plugins/pull/222/commits/f7482e4f80c6195822e56605d50b3e4e923c95d6)

clementroche:
> Can you be more specific?

Here is an example using Greenhouse Job Board API:

Jobs:
```
{
    "jobs": [
        {
            "absolute_url": "https://job-boards.greenhouse.io/vaulttec/jobs/260533",
            "data_compliance": [
                {
                    "type": "gdpr",
                    "requires_consent": false,
                    "requires_processing_consent": false,
                    "requires_retention_consent": false,
                    "retention_period": null,
                    "demographic_data_consent_applies": false
                }
            ],
            "education": "education_optional",
            "internal_job_id": 287334,
            "location": {
                "name": "NYC"
            },
            "metadata": null,
            "id": 260533,
            "updated_at": "2025-04-11T16:16:09-04:00",
            "requisition_id": "3",
            "title": "Awkward Cop",
            "company_name": "Vault-Tec Corporation",
            "first_published": null,
            "content": "&lt;p&gt;The Rule Enforcement team is dedicated to keeping all employees in line. &amp;nbsp;Rule enforcers use&amp;nbsp;various tactics such as physical intimidation, awkward pauses, and dramatic coffee sips in order to make sure everyone does what they&#39;re told.&lt;/p&gt;\n&lt;p&gt;As a Good Cop, you&#39;ll work closely with Bad Cops in order to extract information from employees. Don&#39;t let the name of the role fool you however -- you are going to be just as &#39;bad&#39; as Bad Cops, but you&#39;ll preferably have experience playing the part of the kind, compassionate and fair cop.&lt;/p&gt;",
            "departments": [
                {
                    "id": 13580,
                    "name": "Finance",
                    "child_ids": [],
                    "parent_id": null
                }
            ],
            "offices": []
        },
    ],
    "meta": {
    "total" : 5
  }
}
```

Departments:
```
{
    "departments": [
        {
            "id": 13580,
            "name": "Finance",
            "parent_id": null,
            "child_ids": [],
            "jobs": [
                {
                    "absolute_url": "https://job-boards.greenhouse.io/vaulttec/jobs/260533",
                    "data_compliance": [
                        {
                            "type": "gdpr",
                            "requires_consent": false,
                            "requires_processing_consent": false,
                            "requires_retention_consent": false,
                            "retention_period": null,
                            "demographic_data_consent_applies": false
                        }
                    ],
                    "education": "education_optional",
                    "internal_job_id": 287334,
                    "location": {
                        "name": "NYC"
                    },
                    "metadata": null,
                    "id": 260533,
                    "updated_at": "2025-04-11T16:16:09-04:00",
                    "requisition_id": "3",
                    "title": "Awkward Cop",
                    "company_name": "Vault-Tec Corporation",
                    "first_published": null
                }
            ]
        },
    ]
}
```

So in Framer CMS it will be two collections: Jobs & Departements where a Job reference to Departements through `multiCollectionReference` departments field. To retrieve these reference departements from a job I (as a CMS Starter user) need to pass the id of the departement entry. So at some point I have to define which field will be the id and this will remain. 

BTW the end user of the plugin will never have to choose what's the id field. It's a choice that the CMS starter user will do to ensure data coherence depending the data that needs to be mapped.

You can check my [Greenhouse plugin](https://github.com/darkroomengineering/framer-plugins/tree/greenhouse/plugins/greenhouse) that has been built on top of CMS starter. And more specifically [data source folder](https://github.com/darkroomengineering/framer-plugins/tree/greenhouse/plugins/greenhouse/src/data-source).

triozer:
It will be easier to create a new PR after Greenhouse is merged

PR #229, opened 2025-05-30
Plugin Permissions update

triozer:
### Description

This PR updates the Framer plugins to use the latest versions of dependencies, particularly upgrading `framer-plugin` to version 3.2.0 and `vite-plugin-framer` to version 1.0.7. It also adds permission checks throughout the plugins to ensure they respect user permissions when performing actions like adding images, setting attributes, or modifying collections.

Key changes:
- Upgraded `framer-plugin` from v2.x to v3.2.0 across all plugins
- Updated `vite-plugin-framer` to v1.0.7 for consistency
- Added permission checks using `useIsAllowedTo` and `isAllowedTo` hooks
- Disabled UI elements when users don't have sufficient permissions
- Added helpful tooltips to explain why actions might be disabled

### Testing

- [x] Verify permission checks work correctly
  - [x] Test plugins with a user that has full permissions
  - [x] Test plugins with a user that has limited permissions
  - [x] Confirm disabled buttons show appropriate tooltips


triozer:
* **#229** <a href="https://app.graphite.dev/github/pr/framer/plugins/229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/framer/plugins/229?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

plugins/color-extract/src/App.tsx
```diff
@@ -105,11 +110,10 @@ export function App() {
                 <>
                     <div className="interface">{colorList}</div>
                     <button
-                        disabled={!supportsGradient}
+                        disabled={!supportsGradient || !isAllowedToSetAttributes}
                         style={{ opacity: supportsGradient ? 1 : 0.5 }}
-                        onClick={() => {
-                            setAsGradient(currentSelection, colors)
-                        }}
+                        onClick={() => setAsGradient(currentSelection, colors)}
```
alecmev: Nit: I'd keep it wrapped, it's a good idea to not return when you don't mean to return 😌 (because in the future `setAsGradient` might start returning something other than `void`, and `onClick` might have special behavior for those cases, resulting in a potential bug)

plugins/design-system/src/App.tsx
```diff
@@ -272,12 +285,22 @@ export function App() {
...
+                                >
                                     Add
                                 </button>
                             </div>
                         ))}
-                        <button className="action-button" onClick={() => handleAddColors(filteredColorItems)}>
+                        <button
+                            className="action-button"
+                            disabled={!isAllowedToUpsertColor}
+                            title={isAllowedToUpsertColor ? undefined : "Insufficient permissions"}
```
alecmev: I guess this could be a helper, since it's done more than once here, something like this:

```ts
const getPermissionTitle = (isAllowed) => isAllowed ? undefined : "Insufficient permissions"
```

plugins/dither/src/App.tsx
```diff
@@ -39,6 +39,9 @@ export interface DroppedAsset {
 }
 
 function DitherImage({ image }: { image: ImageAsset | null }) {
+    const isAllowedToSetImage = useIsAllowedTo("setImage")
+    const isAllowedToUploadImage = useIsAllowedTo("addImage")
```
alecmev: In this plugin, is it ever useful only with one of these being `true`? If not, maybe combine them into a single boolean?

triozer: I was exactly thinking of that + one always comes with the other so yeah, let's combine them

plugins/ascii/src/App.tsx
```diff
@@ -53,8 +53,7 @@ function ASCIIPlugin({ framerCanvasImage }: { framerCanvasImage: ImageAsset | nu
     const asciiRef = useRef()
     const { gl, toBytes, program, setProgram, setResolution } = useOGLPipeline()
 
-    const isAllowedToSetImage = useIsAllowedTo("setImage")
-    const isAllowedToUploadImage = useIsAllowedTo("addImage")
+    const isAllowedToUpsertImage = useIsAllowedTo("addImage") && useIsAllowedTo("setImage")
```
alecmev: 😉 

```suggestion
    const isAllowedToUpsertImage = useIsAllowedTo("addImage", "setImage")
```

triozer: Forgot about this haha, API is 💪🏽 !!!

triozer:
Actually for CSV Import, we are safe, we auto-close the plugin in case the collection is `readonly` (e.g. the user don't have perm or it's a managed collection)

plugins/unsplash/package.json
```diff
@@ -38,6 +38,6 @@
         "tailwindcss": "^3.4.1",
         "typescript": "^5.4.5",
         "vite": "^6",
-        "vite-plugin-framer": "^1.0.1"
```
alecmev, line 41: This is something I've been seeing Hunter do too 😁 I think it's not a good idea. There is a lockfile, but in the absence of it, `^1` might resolve to `1.0.0`, while `^1.0.1` is guaranteed to be `1.0.1+`, which might have a crucial API/typing fix, which is broken in `1.0.0`. I think it's always good to keep `package.json` as specific as possible, in-sync with the lockfile.

triozer: Yeah, ok, I updated to use "^1.0.7" explicitly

package-lock.json
alecmev, line 1: Can you, please, go through this lockfile and ensure that every `"resolved"` for `framer-plugin` is `https://registry.npmjs.org/framer-plugin/-/framer-plugin-3.2.0.tgz`? There are a few that are `3.0.0` and one that is `2.x`.

plugins/ascii/src/utils.ts
```diff
@@ -45,3 +45,7 @@ export function assert(condition: unknown, ...msg: unknown[]): asserts condition
     }
     throw e
 }
+
+export function getPermissionTitle(isAllowed: boolean) {
```
alecmev: Nit:

```suggestion
export function getPermissionTitle(isAllowed: boolean): string | undefined {
```

plugins/color-extract/src/App.tsx
```diff
@@ -92,7 +96,8 @@ export function App() {
                     restSpeed: 0.0001,
                 }}
                 onClick={() => {
-                    handleOnClick(currentSelection, color.hex)
+                    if (!isAllowedToSetAttributes) return
```
alecmev, line 103: This part could be baked into `handleOnClick`, right?

plugins/csv-import/src/App.tsx
```diff
@@ -215,10 +232,11 @@ export function App({ collection }: { collection: Collection }) {
             form.current?.removeEventListener("dragleave", handleDragLeave)
             form.current?.removeEventListener("drop", handleDrop)
         }
-    }, [])
+    }, [isAllowedToAddItems])
 
     useEffect(() => {
         const handlePaste = async (event: ClipboardEvent) => {
+            if (!isAllowedToAddItems) return
```
alecmev: Could move one line up maybe?

plugins/design-system/src/App.tsx
```diff
@@ -236,6 +251,8 @@ export function App() {
                             <button
                                 key={section.key}
                                 className="section-button"
+                                disabled={!isAllowedToAddDetachedComponentLayers}
```
alecmev, line 255: Does this also disable the nested `Draggable`?

plugins/phosphor/src/App.tsx
```diff
@@ -104,6 +108,21 @@ function IconGrid(props: any) {
...
+                        title={isAllowedToAddSVG ? undefined : "Insufficient permissions"}
+                    >
+                        <Icon size={32} color={"var(--framer-color-text)"} weight={weight} />
+                    </button>
+                )
+
+                if (!isAllowedToAddSVG) {
+                    return button
+                }
+
```
alecmev: `Draggable` isn't possible to disable?

plugins/pong/src/App.tsx
```diff
@@ -152,7 +154,12 @@ export function App() {
                     Stop Game
                 </button>
             ) : (
-                <button className="framer-button-primary" onClick={setupFromSelection}>
+                <button
+                    className="framer-button-primary"
+                    onClick={setupFromSelection}
```
alecmev: Not adding an early return to the handler too? I'm not sure if it's absolutely necessary, might actually make sense to omit it in the other plugins, disabling the ability to ever trigger the handler should be enough.

plugins/renamer/src/app.svelte
```diff
@@ -16,6 +16,8 @@
   import { renameResult } from "./search/result_processors/rename_result";
   import { cleanUpResult } from "./search/result_processors/clean_up_result";
 
+  const isAllowedToSetAttributes = framer.isAllowedTo("Node.setAttributes")
```
alecmev: Not reactive 🥲

plugins/rss-feeds/src/main.tsx
```diff
@@ -6,8 +6,20 @@ import ReactDOM from "react-dom/client"
 import { App, importData } from "./App.tsx"
 
 async function runPlugin() {
+    const isAllowedToImportData = framer.isAllowedTo(
+        "ManagedCollection.addItems",
+        "ManagedCollection.removeItems",
+        "ManagedCollection.setFields",
+        "ManagedCollection.setPluginData"
+    )
```
alecmev: Why not leave it to the UI? 🤔

plugins/unsplash/src/App.tsx
```diff
@@ -238,6 +257,38 @@ const GridItem = memo(function GridItem({ photo, loading, height, width, onSelec
...
+                <div className="absolute top-0 left-0">
+                    <Blurhash hash={photo.blur_hash} width={width} height={height} />
+                </div>
+            )}
+        </button>
+    )
+
+    if (!isAllowedToUpsertImage) {
+        return button
+    }
```
alecmev: Same question as the other plugin.

alecmev (changes requested):
Great stuff, thank you again! 😎 Just some minor comments.

package-lock.json
triozer: Updated the CMS Export. For the rest, `2.x` are other CMS related plugins, `3.0.0` is Airtable and `3.1.0` is the CMS Starter

plugins/design-system/src/App.tsx
triozer: Actually it seems like yes!

plugins/phosphor/src/App.tsx
triozer: No, there is no way yet, but seems like disabling the button will also disable drag.

plugins/rss-feeds/src/main.tsx
triozer: Oups, this should have been under! Sync happens without UI, there is also checks in the UI when creating.

plugins/pong/src/App.tsx
triozer: I will add to the handler too, the idea is we want to be safe against people removing the disabled prop.

alecmev: Well, if somebody removes the disabled prop, they can also remove the early return in the handler, haha 😄

plugins/renamer/src/app.svelte
triozer: Shall I make it reactive? 👀

alecmev: I think so, no? Like `useIsAllowedTo` but for Svelte. If you feed the source code of the React hook into an LLM it should be able to come up with something reasonable, probably?

triozer: Done!

plugins/renamer/src/app.svelte
```diff
@@ -16,6 +16,16 @@
...
 
+  let isAllowedToSetAttributes = $state(framer.isAllowedTo("Node.setAttributes"));
+
+  $effect(() => {
+    const unsubscribe = framer.subscribeToIsAllowedTo("Node.setAttributes", (newIsAllowed) => {
+      isAllowedToSetAttributes = newIsAllowed;
+    });
+
+    return unsubscribe;
+  });
```
alecmev, line 27: Nit: Put `"Node.setAttributes"` in a shared constant, maybe?

plugins/tidyup/package.json
```diff
@@ -12,7 +12,7 @@
         "typecheck": "tsc --noEmit"
     },
     "dependencies": {
-        "framer-plugin": "3.3.0-alpha.11",
+        "framer-plugin": "^3.3.0-beta.1",
```
alecmev, line 15: Hmm. We didn't release the beta stuff in 3.3?

alecmev (approved):
LGTM, but let's wait for 3.3.2 and then merge!

plugins/tidyup/package.json
triozer: We did in: https://www.framer.com/developers/changelog#3.3.0! Yeah I'll update again for 3.3.2 haha

triozer: Done

plugins/doodles/package.json
```diff
@@ -17,12 +17,12 @@
...
         "extract-colors": "^4.0.2",
         "framer-motion": "^11.0.18",
-        "framer-plugin": "^3.3.1",
+        "framer-plugin": "^3.3.2",
         "react": "^18",
         "react-color": "^2.19.3",
         "react-dom": "^18",
         "react-sketch-canvas": "^6.2.0",
-        "svgo": "^3.3.1",
+        "svgo": "^3.3.2",
```
alecmev, line 25: Mmmm 😁

triozer: hahahaha

triozer: Well lucky us there is a `3.3.2` out :shipit:

alecmev: Funnily... `3.3.1` isn't, so this is an accidental bugfix 😂

PR #232, opened 2025-06-10
Use a shared universal `tsconfig.json`

alecmev:
### Description

The more reuse, the better.

### Testing

- [x] Select a couple of plugins at random and run `npx tsc` in them and see if everything is okay.


triozer (approved):
LGTM

PR #235, opened 2025-06-10
Add GitHub CI workflow for typechecking

triozer:
### Description

This pull request adds a GitHub Actions CI workflow to automatically run type checking on the codebase. The workflow is triggered on pushes to the main branch and pull requests targeting main. It sets up a Node.js environment, installs dependencies, and runs the typecheck script to ensure type safety.

> [!TIP]
> Use the "Hide whitespace" option for better readability! 
> image

### QA

- [x] CI is happy

### DIFF
--- a/plugins/ascii/src/dropzone/drag-and-drop.tsx
+++ b/plugins/ascii/src/dropzone/drag-and-drop.tsx
@@ -1,6 +1,6 @@
 import { useState } from "react"
 import Dropzone from "react-dropzone"
-import { type DroppedAsset } from "../App"
+import type { DroppedAsset } from "../App"
 import { getPermissionTitle } from "../utils"
 
 export function Upload({
--- a/plugins/ascii/src/number-input.tsx
+++ b/plugins/ascii/src/number-input.tsx
@@ -55,11 +55,10 @@ export function NumberInput({
                 }}
                 onKeyDown={e => {
                     if (e.key === "Enter" || e.key === "Tab" || e.key === "ArrowUp" || e.key === "ArrowDown") {
-                        const target = e.target as HTMLInputElement
-                        const value = sanitize(target.value)
+                        const value = sanitize(e.currentTarget.value)
 
                         onValueChange?.(value)
-                        target.value = value.toString()
+                        e.currentTarget.value = value.toString()
                     }
                 }}
             />
--- a/plugins/dither/src/dropzone/drag-and-drop.tsx
+++ b/plugins/dither/src/dropzone/drag-and-drop.tsx
@@ -1,6 +1,6 @@
 import { forwardRef, useImperativeHandle, useRef, useState } from "react"
 import Dropzone from "react-dropzone"
-import { type DroppedAsset } from "../App"
+import type { DroppedAsset } from "../App"
 import { getPermissionTitle } from "../utils"
 
 export const Upload = forwardRef(function Upload(
--- a/plugins/dither/src/inputs/number-input.tsx
+++ b/plugins/dither/src/inputs/number-input.tsx
@@ -55,11 +55,10 @@ export function NumberInput({
                 }}
                 onKeyDown={e => {
                     if (e.key === "Enter" || e.key === "Tab" || e.key === "ArrowUp" || e.key === "ArrowDown") {
-                        const target = e.target as HTMLInputElement
-                        const value = sanitize(target.value)
+                        const value = sanitize(e.currentTarget.value)
 
                         onValueChange?.(value)
-                        target.value = value.toString()
+                        e.currentTarget.value = value.toString()
                     }
                 }}
             />


.vscode/settings.json
```diff
@@ -1,4 +1,7 @@
 {
     "editor.rulers": [120],
-    "eslint.workingDirectories": [{ "mode": "auto" }]
+    "eslint.workingDirectories": [{ "mode": "auto" }],
+    "editor.codeActionsOnSave": {
```
triozer, line 4: Drive-by: I don't have a strong opinion

alecmev: Can we delegate this to Biome? I assume this uses the TypeScript thing?

plugins/airtable/package.json
```diff
@@ -8,7 +8,8 @@
         "build": "vite build",
         "lint": "eslint . --max-warnings 0",
         "preview": "vite preview",
-        "pack": "npx framer-plugin-tools@latest pack"
+        "pack": "npx framer-plugin-tools@latest pack",
+        "typecheck": "tsc --noEmit"
```
alecmev: Nit: We have `noEmit` in all configs 😌 

```suggestion
        "typecheck": "tsc"
```

plugins/ascii/src/App.tsx
```diff
@@ -63,6 +63,7 @@ function ASCIIPlugin({ framerCanvasImage }: { framerCanvasImage: ImageAsset | nu
         gl,
         droppedAsset?.src ? (droppedAsset?.type === "image" ? droppedAsset?.src : undefined) : framerCanvasImage?.url,
         texture => {
+            // @ts-expect-error - TODO: not sure why this is needed
```
alecmev, line 66: I don't understand how it's supposed to work, because there's no mention of `texture` in `Program`: https://github.com/oframe/ogl/blob/master/src/core/Program.js And nobody reads it. Try removing and see if it's till okay?

plugins/ascii/src/App.tsx
```diff
@@ -212,6 +217,7 @@ function ASCIIPlugin({ framerCanvasImage }: { framerCanvasImage: ImageAsset | nu
             <div className={cn("gui")}>
                 <ASCII
                     ref={node => {
+                        if (!node) return
```
alecmev, line 220: Is the case of it being `null` invalid?

plugins/ascii/src/dropzone/drag-and-drop.tsx
```diff
@@ -1,6 +1,6 @@
-import Dropzone from "react-dropzone"
 import { useState } from "react"
-import { DroppedAsset } from "../App"
+import Dropzone from "react-dropzone"
+import { type DroppedAsset } from "../App"
```
alecmev: Nit:

```suggestion
import type { DroppedAsset } from "../App"
```

plugins/ascii/src/hooks/use-glb-texture.tsx
```diff
@@ -1,17 +1,19 @@
-import { useCallback, useEffect, useRef, useState } from "react"
+// @ts-nocheck - TODO: fix this
```
alecmev, line 1: Is it that bad? 🥲 Just would be preferable to avoid these todos, realistically there's a 1% chance they'll be acted on.

plugins/ascii/src/number-input.tsx
```diff
@@ -55,10 +55,11 @@ export function NumberInput({
                 }}
                 onKeyDown={e => {
                     if (e.key === "Enter" || e.key === "Tab" || e.key === "ArrowUp" || e.key === "ArrowDown") {
-                        const value = sanitize(e.target.value)
+                        const target = e.target as HTMLInputElement
```
alecmev: How come?

plugins/cms-export/src/PreviewTable.tsx
```diff
@@ -32,7 +32,10 @@ export function PreviewTable({ collection }: Props) {
             const previewItems: CollectionItem[] = []
 
             for (let index = 0; index < items.length; index++) {
-                previewItems.push(items[index])
+                const item = items[index]
+                if (!item) continue
```
alecmev: `for-of` maybe?

plugins/csv-import/src/csv.ts
```diff
@@ -303,22 +312,30 @@ export async function processRecords(collection: Collection, records: CSVRecord[
...
     for (const record of records) {
         let slug: string | undefined
         const values = Object.values(record)
 
         // Try to get slug from the slug field first
         if (slugIndex !== -1) {
-            slug = slugify(values[slugIndex])
+            const slugValue = values[slugIndex]
+            if (!isDefined(slugValue)) {
+                result.warnings.missingSlugCount++
```
alecmev: We should get Biome on this 😛 

```suggestion
                result.warnings.missingSlugCount += 1
```

plugins/csv-import/src/csv.ts
```diff
@@ -467,3 +484,7 @@ const trimSlugRegExp = /^-+|-+$/gu
 export function slugify(value: string): string {
     return value.toLowerCase().replace(nonSlugCharactersRegExp, "-").replace(trimSlugRegExp, "")
 }
+
+function isDefined<T>(value: T | null | undefined): value is T {
+    return value !== null && value !== undefined
+}
```
alecmev: Since this is a new addition, it's something that has come up with Adam before, we should be leaning toward `isUndefined` and `isNull` (so, `===` checks), and then combining/negating them as needed. There have been cases where the zoo of these helpers ultimately led to misconceptions and bugs, reportedly.

plugins/google-search-console/package.json
```diff
@@ -8,7 +8,8 @@
     "build": "VITE_OAUTH_API_DOMAIN=https://oauth.fetch.tools/google-search-console-plugin vite build --base=${PREFIX_BASE_PATH:+/$npm_package_name}/",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "pack": "npx framer-plugin-tools@latest pack",
-    "preview": "vite preview"
+    "preview": "vite preview",
+        "typecheck": "tsc --noEmit"
```
alecmev: Spaces? 🤔

plugins/notion/src/notion.ts
```diff
@@ -630,6 +635,8 @@ export function hasFieldConfigurationChanged(a: ManagedCollectionField[], b: Man
         const fieldA = a[i]
         const fieldB = b[i]
 
+        if (!fieldA || !fieldB) return true
```
alecmev: Mm, this should be an assert probably?

plugins/notion/src/utils.ts
```diff
@@ -73,7 +73,8 @@ export function generateRandomId() {
 
     let id = ""
     for (let i = 0; i < array.length; i++) {
-        id += array[i].toString(16).padStart(2, "0")
+        const value = array.at(i) ?? 0
```
alecmev: `for-of` probably better then, or an assert?

plugins/threshold/src/worker/worker.ts
```diff
@@ -33,7 +33,11 @@ export class CanvasWorker {
         const imgData = imageData.data
 
         for (let i = 0; i < imgData.length; i += 4) {
-            const grayscale = imgData[i] * 0.3 + imgData[i + 1] * 0.59 + imgData[i + 2] * 0.11
+            let r = imgData[i] ?? 0
+            let g = imgData[i + 1] ?? 0
+            let b = imgData[i + 2] ?? 0
+
```
alecmev: Should also be non-`undefined` asserts, IMO 🤔

alecmev:
Airtable changes won't be needed, by the way, if the permission PR is merged first: https://github.com/framer/plugins/pull/237

.vscode/settings.json
triozer: We don't use Biome in this repo (yet?)

plugins/ascii/src/number-input.tsx
triozer: I guess we could use `currentTarget` instead

plugins/ascii/src/hooks/use-glb-texture.tsx
triozer: Yeah... A lot of unknown about `ogl` and the types here. I don't like the todos but seems out-of-scope.

PR #237, opened 2025-06-10
Check permissions in Airtable plugin

alecmev:
### Description

Fixes: https://github.com/framer/company/issues/32801

### Testing

- [x] Steps:

1. Make sure both "Plugin Permissions" and "Plugin Delayed Toast Clearing" experiments are on.
1. Mess around, import a collection, sync, manage, confirm it's all good.
1. Confirm that no toasts appear, talking about unchecked permissions.
1. Take away "Design" and/or "Content" permissions.
1. Synchronization should still work with "Design" disabled.
1. Management should require both.
1. The UI should react to the changes in the permissions.

### DIFF
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -2,9 +2,9 @@ import type { ManagedCollection, ManagedCollectionFieldInput, Field } from "fram
 import type { DataSource } from "./data"
 import type { PossibleField } from "./fields"
 
-import { framer } from "framer-plugin"
+import { framer, useIsAllowedTo } from "framer-plugin"
 import { useState, useEffect, memo } from "react"
-import { mergeFieldsWithExistingFields, syncCollection } from "./data"
+import { mergeFieldsWithExistingFields, syncCollection, syncMethods } from "./data"
 import { ALLOWED_FILE_TYPES, isCollectionReference } from "./utils"
 
 function ChevronIcon() {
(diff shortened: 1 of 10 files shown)


plugins/airtable/src/FieldMapping.tsx
```diff
@@ -244,8 +255,20 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
+                    name: field.name.trim() || field.id,
+                }))
+                .filter(field => field.type !== "unsupported")
+                .filter(
+                    field =>
+                        (field.type !== "collectionReference" && field.type !== "multiCollectionReference") ||
+                        field.collectionId !== ""
+                )
+
+            await collection.setFields(fieldsToSync)
```
triozer, line 271: This should always happen on sync. If an enum case get added in Airtable, we should be able to sync it without going through the management UI

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -38,34 +39,44 @@ const fieldTypeOptions: { type: Field["type"]; label: string }[] = [
...
+                    <input type="checkbox" checked={!isIgnored} tabIndex={-1} readOnly disabled={disabled} />
                     <span>{originalFieldName ?? field.id}</span>
                 </button>
                 <ChevronIcon />
                 <input
                     type="text"
                     style={{
                         width: "100%",
-                        opacity: disabled ? 0.5 : 1,
+                        opacity: isIgnored || disabled ? 0.5 : 1,
```
triozer, line 77: Should we make `isIgnored || disabled` a variable?

triozer (commented):
Nice one!

plugins/airtable/src/FieldMapping.tsx
alecmev: Thought about it, but the best name I could come up with is `isIgnoredOrDisabled`, so, might as well skip the indirection and keep as-is 😄

plugins/airtable/src/FieldMapping.tsx
alecmev: We also did this in CMS Starter, just in case. The thinking is that otherwise people without Design won't be able to sync at all. Also, if you trace the calls, `syncExistingCollection` just passes `await collection.getFields()` as `fieldsToSync`, so it never even looks at potential changes in enums. Which is worth fixing! But out of scope here.

PR #240, opened 2025-06-11
Check permissions in Hubspot plugin

alecmev:
### Description

Fixes: https://github.com/framer/company/issues/32803

### Testing

- [ ] Same QA steps as Airtable: https://github.com/framer/plugins/pull/237


plugins/hubspot/src/blog.ts
```diff
@@ -81,29 +87,77 @@ export async function getBlogPluginContext(): Promise<BlogPluginContext> {
...
 
 function processPost({ post, fieldsById, unsyncedItemIds, status }: ProcessPostParams) {
     let slugValue: string | null = null
-    const fieldData: Record<string, unknown> = {}
+    const fieldData: FieldDataInput = {}
 
     unsyncedItemIds.delete(post.id)
 
-    for (const [propertyName, propertyValue] of Object.entries(post)) {
+    for (const [propertyName, propertyValue] of Object.entries(post) as [string, unknown][]) {
```
triozer, line 160: is this required?

alecmev: Yeah, it's `[string, any][]` otherwise, which suppresses a bunch of type-checking below.

PR #243, opened 2025-06-13
Fix file type handling

triozer:
### Description

This PR improves file handling in the Notion plugin by checking the field type before processing file properties. Now, when a Notion file property is mapped, the system correctly handles it based on whether it's configured as a "file" or "image" field type in the collection.

### Testing

- [x] Test file handling functionality
  - [x] Create a Notion database with file attachments
  - [x] Map the file field as an image type
  - [x] Verify files appear correctly as images in the collection
  - [x] Map the file field as a file type
  - [x] Verify files appear correctly as files in the collection

### DIFF
--- a/plugins/notion/src/notion.ts
+++ b/plugins/notion/src/notion.ts
@@ -21,7 +21,7 @@ import {
     type ManagedCollectionField,
     type ManagedCollectionFieldInput,
     type ManagedCollectionItemInput,
-    framer
+    framer,
 } from "framer-plugin"
 import pLimit from "p-limit"
 import { blocksToHtml, richTextToHTML } from "./blocksToHTML"
@@ -396,6 +396,7 @@ export function richTextToPlainText(richText: RichTextItemResponse[]) {
 
 export function getFieldDataEntryInput(
     property: PageObjectResponse["properties"][string],
+    fieldType: ManagedCollectionField["type"],
     { supportsHtml }: { supportsHtml: boolean }
 ): FieldDataEntryInput | undefined {
     switch (property.type) {
@@ -464,7 +465,8 @@ export function getFieldDataEntryInput(
             }
         }
         case "unique_id": {
-            if (Number.isNaN(property.unique_id.number) || typeof property.unique_id.number !== "number") return undefined
+            if (Number.isNaN(property.unique_id.number) || typeof property.unique_id.number !== "number")
+                return undefined
             return {
                 type: "number",
                 value: property.unique_id.number,
@@ -484,10 +486,11 @@ export function getFieldDataEntryInput(
         }
         case "files": {
             const firstFile = property.files[0]
-            if (!firstFile) return {
-                type: "file",
-                value: null,
-            }
+            if (!firstFile)
+                return {
+                    type: "file",
+                    value: null,
+                }
 
             if (firstFile.type === "external" && firstFile.external.url) {
                 return {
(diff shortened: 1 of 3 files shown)


triozer:
* **#243** <a href="https://app.graphite.dev/github/pr/framer/plugins/243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/framer/plugins/243?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

plugins/notion/src/notion.ts
```diff
@@ -464,7 +465,8 @@ export function getFieldDataEntryInput(
             }
         }
         case "unique_id": {
-            if (Number.isNaN(property.unique_id.number) || typeof property.unique_id.number !== "number") return undefined
+            if (Number.isNaN(property.unique_id.number) || typeof property.unique_id.number !== "number")
+                return undefined
```
alecmev: Nit: `{}`

plugins/notion/src/notion.ts
```diff
@@ -484,10 +486,11 @@ export function getFieldDataEntryInput(
...
             const firstFile = property.files[0]
-            if (!firstFile) return {
-                type: "file",
-                value: null,
-            }
+            if (!firstFile)
+                return {
+                    type: "file",
+                    value: null,
+                }
```
alecmev: Nit:

```suggestion
            if (!firstFile) return { type: "file", value: null }
```

plugins/notion/src/notion.ts
```diff
@@ -396,6 +396,7 @@ export function richTextToPlainText(richText: RichTextItemResponse[]) {
 
 export function getFieldDataEntryInput(
     property: PageObjectResponse["properties"][string],
+    fieldType: ManagedCollectionField["type"],
```
alecmev: Nit: Put in the option object, perhaps, since it's already there?

plugins/notion/src/notion.ts
```diff
@@ -396,6 +396,7 @@ export function richTextToPlainText(richText: RichTextItemResponse[]) {
 
 export function getFieldDataEntryInput(
```
alecmev, line 397: Nit: Since we have `fieldType`, then maybe also check that `property.type` is compatible with it? So, at the beginning of every case check if `fieldType` is good, and return `undefined` if not? And then every `type: ...` can become `type: fieldType` (or rename `fieldType` to `type` and then it's just `{ type, ... }` 😛 ).

alecmev (approved):
LGTM, only nits! 👌

plugins/notion/src/notion.ts
triozer: Mhh, yeah but it's not the same. The `property` field type comes from Notion

plugins/notion/src/notion.ts
triozer: I don’t think it’s worth the hassle right now, since we’re currently rewriting the plugin and this logic will be updated as part of that process. Let’s see if we still need this once the new rewrite is done

hacksman:
Hi team, thanks for the fix – I just tested the update.

Some of the images are now syncing correctly, which is great progress.
However, there are still a significant number of images that failed to sync in the same collection setup. It seems the issue is not fully resolved yet.

I've re-checked the field mapping (image type is correctly set), and even re-synced the database, but some social preview images remain missing.

image

hacksman:
> Hi @hacksman, could you please create a new issue? This will help us track and explore it more efficiently. Also, can you please share any log showing in your browser console/toast? Thanks!

Hi @triozer, thanks for the quick follow-up!

I've created a new issue as requested: 👉 https://github.com/framer/plugins/issues/245

The issue includes all the details, reproduction steps, console logs, and screenshots. Let me know if there's anything else I can provide to help investigate further.

PR #246, opened 2025-06-18
Fix ESLint, check in CI

alecmev:
### Description

Fixes: https://github.com/framer/company/issues/33125

The issue describes the prior state of things.

Chose to go for an `eslint-config` package because then the ESLint-related dependencies can be contained to it, rather than leaking into the root `package.json`, and could potentially be published in the future, if we desire.

Plugins delegate the invocation of `eslint` to the root `package.json` because then we can adjust the flags and such in a single place, rather than having to update every plugin. It's also a tiny bit cleaner boundary-wise, since ESLint is a dependency only in the root.

Enabling the bare minimum rule-wise as to not get bogged down by fixing the new errors. Not enabling `--max-warnings 0` either, since there's a non-trivial amount of fixing needed. Will do in follow-ups, after Biome.

### Testing

- [x] CI passes.
- [x] Test Photobooth.
- [x] Test RSS Feeds.


plugins/airtable/package.json
```diff
@@ -6,10 +6,10 @@
     "scripts": {
         "dev": "vite",
         "build": "vite build",
-        "lint": "eslint . --max-warnings 0",
+        "check-eslint": "run g:check-eslint",
```
triozer, line 9: Is there a way to not run the tests for all plugins when I am working on a specific plugin?

plugins/ascii/package.json
```diff
@@ -7,11 +7,11 @@
     "scripts": {
         "dev": "vite",
         "build": "vite build --base=${PREFIX_BASE_PATH:+/$npm_package_name}/",
-        "format": "prettier . --write",
-        "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+        "format": "run g:format",
```
triozer: There is no `g:format` right?

plugins/rss-feeds/src/importData.ts
triozer, line 1: **praise:** nice one, let's call the file `data.ts`

package.json
```diff
@@ -1,20 +1,25 @@
...
         "plugins/*"
     ],
     "scripts": {
+        "check": "turbo run --continue check-eslint check-typescript",
         "clean": "rm -rf ./p",
         "copy": "bash ./scripts/copy-build.sh",
         "dev": "turbo run dev --concurrency=20",
         "build": "turbo run build --env-mode=loose && run copy",
-        "typecheck": "turbo run typecheck"
+        "g:check-eslint": "cd $INIT_CWD && DEBUG='eslint:eslint' eslint .",
```
triozer, line 17: Will `$INIT_CWD` fine for Windows too? Probably we should have something like `cross-env-shell`? Not sure tho

alecmev: One more win for Yarn I forgot to mention, actually! It runs these in [its own shell](https://yarnpkg.com/features/scripting#portable-shell), so it'll "just work" 😌

plugins/rss-feeds/src/importData.ts
alecmev: Yup, good call! And the praise goes to `eslint-plugin-react-refresh` 😄

plugins/ascii/package.json
alecmev: Indeed there isn't 🥲 Cursor being overly "helpful". Thanks!

plugins/airtable/package.json
alecmev: This checks only this plugin 😉 `g:check-eslint` does this:

```sh
cd $INIT_CWD && DEBUG='eslint:eslint' eslint .
```

Where `$INIT_CWD` is `plugins/airtable`. Same for TypeScript:

```sh
tsc --project $INIT_CWD
```

PR #247, opened 2025-06-18
Fix files property types in Notion plugin

madebyisaacr:
### Description

This pull request fixes the types of files properties in Notion. The files property type can be imported as a File or Image field, but it was hardcoded to use "link" or "file" for the type. This causes "external" type images to fail to import because of the type mismatch.

For the last week, any image added in Notion with "Embed link" fails to import because of this type mismatch error.
image

There's a TypeScript error on `type: fieldType`, but I can't figure out why because the types look correct.

### Testing

You can use this database to test:
https://www.notion.so/framer/Notion-Plugin-Test-Database-1efadf6e8c9680b38910d2d0575513a0?source=copy_link

The first item (Red) has an external image. On the production version of the Notion plugin, it doesn't import that image, and in this PR version it does. 

CMS plugins such as CMS Export also fail to load on the production version due to the type mismatch errors.

### DIFF
--- a/plugins/notion/src/notion.ts
+++ b/plugins/notion/src/notion.ts
@@ -487,22 +487,18 @@ export function getFieldDataEntryInput(
         }
         case "files": {
             const firstFile = property.files[0]
-            if (!firstFile) return { type: "file", value: null }
 
-            if (firstFile.type === "external" && firstFile.external.url) {
-                return {
-                    type: "link",
-                    value: firstFile.external.url,
+            if (firstFile) {
+                if (firstFile.type === "external") {
+                    return { type: fieldType, value: firstFile.external?.url ?? "" }
+                }
+
+                if (firstFile.type === "file") {
+                    return { type: fieldType, value: firstFile.file?.url ?? "" }
                 }
             }
 
-            const isFileOrImage = fieldType === "file" || fieldType === "image"
-            if (firstFile.type === "file" && isFileOrImage) {
-                return {
-                    type: fieldType,
-                    value: firstFile.file.url,
-                }
-            }
+            return { type: fieldType, value: "" }
         }
     }
 }


hacksman:
Thanks for the detailed explanation. I just verified this on our end using our Notion-based CMS. Here's what I found:

1. Images uploaded directly to Notion (via file upload) are stored under Notion's secure CDN (e.g. notion-static.com) and import successfully.
2. Images embedded via "Embed link" (e.g. using https://s3.amazonaws.com/assets.xmind.net..., or external CDNs) consistently fail to import on the production version of the plugin.
image

Thanks again for fixing this in the PR!

plugins/notion/src/notion.ts
```diff
@@ -487,22 +487,18 @@ export function getFieldDataEntryInput(
...
             const firstFile = property.files[0]
-            if (!firstFile) return { type: "file", value: null }
 
-            if (firstFile.type === "external" && firstFile.external.url) {
-                return {
-                    type: "link",
-                    value: firstFile.external.url,
+            if (firstFile) {
+                if (firstFile.type === "external") {
+                    return { type: fieldType, value: firstFile.external?.url ?? "" }
```
RayJason: I ran your code locally. There is an error locally. You can specify the type here, which is either file or image.

image


I fixed the issue mentioned in this way:
```typescript
case "files": {
    const firstFile = property.files[0]
    const type = fieldType === 'image' ? 'image' : 'file'
    if (!firstFile) return { type, value: null }
    
    if (firstFile.type === "external") {
        return { type, value: firstFile.external?.url ?? null }
    }
    if (firstFile.type === "file") {
        return { type, value: firstFile.file?.url ?? null }
    }
}
```

madebyisaacr: Fixed! Thanks @RayJason

plugins/notion/src/notion.ts
```diff
@@ -487,22 +487,18 @@ export function getFieldDataEntryInput(
...
-            if (firstFile.type === "file" && isFileOrImage) {
-                return {
-                    type: fieldType,
-                    value: firstFile.file.url,
+                if (firstFile.type === "file") {
+                    return { type, value: firstFile.file?.url ?? null }
                 }
             }
+
+            return { type, value: null }
```
triozer: Can we use a switch instead?

```ts
            switch (firstFile?.type) {
                case "external":
                    return { type, value: firstFile.external.url }
                case "file":
                    return { type, value: firstFile.file.url }
                default:
                    return { type, value: null }
            }
```

triozer (approved):
Nice fix! Thanks @madebyisaacr, left a small nit

plugins/notion/src/notion.ts
triozer: Applied the fix, will release a new version.

AndiCrafthunt:
Still no upload possible for me..

PR #249, opened 2025-06-19
Improve Airtable plugin design

madebyisaacr:
### Description

This pull request improves the design and UX of the Airtable plugin.

- Increase default size of plugin window to prevent text getting cut off.
- Improve appearance and sorting of unsupported fields.
- Make field types with one option non-clickable.
- Add heading that links to the Airtable base.
- Use primary button style.

Before and after:
image

### DIFF
--- a/plugins/airtable/src/App.tsx
+++ b/plugins/airtable/src/App.tsx
@@ -24,11 +24,11 @@ export function App({ collection, previousBaseId, previousTableId, previousSlugF
     const [noTableAccess, setNoTableAccess] = useState(false)
 
     useLayoutEffect(() => {
-        const hasDataSourceSelected = Boolean(dataSource)
+        const hasDataSourceSelected = Boolean(dataSource) || isLoadingDataSource
 
         framer.showUI({
-            width: hasDataSourceSelected ? 360 : 320,
-            height: hasDataSourceSelected ? 425 : 350,
+            width: hasDataSourceSelected ? 600 : 320,
+            height: hasDataSourceSelected ? 500 : 345,
             minWidth: hasDataSourceSelected ? 360 : undefined,
             minHeight: hasDataSourceSelected ? 425 : undefined,
             resizable: hasDataSourceSelected,
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -6,6 +6,7 @@ import { framer, useIsAllowedTo } from "framer-plugin"
 import { useState, useEffect, memo } from "react"
 import { mergeFieldsWithExistingFields, syncCollection, syncMethods } from "./data"
 import { ALLOWED_FILE_TYPES, isCollectionReference } from "./utils"
+import cx from "classnames"
 
 function ChevronIcon() {
     return (
@@ -41,9 +42,16 @@ interface FieldMappingRowProps {
     originalFieldName: string | undefined
     isIgnored: boolean
     disabled: boolean
+    unsupported?: boolean
+    missingCollection?: boolean
     onToggleIgnored?: (fieldId: string) => void
     onNameChange?: (fieldId: string, name: string) => void
     onTypeChange?: (fieldId: string, type: string) => void
+}
+
+interface SelectOption {
+    id: string
+    label: string
 }
 
 const FieldMappingRow = memo(
@@ -52,15 +60,35 @@ const FieldMappingRow = memo(
         originalFieldName,
         isIgnored,
         disabled,
+        unsupported,
+        missingCollection,
         onToggleIgnored,
         onNameChange,
         onTypeChange,
     }: FieldMappingRowProps) => {
+        let selectOptions: SelectOption[] = []
+        if (!unsupported && !missingCollection) {
+            if (isCollectionReference(field)) {
+                selectOptions = field.supportedCollections.map(collection => ({
+                    id: collection.id,
+                    label: collection.name,
+                }))
+            } else if (Array.isArray(field.allowedTypes)) {
+                selectOptions = field.allowedTypes.map(type => {
+                    const capitalizedType = type.charAt(0).toUpperCase() + type.slice(1)
+                    return {
+                        id: type,
+                        label: fieldTypeOptions.find(option => option.type === type)?.label ?? capitalizedType,
+                    }
+                })
+            }
+        }
+
         return (
             <>
                 <button
                     type="button"
-                    className="source-field"
+                    className={cx("source-field", unsupported || missingCollection ? "unsupported" : "")}
                     aria-disabled={isIgnored}
                     onClick={disabled ? undefined : () => onToggleIgnored?.(field.id)}
                     tabIndex={0}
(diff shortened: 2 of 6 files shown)


plugins/airtable/src/FieldMapping.tsx
```diff
@@ -41,26 +42,53 @@ interface FieldMappingRowProps {
...
+                })
+            }
+        }
+
         return (
             <>
                 <button
                     type="button"
-                    className="source-field"
+                    className={cx("source-field", unsupported || missingCollection ? "unsupported" : "")}
```
melissarofman: Using a css class to set pointer-events to none seems the wrong approach here. I'd set the button to disabled and style `sourcefield:disabled` instead (if needed, you're not really changing the styles now, I think). And please remove classnames as a dependency, since it will no longer be used.

As a nit for future work, `&&` is nicer than a ternary when there's no class set if the condition is false
```suggestion
                    className={cx("source-field", (unsupported || missingCollection) && "unsupported")}
```

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -41,26 +42,53 @@ interface FieldMappingRowProps {
...
+}
+
 const FieldMappingRow = memo(
     ({
         field,
         originalFieldName,
         isIgnored,
         disabled,
+        unsupported,
+        missingCollection,
```
melissarofman, line 62: `missingCollection` is not used anywhere? Can we remove it or make sure it's correctly used? We don't need to support unused code paths.

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -70,55 +98,46 @@ const FieldMappingRow = memo(
...
+                            >
+                                {selectOptions.map(option => (
+                                    <option key={option.id} value={option.id}>
+                                        {option.label}
+                                    </option>
+                                ))}
+                            </select>
+                        )}
+                    </>
+                )}
```
melissarofman: Let's avoid nested ternaries.Please refactor this into subcomponents that early return the correct UI.

melissarofman (changes requested):
Left some feedback on the code itself, but also have some questions about the redesign:
1. I'm not conviced about removing one of the carets in the field rows
2. I don't think it's necessary to say "Airtable Column"

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -24,14 +25,14 @@ function ChevronIcon() {
...
-    { type: "link", label: "Link" },
     { type: "date", label: "Date" },
+    { type: "link", label: "Link" },
+    { type: "image", label: "Image" },
+    { type: "color", label: "Color" },
+    { type: "boolean", label: "Toggle" },
+    { type: "number", label: "Number" },
     { type: "enum", label: "Option" },
     { type: "file", label: "File" },
 ]
```
madebyisaacr, line 36: This change makes the field type names and order match the field editing menu in Framer.

Boolean -> Toggle
String -> Plain Text

image

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: `missingCollection` is used for Airtable reference fields that don't have a corresponding collection in Framer.

image


Line 310-312 in FieldMapping.tsx
```tsx
const missingCollectionFields = fields.filter(
    field => isCollectionReference(field) && field.supportedCollections.length === 0
)
```

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: Ok, removed.

madebyisaacr:
> I'm not conviced about removing one of the carets in the field rows

The type and name columns are combined without a caret because the left side is Airtable and right side is Framer. The Notion and Sheets plugins don't have a second caret between type and name either.

> I don't think it's necessary to say "Airtable Column"

Good point, I changed it back to just "Column"

The before and after screenshot is up to date now.

melissarofman (approved):

madebyisaacr:
Done. Fixed the linter type error, rebased, and ran the biome command:

`yarn biome format plugins/airtable --write`
`Formatted 15 files in 17ms. No fixes applied.`

plugins/airtable/src/App.tsx
```diff
@@ -23,11 +23,11 @@ export function App({ collection, previousBaseId, previousTableId, previousSlugF
     const [noTableAccess, setNoTableAccess] = useState(false)
 
     useLayoutEffect(() => {
-        const hasDataSourceSelected = Boolean(dataSource)
+        const hasDataSourceSelected = Boolean(dataSource) || isLoadingDataSource
```
triozer: Why do we need `isLoadingDataSource`? In a case where fetching the datasource fails, we would see a window **re**size.

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -63,61 +90,46 @@ const FieldMappingRow = memo(
...
-                >
-                    {isCollectionReference(field) && (
-                        <>
-                            {field.supportedCollections.length === 0 && (
-                                <option value="unsupported">Unsupported</option>
-                            )}
-                            {field.supportedCollections.map(collection => (
-                                <option key={collection.id} value={collection.id}>
-                                    {collection.name}
+                {unsupported || missingCollection ? (
```
triozer, line 99: Should this be wrapped in parentheses?

```suggestion
                {(unsupported || missingCollection) ? (
```

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -325,14 +352,10 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
-                <span className="column-span-2">Field</span>
                 <span>Type</span>
+                <span>Name</span>
                 {fields
-                    .filter(
-                        field =>
-                            field.type !== "unsupported" &&
-                            (!isCollectionReference(field) || field.supportedCollections.length > 0)
-                    )
+                    .filter(field => !unsupportedFields.includes(field) && !missingCollectionFields.includes(field))
```
triozer: Is it possible to check by field.id here instead of relying on object reference? Would that be more reliable in this context?

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: Prettier auto removes the parentheses when you format the file, and it works as expected. So I guess not unless we need to change a global prettier config.

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: `unsupportedFields` and `missingCollectionFields` are filtered arrays of `fields`, so it wouldn't be more or less reliable than the current method.

```tsx
const unsupportedFields: PossibleField[] = fields.filter(field => field.type === "unsupported")
const missingCollectionFields: PossibleField[] = fields.filter(
    field => isCollectionReference(field) && field.supportedCollections.length === 0
)
```

plugins/airtable/src/FieldMapping.tsx
triozer: OK! Also, we don't use Prettier anymore, so it seems to be a Biome thing 👌🏽

plugins/airtable/src/FieldMapping.tsx
triozer: It's not entirely accurate to say there's no difference. Using object reference equality (i.e., `unsupportedFields.includes(field)`) will only work if the `field` objects in all arrays are the exact same instance (reference) in memory. If, for any reason, new objects are created or fields are fetched/constructed separately, then checking by reference can fail—even if the objects have the same `id` and properties.

Comparing by `field.id` (or another unique identifier) is generally more robust and reliable, especially if there's any chance that the arrays might contain different object instances representing the same logical field.

plugins/airtable/src/App.tsx
madebyisaacr: The loading screen is visible when `isLoadingDataSource` is true:

image

This code change prevents it from jumping window sizes when it goes from loading to field mapping states.

triozer: Feels like we should keep the window fixed and the same size as the SelectDataSource screen until we have something to display, otherwise the resizing just adds noise.

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: Ok, I changed it.

plugins/airtable/src/App.tsx
madebyisaacr: Yes, preventing resizing is what this change does. Previously it would resize when it finished loading.

PR #250, opened 2025-06-19
Airtable: add support for color fields

madebyisaacr:
### Description

This pull request adds support for importing single line text fields from Airtable as colors in Framer. Airtable does not have a color field type built-in, so this update lets you enter colors as text.

The default color if a color field has no value is #000000.

### Testing

- [x] Add a single line text column to your Airtable base.
- [x] Fill in the column with color values in different formats such as hex, hsl, and rgba.
- [x] Select the "Color" type in the Airtable plugin when importing.

### DIFF
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -219,6 +219,16 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                         return { ...field, type: "string" } as PossibleField
                     case "formattedText":
                         return { ...field, type: "formattedText" } as PossibleField
+                    case "number":
+                        return { ...field, type: "number" } as PossibleField
+                    case "boolean":
+                        return { ...field, type: "boolean" } as PossibleField
+                    case "color":
+                        return { ...field, type: "color" } as PossibleField
+                    case "date":
+                        return { ...field, type: "date" } as PossibleField
+                    case "enum":
+                        return { ...field, type: "enum" } as PossibleField
                     default:
                         return field
                 }
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -219,16 +219,8 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                         return { ...field, type: "string" } as PossibleField
                     case "formattedText":
                         return { ...field, type: "formattedText" } as PossibleField
-                    case "number":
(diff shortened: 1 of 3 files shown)


plugins/airtable/src/data.ts
```diff
@@ -248,6 +247,12 @@ export async function getItems(dataSource: DataSource, slugFieldId: string) {
                             type: field.type,
                         }
                         break
+                    case "color":
+                        fieldData[field.id] = {
+                            value: "#000000",
+                            type: "color",
+                        }
+                        break
```
triozer: Color can be set to null, don't need to specify a default color

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -219,6 +219,16 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
+                    case "number":
+                        return { ...field, type: "number" } as PossibleField
+                    case "boolean":
+                        return { ...field, type: "boolean" } as PossibleField
+                    case "color":
+                        return { ...field, type: "color" } as PossibleField
+                    case "date":
+                        return { ...field, type: "date" } as PossibleField
+                    case "enum":
+                        return { ...field, type: "enum" } as PossibleField
```
triozer: We only need to add the `color` entry here right?

triozer: _unrelated:_ but we probably need to find a fix for those `as PossibleField`

plugins/airtable/src/data.ts
madebyisaacr: When color is null, it uses #0099FF (blue) as the default color.

What do you think about changing it to use transparent `rgba(0, 0, 0, 0)` as the default color when there's no color value in Airtable? Solid black or transparent makes more sense to use as the default color than blue.

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: For this PR, yes.

https://github.com/framer/plugins/pull/248 includes this same change but it's needed to fix a bug there.

plugins/airtable/src/data.ts
triozer: I think that's fine, it's the default color used by Framer when adding a Color field

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: Removed the extra types.

triozer: Cool, let's do minimal change in this PR

PR #251, opened 2025-06-19
[Main task] Add and use a shared Vite config

alecmev:
### Description

Fixes: https://github.com/framer/company/issues/33180

In a similar vein to #246, it's just nice to keep plugin-specific directories and their `package.json` focused on the actual functionality of the plugin rather than the intricacies of building, linting, type-checking, formatting, etc., which can be shared and reused easily.

Indeed, `--base=${PREFIX_BASE_PATH:+/$npm_package_name}/` served no purpose, see [Slack](https://framer-team.slack.com/archives/C06L5H5ADK2/p1727419972613179?thread_ts=1727373105.569579&cid=C06L5H5ADK2).

`strictImportMetaEnv` makes `any` go away from the types, forcing you to augment the interface with the specific variables you're expecting. Helps with typos.

The CI check parallelization takes the total time from ~1:45 to ~1:00. Could be improved a lot with a beefier runner.

### Testing

- [x] Try out `yarn dev`, `yarn build`, `yarn preview` in the root.


package.json
```diff
@@ -12,14 +12,23 @@
...
+        "dev": "turbo run dev --concurrency=40",
         "build": "turbo run build --env-mode=loose && run copy",
+        "g:build": "cd $INIT_CWD && run g:vite build",
         "g:check-eslint": "cd $INIT_CWD && DEBUG='eslint:eslint' eslint .",
-        "g:check-typescript": "tsc --project $INIT_CWD"
+        "g:check-typescript": "tsc --project $INIT_CWD",
+        "g:dev": "cd $INIT_CWD && run g:vite",
+        "g:preview": "cd $INIT_CWD && run g:vite preview",
+        "g:vite": "cd $INIT_CWD && NODE_OPTIONS='--no-warnings=ExperimentalWarning' vite --config $PROJECT_CWD/packages/vite-config/src/index.ts",
+        "preview": "turbo run preview --concurrency=40"
```
triozer, line 23: Do we need the preview script? I don't think we use it?

triozer (approved):
LGTM!

package.json
alecmev: Thought about dropping it altogether, but the cost of having it felt low enough to just leave it alone, and who knows, it might come in handy for troubleshooting discrepancies between development and production. I'll leave it in for now, let me know if you think it's worth ripping out, will do in a follow-up! 😌

PR #252, opened 2025-06-20
fix: notion plugin selected error

RayJason:
### Description

image

This pull request is for fix the React dom error
```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
```

### Testing

- [x] Description of test case one
  - [x] Open Notion Plugin
  - [x] Review console panel

### DIFF
--- a/plugins/notion/src/SelectDatabase.tsx
+++ b/plugins/notion/src/SelectDatabase.tsx
@@ -77,19 +77,19 @@ export function SelectDatabase({ onDatabaseSelected }: SelectDatabaseProps) {
             <div className="flex flex-col gap-[10px] mt-auto pb-[15px]">
                 <div className="flex gap-[10px]">
                     <select
-                        value={selectedDatabase ?? (isLoadingOrFetching ? "loading" : "placeholder")}
+                        value={selectedDatabase ?? ""}
                         onChange={event => setSelectedDatabase(event.target.value)}
                         className="flex-1 shrink-1"
                         disabled={!selectEnabled}
                     >
                         {isLoadingOrFetching && (
-                            <option disabled value="loading">
+                            <option disabled value="">
                                 Loading…
                             </option>
                         )}
                         {!isLoadingOrFetching && (
                             <>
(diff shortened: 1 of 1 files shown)


plugins/notion/src/SelectDatabase.tsx
```diff
@@ -77,19 +77,19 @@ export function SelectDatabase({ onDatabaseSelected }: SelectDatabaseProps) {
...
                     <select
-                        value={selectedDatabase ?? ""}
+                        value={selectedDatabase ?? (isLoadingOrFetching ? "loading" : "placeholder")}
                         onChange={event => setSelectedDatabase(event.target.value)}
                         className="flex-1 shrink-1"
                         disabled={!selectEnabled}
                     >
                         {isLoadingOrFetching && (
-                            <option disabled selected>
+                            <option disabled value="loading">
```
triozer: If we do `value=""`, then we don't have to add per state value.
```suggestion
                            <option disabled value="">
```

plugins/notion/src/SelectDatabase.tsx
```diff
@@ -77,19 +77,19 @@ export function SelectDatabase({ onDatabaseSelected }: SelectDatabaseProps) {
...
                                 Loading…
                             </option>
                         )}
                         {!isLoadingOrFetching && (
                             <>
-                                <option disabled selected>
+                                <option disabled value="placeholder">
```
triozer: If we do `value=""`, then we don't have to add per state value.
```suggestion
                                <option disabled value="">
```

triozer (commented):
Nice fix, thanks! Left some comment

plugins/notion/src/SelectDatabase.tsx
RayJason: Adopted.

plugins/notion/src/SelectDatabase.tsx
RayJason: Adopted.

triozer (approved):

PR #225, opened 2025-05-19
Greenhouse Plugin

clementroche:
This pull request add Greenhouse Plugin with the following features:
- sync Greenhouse Job Board
- add a Job Application form to Framer canvas

Test URL: https://framer-greenhouse.darkroom.engineering/

PR #253, opened 2025-06-20
Add Greenhouse Plugin

triozer:
PR based on https://github.com/framer/plugins/pull/225.

This pull request introduces the Greenhouse plugin, which enables syncing Greenhouse job data into the Framer CMS. The changes include the addition of plugin configuration files, core functionality implementation, and supporting assets.

### Plugin Configuration and Setup:

* Added a `.gitignore` file to exclude common build artifacts, debug files, and local environment files. (`plugins/greenhouse/.gitignore`)
* Created `framer.json` to define plugin metadata, including supported modes and the plugin icon. (`plugins/greenhouse/framer.json`)
* Added `package.json` with dependencies, scripts for development and build processes, and plugin tools integration. (`plugins/greenhouse/package.json`)
* Added `index.html` as the entry point for the plugin's UI, with basic HTML structure and script inclusion. (`plugins/greenhouse/index.html`)
* Added a README file providing an overview of the Greenhouse plugin and its purpose. (`plugins/greenhouse/README.md`)

### Core Functionality Implementation:

* Implemented the main application logic in `App.tsx`, including UI state management, data source handling, and plugin lifecycle integration. (`plugins/greenhouse/src/App.tsx`)
* Developed the `FieldMapping` component to manage field mapping between Greenhouse data and the Framer CMS, with support for slug field selection and synchronization. (`plugins/greenhouse/src/components/FieldMapping.tsx`)

### Styling and Linting:

* Added `App.css` to define styles for the plugin UI, including layouts, forms, and interactive elements. (`plugins/greenhouse/src/App.css`)
* Configured ESLint with support for TypeScript, React hooks, and React Refresh for development consistency. (`plugins/greenhouse/eslint.config.js`)


plugins/greenhouse/src/App.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    useLayoutEffect(() => {
+        const hasDataSourceSelected = Boolean(dataSource)
+
+        framer.showUI({
+            width: hasDataSourceSelected ? 400 : 320,
+            height: hasDataSourceSelected ? 427 : 325,
+            minHeight: hasDataSourceSelected ? 427 : undefined,
+            minWidth: hasDataSourceSelected ? 400 : undefined,
+            resizable: hasDataSourceSelected,
```
kaloyanvi, line 31: instead of having a ternary on each object field, it'd be much nicer to just have two different objects

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+                aria-disabled={disabled || isMissingReference}
+                onClick={() => onToggleDisabled(field.id)}
+                tabIndex={0}
+            >
+                <input type="checkbox" checked={!disabled && !isMissingReference} tabIndex={-1} readOnly />
+                <span>{originalFieldName ?? field.id}</span>
+            </button>
+            <ChevronIcon />
+            {isCollectionReference(field) && (
+                <>
```
kaloyanvi: Fragment is not needed here

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+                className="source-field"
+                aria-disabled={disabled || isMissingReference}
+                onClick={() => onToggleDisabled(field.id)}
+                tabIndex={0}
+            >
+                <input type="checkbox" checked={!disabled && !isMissingReference} tabIndex={-1} readOnly />
+                <span>{originalFieldName ?? field.id}</span>
+            </button>
+            <ChevronIcon />
+            {isCollectionReference(field) && (
```
kaloyanvi: In the current state of `isCollectionReference`, this code might as well be written in JS instead.

plugins/greenhouse/src/utils.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function decodeHtml(html: string) {
+    const textarea = document.createElement("textarea")
+    textarea.innerHTML = html
+    return textarea.value
+}
+
+export function isCollectionReference(field: GreenhouseField) {
+    return field.type === "collectionReference" || field.type === "multiCollectionReference"
+}
```
kaloyanvi: this should be something like this:

```ts
type ReferenceField = CollectionReferenceField | MultiCollectionReferenceField
export function isCollectionReference(field: GreenhouseField): field is ReferenceField {
    return field.type === "collectionReference" || field.type === "multiCollectionReference"
}
```

plugins/greenhouse/src/utils.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function isMissingReferenceField(field: GreenhouseField) {
```
kaloyanvi: this should also be narrowing the type in the return

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+                </>
+            )}
+            {!isCollectionReference(field) && (
+                <input
+                    type="text"
+                    className="target-field"
+                    disabled={disabled}
+                    placeholder={originalFieldName}
+                    value={field.name !== originalFieldName ? field.name : ""}
+                    onChange={event => onNameChange(field.id, event.target.value || (originalFieldName ?? ""))}
```
kaloyanvi:
```suggestion
                    onChange={event => onNameChange(field.id, event.target.value ?? (originalFieldName ?? ""))}
```

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+    const [possibleSlugFields] = useState(() =>
+        dataSource.fields.filter(field => field.type === "string" && field.canBeUsedAsSlug)
+    )
+
+    const [selectedSlugField, setSelectedSlugField] = useState<GreenhouseField | null>(
+        possibleSlugFields.find(field => field.id === initialSlugFieldId) ?? possibleSlugFields[0] ?? null
+    )
+
+    const [fields, setFields] = useState(initialManagedCollectionFields)
```
kaloyanvi: nit: this is difficult to read/follow. Can we make the const something like this:

```tsx
const emptyArray: GreenhouseField[] = []
// ...
const [fields, setFields] = useState<GreenhouseField[]>(emptyArray)
```

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+    const [ignoredFieldIds, setIgnoredFieldIds] = useState(() => {
+        const initialFieldIds = new Set()
+
+        for (const field of dataSource.fields) {
+            if (isMissingReferenceField(field)) {
+                initialFieldIds.add(field.id)
+            }
```
kaloyanvi:
```suggestion
            if (!isMissingReferenceField(field)) continue
            initialFieldIds.add(field.id)
```

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+                if (abortController.signal.aborted) return
+
+                setFields(mergeFieldsWithExistingFields(dataSource.fields, collectionFields))
+
+                const existingFieldIds = new Set(collectionFields.map(field => field.id))
+                const ignoredFields = dataSource.fields.filter(sourceField => !existingFieldIds.has(sourceField.id))
+
+                if (initialSlugFieldId) {
+                    setIgnoredFieldIds(new Set(ignoredFields.map(field => field.id)))
+                }
```
kaloyanvi: We dont need two loops for this. Should be possible to build in a single pass.

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+    const changeCollectionId = (fieldId: string, collectionId: string) => {
+        setFields(prevFields => {
+            const updatedFields = prevFields.map(field => {
+                if (field.id !== fieldId) return field
+                if (!isCollectionReference(field)) return field
+                return { ...field, collectionId }
+            })
+            return updatedFields
+        })
+    }
```
kaloyanvi: Big part of the logic here is identical. Can we make something reusable that can be used by both `changeCollectionId` and `changeFieldName`?

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+                .filter(field => !ignoredFieldIds.has(field.id))
+                .map(field => ({
+                    ...field,
+                    name: field.name.trim() || field.id,
+                }))
+                .filter(
+                    field =>
+                        (field.type !== "collectionReference" && field.type !== "multiCollectionReference") ||
+                        field.collectionId !== ""
+                )
```
kaloyanvi: What is going on here?

kaloyanvi: Very inefficient, can easily be reduced to 1 loop instead of 3.

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+        } finally {
+            setStatus("mapping-fields")
+        }
+    }
+
+    if (isLoadingFields) {
+        return (
+            <main className="loading">
+                <div className="framer-spinner" />
+            </main>
```
kaloyanvi: nit: This exact bit is already used the same way in `App.tsx`, can we make something shareable?

plugins/greenhouse/src/components/SelectDataSource.tsx
```diff
@@ -0,0 +1,93 @@
...
+                            Choose Source…
+                        </option>
+                        {dataSources.map(({ id, name }) => (
+                            <option key={id} value={id}>
+                                {name}
+                            </option>
+                        ))}
+                    </select>
+                </label>
+                <button disabled={!boardToken || !selectedDataSourceId || isLoading || !isAllowedToManage}>
```
kaloyanvi: can we extract this into a const

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+    fieldToCollectionsMap: Map<string, ManagedCollection[]>
+): GreenhouseDataSource {
+    const fields = dataSource.fields.map(field => {
+        if (!isCollectionReference(field)) return field
+
+        const matchingCollections = fieldToCollectionsMap.get(field.id)
+
+        return {
+            ...field,
+            collectionId: matchingCollections?.[0]?.id ?? "",
```
kaloyanvi, line 28: How does that work? Why do we need the `id` of the first one?

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+        if (collectionBoardToken !== boardToken) {
+            continue
+        }
+
+        boardCollections.push(collection)
+    }
+
+    if (boardCollections.length > 0) {
+        for (const field of dataSource.fields) {
+            if (field.type !== "multiCollectionReference" && field.type !== "collectionReference") {
```
kaloyanvi: shouldn't this use `isCollectionReference()` util?

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+
+async function getItems(
+    dataSource: GreenhouseDataSource,
+    fieldsToSync: readonly ManagedCollectionFieldInput[],
+    { boardToken, slugFieldId }: { boardToken: string; slugFieldId: string }
+): Promise<ManagedCollectionItemInput[]> {
+    const items: ManagedCollectionItemInput[] = []
+
+    const response = await fetch(getApiEndpoint(boardToken, dataSource))
+    const data = await response.json()
```
kaloyanvi: This data needs some validation and types.

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+                    fieldData[field.id] = { value: ids, type: field.type }
+                    break
+                }
+                case "collectionReference": {
+                    if (typeof value !== "object" || value == null || !("id" in value)) {
+                        continue
+                    }
+
+                    fieldData[field.id] = { value: String(value.id), type: field.type }
+                    break
```
kaloyanvi: Repeating `type: field.type` 10 times or something, can we define before or implement in other way so it doesnt repeat.

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+    boardToken: string,
+    collection: ManagedCollection,
+    dataSource: GreenhouseDataSource,
+    fields: readonly ManagedCollectionFieldInput[],
+    slugField: ManagedCollectionFieldInput
+) {
+    const existingItemsIds = await collection.getItemIds()
+    const items = await getItems(dataSource, fields, { boardToken, slugFieldId: slugField.id })
+    const unsyncedItems = new Set<string>(
+        existingItemsIds.filter(existingItemId => !items.find(item => item.id === existingItemId))
```
kaloyanvi: Each lookup here does a new loop.

kaloyanvi: You can use a set for better performance. Something like this:

```tsx
    const itemIds = new Set(items.map(item => item.id))
    const unsyncedItems = new Set<string>(
        existingItemsIds.filter(existingItemId => !itemIds.has(existingItemId))
    )
```

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+export async function syncCollection(
+    boardToken: string,
+    collection: ManagedCollection,
+    dataSource: GreenhouseDataSource,
+    fields: readonly ManagedCollectionFieldInput[],
+    slugField: ManagedCollectionFieldInput
+) {
+    const existingItemsIds = await collection.getItemIds()
+    const items = await getItems(dataSource, fields, { boardToken, slugFieldId: slugField.id })
+    const unsyncedItems = new Set<string>(
```
kaloyanvi: this is not items really though, its the ids right?

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+        existingItemsIds.filter(existingItemId => !items.find(item => item.id === existingItemId))
+    )
+
+    await collection.removeItems(Array.from(unsyncedItems))
```
kaloyanvi: Why are you making `unsyncedItems` a set if you never use it but directly convert it to an array?

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+    "ManagedCollection.addItems",
+    "ManagedCollection.setPluginData",
+] as const satisfies ProtectedMethod[]
+
+export async function syncExistingCollection(
+    collection: ManagedCollection,
+    previousDataSourceId: string | null,
+    previousSlugFieldId: string | null,
+    previousBoardToken: string | null
+): Promise<{ didSync: boolean }> {
```
kaloyanvi, line 256: At the consumer this is being destructured like this:

```
const { didSync } = await syncExistingCollection(
    activeCollection,
    previousDataSourceId,
    previousSlugFieldId,
    previousBoardToken
)
```

and in every condition in `syncExistingCollection` we return `{ didSync: <value> }`?

kaloyanvi: Am I missing something or this return type should just be a boolean?

plugins/greenhouse/src/dataSources.ts
```diff
@@ -0,0 +1,161 @@
...
+    id: string
+    name: string
+    /**
+     * The fields of the data source.
+     *
+     * The first field is the ID field.
+     * The rest of the fields are the fields of the data source.
+     */
+    fields: readonly GreenhouseField[]
+    apiEndpoint: string
```
kaloyanvi:
```suggestion
    apiURL: string
```

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,306 @@
...
+export function FieldMapping({ boardToken, collection, dataSource, initialSlugFieldId }: FieldMappingProps) {
+    const [status, setStatus] = useState<"mapping-fields" | "loading-fields" | "syncing-collection">(
+        initialSlugFieldId ? "loading-fields" : "mapping-fields"
+    )
+    const isSyncing = status === "syncing-collection"
+    const isLoadingFields = status === "loading-fields"
+
+    const [possibleSlugFields] = useState(() =>
+        dataSource.fields.filter(field => field.type === "string" && field.canBeUsedAsSlug)
+    )
```
kaloyanvi: This be recomputed on every render. Can we memoize?

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,275 @@
...
+}
+
+async function getItems(
+    dataSource: GreenhouseDataSource,
+    fieldsToSync: readonly ManagedCollectionFieldInput[],
+    { boardToken, slugFieldId }: { boardToken: string; slugFieldId: string }
+): Promise<ManagedCollectionItemInput[]> {
+    const items: ManagedCollectionItemInput[] = []
+
+    const response = await fetch(getApiEndpoint(boardToken, dataSource))
```
kaloyanvi: This needs error handling.

plugins/greenhouse/src/App.tsx
```diff
@@ -0,0 +1,84 @@
...
+    const [dataSource, setDataSource] = useState<GreenhouseDataSource | null>(null)
+    const [isLoading, setIsLoading] = useState(false)
+
+    useLayoutEffect(() => {
+        const hasDataSourceSelected = Boolean(dataSource)
+
+        framer.showUI({
+            width: hasDataSourceSelected ? 400 : 320,
+            height: hasDataSourceSelected ? 427 : 325,
+            minHeight: hasDataSourceSelected ? 427 : undefined,
```
kaloyanvi, line 29: How are the magic numbers derived? Should there be a comment or some logic about that?

plugins/greenhouse/src/dataSources.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Remove Greenhouse-specific keys from the fields. This is used to ensure that the fields are compatible with Framer API.
+ *
+ * @param fields - The fields to remove the keys from.
+ * @returns The fields with the keys removed.
+ */
+export function removeGreenhouseKeys(fields: GreenhouseField[]): ManagedCollectionFieldInput[] {
+    return fields.map(field => {
+        delete field.getValue
```
kaloyanvi, line 277: Doesn't this mutate the original fields, this should probably return new object.

kaloyanvi:
When using the plugin for the first time, permissions are not checked (correctly), which results to a broken state. 

https://github.com/user-attachments/assets/cbed02f8-6cb3-4ab9-af16-1f1d92a111ba

kaloyanvi:
I understand why this is done like this but in a small UI like this one, I think clicking the label should not focus the input because users expect to unfocus the input when clicking away but theres minimal space to click away which results in often clicking the label again.

https://github.com/user-attachments/assets/106e2017-dbbf-4271-937c-73036dbb7436

kaloyanvi:
Disabled state and not active state look exactly the same. Easy win would be to make the cursor not be `pointer` for the disabled state.

https://github.com/user-attachments/assets/243465d2-6364-4770-a1dc-335fd6add9a4

plugins/greenhouse/src/App.tsx
clementroche: This comes from CMS starter.

https://github.com/framer/plugins/blob/246882a7b8237e66e85ebbf416244e62c9ab1507/starters/cms/src/App.tsx#L22-L28

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: This comes from CMS starter

https://github.com/framer/plugins/blob/246882a7b8237e66e85ebbf416244e62c9ab1507/starters/cms/src/FieldMapping.tsx#L78-L82

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: This comes from CMS Starter

https://github.com/framer/plugins/blob/246882a7b8237e66e85ebbf416244e62c9ab1507/starters/cms/src/FieldMapping.tsx#L98-L101

plugins/greenhouse/src/components/FieldMapping.tsx
triozer: This can be updated here, it's just re-naming and I agree it's easier to read

triozer: So we can probably update the CMS Starter too at some point.

plugins/greenhouse/src/components/FieldMapping.tsx
kaloyanvi: We will update the starter in that case to implement more efficient solution for this. We should still address in this PR though. Should be something like this:

```tsx
                const existingFieldIds = new Set(collectionFields.map(field => field.id))

                if (initialSlugFieldId) {
                    const ignoredIds = new Set<string>()
                    for (const sourceField of dataSource.fields) {
                        if (existingFieldIds.has(sourceField.id)) continue
                        ignoredIds.add(sourceField.id)
                    }
                    setIgnoredFieldIds(ignoredIds)
                }
```

plugins/greenhouse/src/data.ts
triozer: This is a choice made because the `syncExistingCollection` method may return more than one _final state_. So to be future proof, we decided to make it like that in the CMS Starter. This discussion should not be part of this PR

plugins/greenhouse/src/data.ts
clementroche: This comes from CMS Starter

https://github.com/framer/plugins/blob/246882a7b8237e66e85ebbf416244e62c9ab1507/starters/cms/src/data.ts#L140

plugins/greenhouse/src/data.ts
clementroche: In FieldMapping view user has to choose which collection needs to be referenced in case several use the same "data source", that's why we need "supportedCollections". "collectionId" picks the first one of this list by default so it's prefilled.

In Framer CMS the collection ID is needed to reference a collection.

https://www.framer.com/developers/cms#adding-fields

plugins/greenhouse/src/data.ts
clementroche: This comes from CMS Starter.

https://github.com/framer/plugins/blob/246882a7b8237e66e85ebbf416244e62c9ab1507/starters/cms/src/data.ts#L153-L191

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: I don't think this is recomputed  on every render, it's computed only once.

https://ranveersequeira.medium.com/what-is-usestate-lazy-initialization-508a76cad8d6

plugins/greenhouse/src/App.tsx
clementroche: You can check the mockups:

https://framer.com/projects/Mocks--ZC4wx7CcLvmzhdm4S8XZ-2U5CZ?node=YdY14hNYR

plugins/greenhouse/src/dataSources.ts
clementroche: This is used to build the URL to fetch so technically "endpoint" is the right word

baseURL/endpoint

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: I wanted to keep it seperated since I got `changeFieldName` from CMS Starter and I thought it would be easier to maintain in the future.

https://github.com/framer/plugins/blob/main/starters/cms/src/FieldMapping.tsx#L118-L126

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: Indeed, I wanted to just extends the code from CMS Stater so it's easier to maintain

https://github.com/framer/plugins/blob/main/starters/cms/src/FieldMapping.tsx#L159

plugins/greenhouse/src/components/FieldMapping.tsx
kaloyanvi: Keeping them separate is fine. But big part is similar/same so that part should be reusable, we should be able to have a shared function `updateField` which can be reused by both `changeFieldName` and `changeCollectionId`

plugins/greenhouse/src/components/FieldMapping.tsx
kaloyanvi: This [bit](https://github.com/framer/plugins/blob/main/starters/cms/src/FieldMapping.tsx#L158-L160) in the CMS starter is done poorly and we shouldnt be doing multiple loops there.

plugins/greenhouse/src/data.ts
kaloyanvi: The reason for using set there is because we can use the set to make more efficient operations with it. See `unsyncedItems.delete(slugValue.value)`, the implementation here is not doing that.
You create the set on one line and immediately on the next is converted to an array.

plugins/greenhouse/src/data.ts
kaloyanvi: Ok, we can keep this as is. I don't think we should've implemented this return type in the starter but it is what it is.

plugins/greenhouse/src/dataSources.ts
kaloyanvi: I mean even `apiURL` is not great here tbh. But "endpoint" is not the right word for this. These are not complete API endpoints (which would include protocol, domain, etc. like https://api.greenhouse.io/v1/departments), but rather API paths that would be appended to a base URL.

It should be something like `"apiPath"`, `"path"` or `"resourcePath"`.

plugins/greenhouse/src/App.tsx
kaloyanvi: Requested access.

plugins/greenhouse/src/components/FieldMapping.tsx
kaloyanvi: Depending on the props of the component it might in future. The point here was that you are using `useState` when you shouldn't be. This should be done with `useMemo` which is exactly made for this use case.

clementroche: This comes from CMS Starter

kaloyanvi: Thanks, I'll add to list to address in cms starter.

plugins/greenhouse/src/utils.tsx
clementroche: Already tried but this increase a lot the complexity of types, not sure how to handle that. cc @triozer

kaloyanvi: The extra complexity gives us type safety then when handling collection references so we want that.

plugins/greenhouse/api
kaloyanvi, line 1: 👀 can we drop?

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,100 @@
...
+export interface Department {
+    id: number
+    name: string
+    jobs: Job[]
+}
+
+export interface Office {
+    id: number
+    name: string
+    location: string
```
kaloyanvi: shouldn't that be `location: Location`? Based on the API [docs](https://developers.greenhouse.io/harvest.html#patch-update-current-offer) seems like it but maybe theres something im missing

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export interface School {
+    id: number
+    text: string
+}
+
+export interface Discipline {
+    id: number
+    text: string
+}
```
kaloyanvi: what is the difference between the above 3 types? 🤔

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export interface DisciplinesResponse {
+    items: Discipline[]
+    meta: EducationMeta
+}
+
+export interface SectionsResponse {
+    sections: Section[]
+    meta: EducationMeta
+}
```
kaloyanvi: For the repetition we should reuse types. 
```suggestion
interface BaseResponse {
    items: (Degree | School | Discipline | Section)[]
    meta: EducationMeta
}

export interface DegreesResponse extends BaseResponse {
    items: Degree[]
}

export interface SchoolsResponse extends BaseResponse {
    items: School[]
}

export interface DisciplinesResponse extends BaseResponse {
    items: Discipline[]
}

export interface SectionsResponse extends BaseResponse {
    items: Section[]
}

```

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,100 @@
...
+export interface Job {
+    id: number
+    internal_job_id: number | null
+    title: string
+    updated_at: string
+    requisition_id: string
+    absolute_url: string
+    location: Location
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    metadata: any[]
```
kaloyanvi: why is this `any`? If we are not using it, it'd be better to use `unknown` because then if we use it, we'd be required to validate the type.

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export type GreenhouseItem = Job | Department | Office | Degree | School | Discipline | Section
+
+export function isGreenhouseItemField<T extends GreenhouseItem>(field: unknown, itemType: T): field is keyof T {
+    if (typeof field === "string" && field in itemType) {
+        return true
+    }
+
+    return false
```
kaloyanvi: This could be just:
```suggestion
    return typeof field === "string" && field in itemType
```

kaloyanvi: Actually this might not detect correctly optional types. We should do this instead because its more reliable:

```suggestion
    if (typeof field !== "string" || field === "") return false
    return Object.prototype.hasOwnProperty.call(itemType, field)
```

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,301 @@
...
+                d="M 3 11 L 6 8 L 3 5"
+                fill="transparent"
+                strokeWidth="1.5"
+                stroke="#999"
+                strokeLinecap="round"
+                strokeLinejoin="round"
+            />
+        </svg>
+    )
+}
```
kaloyanvi: Can we put this in separate file `plugins/greenhouse/src/components/Icons.tsx`?

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,301 @@
...
+                        </option>
+                    )}
+                    {field.supportedCollections?.map(collection => (
+                        <option key={collection.id} value={collection.id}>
+                            {collection.name}
+                        </option>
+                    ))}
+                </select>
+            )}
+            {!isCollectionReference(field) && (
```
kaloyanvi: we are checking twice if its collection reference on the same field. Can we define in const and use ternary here?

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,301 @@
...
+                tabIndex={0}
+            >
+                <input type="checkbox" checked={!disabled && !isMissingReference} tabIndex={-1} readOnly />
+                <span>{originalFieldName ?? field.id}</span>
+            </button>
+            <ChevronIcon />
+            {isCollectionReference(field) && (
+                <select
+                    className="target-field"
+                    disabled={disabled || isMissingReference}
```
kaloyanvi: multiple use cases of `disabled || isMissingReference`. Can we define in a const

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,301 @@
...
+                if (initialSlugFieldId) {
+                    const ignoredIds = new Set<string>()
+                    for (const sourceField of dataSource.fields) {
+                        if (existingFieldIds.has(sourceField.id)) continue
+                        ignoredIds.add(sourceField.id)
+                    }
+                    setIgnoredFieldIds(ignoredIds)
+                }
+
+                setStatus("mapping-fields")
```
kaloyanvi: shouldn't this be set on L133? 🤔

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+            if (updatedIgnoredFieldIds.has(fieldId)) {
+                updatedIgnoredFieldIds.delete(fieldId)
+            } else {
+                updatedIgnoredFieldIds.add(fieldId)
+            }
+
+            return updatedIgnoredFieldIds
+        })
+    }
```
kaloyanvi: we should add `useCallback` wrapper to each function since we are passing down via props

plugins/greenhouse/src/components/Loading.tsx
kaloyanvi, line 1: nice ❤️

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,268 @@
...
+                    break
+                }
+                case "collectionReference": {
+                    if (typeof value !== "object" || value == null || !("id" in value)) {
+                        continue
+                    }
+
+                    fieldData[field.id] = { value: String(value.id), type: "collectionReference" }
+                    break
+                }
```
kaloyanvi: aren't we missing some types here, `"image" | "file" | "enum"`? To prevent this from happenning again, can we define a function `shouldBeNever` and use in the default case? See in CMS Export plugin for example https://github.com/framer/plugins/blob/main/plugins/cms-export/src/assert.ts

```tsx
// ...
default:
   shouldBeNever(field.type)
```

kaloyanvi: actually `assertNever` (see and copy implementation from other plugins) is better here so we can infer the type. Then we can have type safety with something like this:

```tsx
switch (fieldType) {
                case "string":
                    fieldData[field.id] = { value: value ? String(value) : "", type: "string" }
                    break
                case "color":
                case "date":
                case "link":
                    fieldData[field.id] = { value: value ? String(value) : null, type: fieldType }
                    break
                case "number":
                    fieldData[field.id] = { value: Number(value), type: "number" }
                    break
                case "boolean":
                    fieldData[field.id] = { value: Boolean(value), type: "boolean" }
                    break
                case "formattedText":
                    fieldData[field.id] = { value: decodeHtml(String(value)), type: "formattedText" }
                    break
                    const ids: string[] = []
                    if (Array.isArray(value)) {
                        ids.push(...value.map(item => String(item.id)))
                    }
                    fieldData[field.id] = { value: ids, type: "multiCollectionReference" }
                    break
                }
                case "collectionReference": {
                    if (typeof value !== "object" || value == null || !("id" in value)) {
                        continue
                    }
                    fieldData[field.id] = { value: String(value.id), type: "collectionReference" }
                    break
                }
                case "image":
                    // whatever needs to happen here
                    break
                case "file":
                    // whatever needs to happen here
                    break
                case "enum":
                    // whatever needs to happen here
                    break
                default:
                    assertNever(fieldType)
            }
```

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,268 @@
...
+    return items
+}
+
+export async function syncCollection(
+    boardToken: string,
+    collection: ManagedCollection,
+    dataSource: GreenhouseDataSource,
+    fields: readonly ManagedCollectionFieldInput[],
+    slugField: ManagedCollectionFieldInput
+) {
```
kaloyanvi: Since this function is exported, can we add return type? Its a small TS optimization.

plugins/greenhouse/src/dataSources.ts
```diff
@@ -0,0 +1,289 @@
...
+
+            const degrees: Degree[] = []
+            for (let i = 0; i < pages; i++) {
+                const response = await fetch(`${url}?page=${i + 1}`)
+                const data = await response.json()
+
+                degrees.push(...data.items)
+            }
+
+            return degrees
```
kaloyanvi: this part is very similar for multiple data sources, can we create a fetch helper which has error handling and does this? The different return types can be solved quite nicely with function overload as well.

plugins/greenhouse/src/dataSources.ts
```diff
@@ -0,0 +1,289 @@
...
+const jobsDataSourceName = "Jobs"
+
+const departmentsDataSource = createDataSource(
+    {
+        name: "Departments",
+        apiPath: "departments",
+        fetch: async (boardToken: string) => {
+            const response = await fetch(`https://boards-api.greenhouse.io/v1/boards/${boardToken}/departments`)
+            const data = await response.json()
+            return data.departments as Department[]
```
kaloyanvi: why are we casting? We should add a validator to ensure the received json is of type `Department[]`. You should do something like:

```ts
function validateDepartments(data: unknown): asserts data is Department[] {
    if (!Array.isArray(data)) {
        throw new Error("Expected departments data to be an array")
    }
    
    for (const item of data) {
        if (typeof item !== "object" || item === null) {
            throw new Error("Expected department item to be an object")
        }
        if (typeof item.id !== "number") {
            throw new Error("Expected department to have numeric id")
        }
        if (typeof item.name !== "string") {
            throw new Error("Expected department to have string name")
        }
        if (!Array.isArray(item.jobs)) {
            throw new Error("Expected department to have jobs array")
        }
    }
}
```

and then something like:

```ts
            const response = await fetch(`https://boards-api.greenhouse.io/v1/boards/${boardToken}/departments`)
            const data: unknown = await response.json()
            validateDepartments(data)
            return data
```

kaloyanvi: We should also validate the other fetches and not only this one.

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: This comes from CMS Starter.

https://github.com/framer/plugins/blob/main/starters/cms/src/FieldMapping.tsx#L126-L148

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: This comes from Airtable plugin.

https://github.com/framer/plugins/blob/main/plugins/airtable/src/FieldMapping.tsx#L99-L120

plugins/greenhouse/src/data.ts
clementroche: `image` `file` and `enum` don't make sense/will never happen in Greenhouse Plugin

kaloyanvi: Indeed, that why for those we should set the value to `undefined` or something similar but we should still handle those cases. Then in future if we introduce new type, we will immediately have a type error pointing here and we will know what to do.

plugins/greenhouse/src/components/FieldMapping.tsx
kaloyanvi: The CMS starter has been updated to include those. Thats from the link you sent.
CleanShot 2025-06-25 at 10 04 59@2x

plugins/greenhouse/src/api-types.ts
clementroche: No, in that case it's a `string`.

Example: https://boards-api.greenhouse.io/v1/boards/overjet/offices

plugins/greenhouse/src/components/FieldMapping.tsx
clementroche: This comes from Airtable Plugin

https://github.com/framer/plugins/blob/main/plugins/airtable/src/FieldMapping.tsx#L10-L24

plugins/greenhouse/src/data.ts
clementroche: `Promise<void>` ?

plugins/greenhouse/src/components/FieldMapping.tsx
kaloyanvi: Yes and [this](https://github.com/framer/plugins/blob/main/plugins/google-sheets/src/components/Icons.tsx) comes from the google sheets plugin, doesn't mean anything. There is no point in highlighting a bad example which exists to advocate your reasoning, if there is a poor performance implementation somewhere which has not been noticed in review, this does not mean we should be doing that in every other plugin we are building!

plugins/greenhouse/src/data.ts
kaloyanvi: yep

plugins/greenhouse/src/api-types.ts
kaloyanvi: ok 👍🏻

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+        if (typeof item.location !== "object" || item.location === null) {
+            throw new Error("Expected job to have object location")
+        }
+        if (typeof item.location.name !== "string") {
+            throw new Error("Expected job to have string location.name")
+        }
+        if (typeof item.metadata !== "object" || item.metadata === null) {
+            throw new Error("Expected job to have object metadata")
+        }
```
kaloyanvi: What is the type of `item.metadata`? In the type is set to `unknown[]` but here we check of an object.  I think one of them is wrong.

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (typeof field !== "string" || field === "") return false
+    return Object.prototype.hasOwnProperty.call(itemType, field)
+}
+
+export function validateJobs(data: unknown): asserts data is Job[] {
+    if (!Array.isArray(data)) {
+        throw new Error("Expected jobs data to be an array")
+    }
+
+    for (const item of data) {
```
kaloyanvi: Are we not checking the `item.internal_job_id` on purpose? If so, we should add a comment about that. If forgotten we should add.

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,286 @@
...
+    return (
+        <>
+            <button
+                type="button"
+                className={`source-field ${isMissingReference && "missing-reference"}`}
+                aria-disabled={isDisabled}
+                onClick={() => onToggleDisabled(field.id)}
+                tabIndex={0}
+            >
+                <input type="checkbox" checked={!disabled && !isMissingReference} tabIndex={-1} readOnly />
```
kaloyanvi: we can now do `checked={!isDisabled}`

plugins/greenhouse/src/components/FieldMapping.tsx
```diff
@@ -0,0 +1,286 @@
...
+                        <option key={collection.id} value={collection.id}>
+                            {collection.name}
+                        </option>
+                    ))}
+                </select>
+            ) : (
+                <input
+                    type="text"
+                    className="target-field"
+                    disabled={disabled}
```
kaloyanvi: nit: can we add a tiny comment why this doesn't use `isDisabled` but just `disabled`?

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,274 @@
...
+    existingFields: readonly ManagedCollectionFieldInput[]
+): GreenhouseField[] {
+    return sourceFields.map(sourceField => {
+        const existingField = existingFields.find(existingField => existingField.id === sourceField.id)
+        if (existingField) {
+            return { ...sourceField, name: existingField.name }
+        }
+        return sourceField
+    })
+}
```
kaloyanvi: can we use a map here instead of `array.find()` in each iteration for better performance. Should be something like this:

```ts
export function mergeFieldsWithExistingFields(
    sourceFields: readonly GreenhouseField[],
    existingFields: readonly ManagedCollectionFieldInput[]
): GreenhouseField[] {
    const existingFieldsMap = new Map(existingFields.map(field => [field.id, field]))

    return sourceFields.map(sourceField => {
        const existingField = existingFieldsMap.get(sourceField.id)
        if (existingField) {
            return { ...sourceField, name: existingField.name }
        }
        return sourceField
    })
}
```

plugins/greenhouse/src/data.ts
```diff
@@ -0,0 +1,274 @@
...
+                        continue
+                    }
+
+                    fieldData[field.id] = { value: String(value.id), type: "collectionReference" }
+                    break
+                }
+                case "image":
+                case "file":
+                case "enum":
+                    break
```
kaloyanvi: can we throw error here instead, something like: 

```
throw new Error(`${field.type} field is not supported.`)
```

plugins/greenhouse/src/dataSources.ts
```diff
@@ -0,0 +1,279 @@
...
+    name: string
+    /**
+     * The fields of the data source.
+     *
+     * The first field is the ID field.
+     * The rest of the fields are the fields of the data source.
+     */
+    fields: readonly GreenhouseField[]
+    apiPath: string
+    fetch: (boardToken: string) => Promise<Job[] | Department[] | Office[] | Education[] | Section[]>
```
kaloyanvi: can we import `GreenhouseItem` and use here like this

```suggestion
    fetch: (boardToken: string) => Promise<GreenhouseItem[]>
```

plugins/greenhouse/src/dataSources.ts
```diff
@@ -0,0 +1,279 @@
...
+        for (let i = 0; i < pages; i++) {
+            const response = await fetch(`${url}?page=${i + 1}`)
+            const data = await response.json()
+            items.push(...(data[itemsKey] as unknown[]))
+        }
+    } else {
+        items.push(...(data[itemsKey] as unknown[]))
+    }
+
+    return items
```
kaloyanvi: can we add a try catch error handling here. Doesn't have to be super in-depth but we should have something.

plugins/greenhouse/src/utils.tsx
```diff
@@ -0,0 +1,25 @@
...
+    return textarea.value
+}
+
+export function isCollectionReference(
+    field: GreenhouseField
+): field is Extract<GreenhouseField, { type: "collectionReference" | "multiCollectionReference" }> {
+    return field.type === "collectionReference" || field.type === "multiCollectionReference"
+}
+
+export function isMissingReferenceField(field: GreenhouseField) {
```
kaloyanvi: can we add return type `boolean`

plugins/greenhouse/src/utils.tsx
```diff
@@ -0,0 +1,25 @@
+import type { GreenhouseField } from "./dataSources"
+
+export function decodeHtml(html: string) {
```
kaloyanvi: can we add return type to this

kaloyanvi (approved):
LGTM! Nice work 👏🏻

PR #254, opened 2025-06-20
[Main task] Replace Prettier with Biome

alecmev:
### Description

Fixes: https://github.com/framer/company/issues/33115

Commit-by-commit reviewing is encouraged!

Not adding to CMS Starter because it can't have another root Biome config, and it doesn't make sense to add a non-root one, because then it won't work when cloned. No formatting there for now, and will try to figure it out later, likely via `create-plugin` (which will add Biome as a dependency and copy the root config and rewrite the `check-biome` script).

Auto-fixed everything, but yeah, it's a huge diff, which will make every open PR problematic. Can slightly alleviate it by:

1. Disabling import sorting for now.
2. Adding local non-root configs that at least match the indentation and the line length.

But not sure if these half-measures will ultimately make it less painful. Can also wait until the major PRs are in and then format everything.

### Testing

- [x] No runtime changes, just check that `yarn check` works (and runs `check-biome`). Also check that formatting works on save + imports get auto-sorted.


.vscode/settings.json
```diff
@@ -1,12 +1,28 @@
 {
     "editor.rulers": [120],
     "eslint.workingDirectories": [{ "mode": "auto" }],
-    "biome.enabled": false,
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "biome.enabled": true,
+    "editor.defaultFormatter": "biomejs.biome",
```
triozer, line 5: Should we add to recommended extensions then?

starters/cms/.prettierrc.json
triozer, line 1: We should probably keep prettier in the starter for now. Better than nothing

triozer (approved):
LGTM

.vscode/settings.json
alecmev: Yep, done!

starters/cms/.prettierrc.json
alecmev: Yeah, doesn't feel right to leave it hanging if we aren't tackling it immediately, re-added.

PR #255, opened 2025-06-23
Preserve draft status in CMS import/export

madebyisaacr:
### Description

This pull request adds draft support to CSV Import and CMS Export plugins. A new column named `:draft` has been added, which controls the draft status of each CMS item.

It's named ":draft" instead of "Draft" to avoid naming conflicts with a CMS field named "Draft". A similar syntax "Image:alt" is used for image alt text columns.

In the export plugin, the `:draft` column is only included if there are any draft items.

In the import plugin, if the `:draft` column is missing all items are imported as non-drafts.

With this update, you can export a CMS collection with draft items, and when you import them into a new collection the same items will be drafts.

### Testing

- [x] Create a new blog collection.
- [x] Mark some items as draft.
- [x] Run the CMS Export plugin with this PR and save a copy of the collection.
- [x] Add a new collection.
- [x] Run the CSV Import plugin with this PR and import the saved copy of the other collection.
- [x] The same items are marked as draft.
- [x] You can also copy and paste from this sheet to test: https://docs.google.com/spreadsheets/d/1Z0fnCfTMj7NOTBL8Dh2fv73Xa6k1LTH1l8CQDqY3fGc/edit?usp=sharing

### DIFF
--- a/plugins/cms-export/src/PreviewTable.tsx
+++ b/plugins/cms-export/src/PreviewTable.tsx
@@ -6,6 +6,10 @@ import { getDataForCSV } from "./csv"
 
 import "./PreviewTable.css"
 
+const cellTitleOverrides: Record<string, string> = {
+    ":draft": "Is Draft?",
+}
+
 interface Props {
     collection: Collection
 }
@@ -62,7 +66,7 @@ export function PreviewTable({ collection }: Props) {
                         <tr>
                             {previewCSV[0]?.map((cell, columnIndex) => (
                                 <td key={`0-${columnIndex}`} title={cell}>
-                                    {cell === ":draft" ? "Is Draft?" : cell}
+                                    {cellTitleOverrides[cell] ?? cell}
                                 </td>
                             ))}
                         </tr>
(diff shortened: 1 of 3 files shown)


madebyisaacr:
@triozer "Is Draft?" is a nice change, I'll add that.

We shouldn't change booleans to Yes/No because it exports to CSV as true/false, so the exported data wouldn't match the preview.

plugins/cms-export/src/PreviewTable.tsx
```diff
@@ -62,7 +62,7 @@ export function PreviewTable({ collection }: Props) {
                         <tr>
                             {previewCSV[0]?.map((cell, columnIndex) => (
                                 <td key={`0-${columnIndex}`} title={cell}>
-                                    {cell}
+                                    {cell === ":draft" ? "Is Draft?" : cell}
```
triozer: What do you think of having an object `cellTitleOverrides` so that you can have
```ts
const cellTitleOverrides: Record<string, string> = {
    ":draft": "Is Draft?"
}
```
and use it here like `{cellTitleOverrides[cell] ?? cell}`

PR #257, opened 2025-06-24
[Continuation] Add Greenhouse Plugin with Review Fixes (#253)

clementroche:
This PR continues the work from #253 and addresses the review comments. Since I don’t have push access to the original branch, this is a continuation PR.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1916,142 +1916,142 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@rollup/rollup-android-arm-eabi@npm:4.44.0":
-  version: 4.44.0
-  resolution: "@rollup/rollup-android-arm-eabi@npm:4.44.0"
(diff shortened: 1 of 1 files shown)


kaloyanvi (approved):
Approving so this can merge into #253, will do the review on #253.

plugins/greenhouse/src/api-types.ts
```diff
@@ -0,0 +1,100 @@
...
+export interface Job {
+    id: number
+    internal_job_id: number | null
+    title: string
+    updated_at: string
+    requisition_id: string
+    absolute_url: string
+    location: Location
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    metadata: any[]
```
kaloyanvi, line 27: why is this `any`? Do we never use it, if so it's better if we do `metadata: unknown[]` in case we ever do because it will require us to validate the types.

PR #262, opened 2025-06-25
[Side] Update dependencies

alecmev:
### Description

Long-overdue. Skipped updating Tailwind in Notion as it's being dropped there. All Tailwind-related changes are done by their codemod, not hand (except for dropping PostCSS and fixing CSS specificity by importing `framer-plugin` CSS in the base layer rather than at the top). Added `check-svelte` and fixed the errors because it was a 2-minute job.

### Testing

- [x] Run any Tailwind-based plugin, check that it works and looks good.
- [x] Run Renamer, as it had a relatively big Svelte jump.

### DIFF
--- a/plugins/hubspot/src/globals.css
+++ b/plugins/hubspot/src/globals.css
@@ -1,7 +1,7 @@
 @import url("https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap");
 @import "tailwindcss";
 
-@custom-variant dark (&:is([data-framer-theme='dark'] *));
+@custom-variant dark (&:where([data-framer-theme="dark"], [data-framer-theme="dark"] *));
 
 @theme inline {
     --background-color-primary: var(--framer-color-bg);


triozer:
The tailwind upgrade seems to broke Google Sheets 👀 

image

plugins/notion/package.json
```diff
@@ -13,19 +13,19 @@
         "check-typescript": "run g:check-typescript"
     },
     "dependencies": {
-        "@notionhq/client": "^2.2.15",
-        "@tanstack/react-query": "^5.29.2",
+        "@notionhq/client": "^2.3.0",
```
RayJason, line 16: Record a point: the latest version of the `@notionhq/client` is v3.1.3 
Maybe you can commit a separate branch to upgrade

alecmev: Thanks for pointing out, this will be updated (along with dropping Tailwind) in/after the Notion refactor: https://github.com/framer/plugins/pull/221

plugins/hubspot/src/globals.css
```diff
@@ -1,75 +1,124 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import url("https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap");
+@import "tailwindcss";
 
+@custom-variant dark (&:is([data-framer-theme='dark'] *));
```
triozer: nit. funny how there is two way of doing the same thing

```suggestion
@custom-variant dark (&:where([data-framer-theme="dark"], [data-framer-theme="dark"] *));
```

triozer (approved):
LGTM!

plugins/hubspot/src/globals.css
alecmev: Ah, yeah, it was declared slightly differently in the old configs, fixed!

PR #264, opened 2025-06-26
Add zed config

tom-james-watson:
### Description

Adds repo-relevant settings for Zed. Zed would default to using 

### Testing

- [ ] Use Zed, open a typescript file, make an edit and confirm that biome formats correctly, ie not with prettier.

### DIFF
--- a/.zed/settings.json
+++ b/.zed/settings.json
@@ -1,0 +1,44 @@
+{
+    "languages": {
+        "JSONC": {
+            "formatter": {
+                "language_server": {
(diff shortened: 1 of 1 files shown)


.zed/settings.json
triozer, line 1: Can we add a format on save option?

tom-james-watson: It should by default

triozer: Indeed had to install the Biome extension

PR #265, opened 2025-06-26
Add support for galleries to notion

tom-james-watson:
### Description

Adds support for galleries to Notion.

### Testing

  - [ ] Step 1
  - [ ] Step 2
  - [ ] Step 3
  - [ ] Step 1
  - [ ] Step 2
  - [ ] Step 3

### DIFF
--- a/plugins/notion/src/MapFields.tsx
+++ b/plugins/notion/src/MapFields.tsx
@@ -110,6 +110,7 @@ const labelByFieldTypeOption: Record<ManagedCollectionField["type"], string> = {
     string: "String",
     collectionReference: "Reference",
     multiCollectionReference: "Multi Reference",
+    array: "Gallery",
 }
 
 export function MapDatabaseFields({
@@ -141,7 +142,6 @@ export function MapDatabaseFields({
     )
 
     // TODO: Render progress in UI.
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const [_, setProgress] = useState<SynchronizeProgress | null>(null)
 
     assert(isFullDatabase(database))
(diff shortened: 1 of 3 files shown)


plugins/notion/src/data.ts
```diff
@@ -401,6 +412,40 @@ export function getFieldDataEntryForProperty(
...
+                const imageField = field.fields[0]
+
+                return {
+                    type: "array",
+                    value: property.files
+                        .map((file): ArrayItemInput | undefined => {
+                            switch (file?.type) {
+                                case "external":
+                                    return {
+                                        id: file.name,
```
tom-james-watson, line 467: This isn't ideal, but we don't have a real ID here though. The file names as far as I can tell are gibberish asset names and are unique.

plugins/notion/src/data.ts
```diff
@@ -401,6 +412,40 @@ export function getFieldDataEntryForProperty(
...
         }
         case "files": {
+            if (field.type === "array") {
+                const imageField = field.fields[0]
+
+                return {
+                    type: "array",
+                    value: property.files
+                        .map((file): ArrayItemInput | undefined => {
+                            switch (file?.type) {
```
niekert: nit: could we add an exhaustive check on the file type and filter out the undefined values first?

niekert:
Not sure if there's anything we can do on the Plugin side, but isn't it a little weird that a broken image gets uploaded because there's also an `xliff` file in in the files column?

We don't get the mime type for files in the Notion API right?

image

niekert (approved):
Nice 🚀

plugins/notion/src/data.ts
tom-james-watson: Made it exhaustive. Can't filter first without making the code uglier

plugins/notion/package.json
```diff
@@ -14,7 +14,7 @@
     },
     "dependencies": {
         "@notionhq/client": "^3.1.3",
-        "framer-plugin": "^3.3.2",
+        "framer-plugin": "^3.4.0-alpha.0",
```
madebyisaacr:
```suggestion
        "framer-plugin": "^3.4.0",
```

PR #271, opened 2025-06-30
Airtable: add Lookup, Rollup, Barcode, AI Text, User, and Duration fields

madebyisaacr:
### Description

This PR adds support for more Airtable field types and improves other field types. v2 of https://github.com/framer/plugins/pull/248

* Added support for these Airtable field types:
  * Lookup
  * Rollup
  * Barcode
  * AI Text
  * Collaborator
  * Created By
  * Last Edited By
  * Count
* Allow importing URL as Plain Text field.
* Fixed warning spam for empty number fields.
* Fixed importing duration field. Previously it would only import as a number of seconds, now it imports as a string by default and uses the same hour:minute:second formatting as shown in the Airtable app.

**Notes:**
* User/collaborator fields are imported as the user's name (same as the Notion plugin).
* Lookup array fields only import the first value. When we have support for gallery fields and later arrays for all field types, we can update this to import as an array instead of only the first value.

Currently marked as draft. Will open for review after I have tested all changes.

### Testing

None yet.

### DIFF
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -234,8 +234,16 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                         return { ...field, type: "string" } as PossibleField
                     case "formattedText":
                         return { ...field, type: "formattedText" } as PossibleField
+                    case "number":
+                        return { ...field, type: "number" } as PossibleField
+                    case "boolean":
+                        return { ...field, type: "boolean" } as PossibleField
                     case "color":
                         return { ...field, type: "color" } as PossibleField
+                    case "date":
+                        return { ...field, type: "date" } as PossibleField
+                    case "enum":
+                        return { ...field, type: "enum" } as PossibleField
                     default:
                         return field
                 }
@@ -274,10 +282,13 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
 
             const fieldsToSync = fields
                 .filter(field => !ignoredFieldIds.has(field.id))
-                .map(field => ({
-                    ...field,
-                    name: field.name.trim() || field.id,
-                }))
+                .map(field => {
+                    const originalFieldName = dataSource.fields.find(sourceField => sourceField.id === field.id)?.name
+                    return {
+                        ...field,
+                        name: field.name.trim() || originalFieldName || field.id,
+                    }
+                })
                 .filter(field => field.type !== "unsupported")
                 .filter(
                     field =>
@@ -399,13 +410,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                     tabIndex={0}
                     title={isAllowedToManage ? undefined : "Insufficient permissions"}
                 >
-                    {isSyncing ? (
-                        <div className="framer-spinner" />
-                    ) : (
-                        <span>
-                            Import <span style={{ textTransform: "capitalize" }}>{dataSource.tableName}</span>
-                        </span>
-                    )}
+                    {isSyncing ? <div className="framer-spinner" /> : <span>Import {dataSource.tableName}</span>}
                 </button>
             </footer>
         </form>
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -1,6 +1,6 @@
 import type { Field, ManagedCollection, ManagedCollectionFieldInput } from "framer-plugin"
 import { framer, useIsAllowedTo } from "framer-plugin"
-import { memo, useEffect, useState } from "react"
+import { memo, useEffect, useMemo, useState } from "react"
 import type { DataSource } from "./data"
 import { mergeFieldsWithExistingFields, syncCollection, syncMethods } from "./data"
 import type { PossibleField } from "./fields"
@@ -164,6 +164,12 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
     const [fields, setFields] = useState(initialManagedCollectionFields)
     const [ignoredFieldIds, setIgnoredFieldIds] = useState(initialFieldIds)
 
+    // Create a map of field IDs to names for efficient lookup
+    const originalFieldNameMap = useMemo(
+        () => new Map(dataSource.fields.map(field => [field.id, field.name])),
+        [dataSource.fields]
+    )
+
     useEffect(() => {
         const abortController = new AbortController()
 
@@ -283,7 +289,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
             const fieldsToSync = fields
                 .filter(field => !ignoredFieldIds.has(field.id))
                 .map(field => {
-                    const originalFieldName = dataSource.fields.find(sourceField => sourceField.id === field.id)?.name
+                    const originalFieldName = originalFieldNameMap.get(field.id)
                     return {
                         ...field,
                         name: field.name.trim() || originalFieldName || field.id,
@@ -370,7 +376,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                         <FieldMappingRow
                             key={`field-${field.id}`}
                             field={field}
-                            originalFieldName={dataSource.fields.find(sourceField => sourceField.id === field.id)?.name}
+                            originalFieldName={originalFieldNameMap.get(field.id)}
                             isIgnored={ignoredFieldIds.has(field.id)}
                             disabled={!isAllowedToManage}
                             onToggleIgnored={toggleFieldIgnoredState}
@@ -383,7 +389,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                         key={`field-${field.id}`}
                         field={field}
                         missingCollection
-                        originalFieldName={dataSource.fields.find(sourceField => sourceField.id === field.id)?.name}
+                        originalFieldName={originalFieldNameMap.get(field.id)}
                         isIgnored={true}
                         disabled={!isAllowedToManage}
                     />
@@ -395,7 +401,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                             key={`field-${field.id}`}
                             field={field}
                             unsupported
-                            originalFieldName={dataSource.fields.find(sourceField => sourceField.id === field.id)?.name}
+                            originalFieldName={originalFieldNameMap.get(field.id)}
                             isIgnored={true}
                             disabled={!isAllowedToManage}
                         />
--- a/plugins/airtable/src/api.ts
+++ b/plugins/airtable/src/api.ts
@@ -511,3 +511,16 @@ export const fetchAllBases = async () => {
 
     return allBases
 }
+
+
+export function isBarcodeValue(value: unknown): value is BarcodeValue {
+    return !!(value && typeof value === "object" && "text" in value && typeof value.text === "string")
+}
+
+export function isAiTextValue(value: unknown): value is AiTextValue {
+    return !!(value && typeof value === "object" && "value" in value && typeof value.value === "string")
+}
+
+export function isCollaboratorValue(value: unknown): value is CollaboratorValue {
+    return !!(value && typeof value === "object" && "name" in value && typeof value.name === "string")
+}
--- a/plugins/airtable/src/api.ts
+++ b/plugins/airtable/src/api.ts
@@ -54,7 +54,6 @@ type CountValue = number
 type DateValue = string
 type DateTimeValue = string
 type PhoneNumberValue = string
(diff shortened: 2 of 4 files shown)


plugins/airtable/src/api.ts
```diff
@@ -54,7 +54,6 @@ type CountValue = number
 type DateValue = string
 type DateTimeValue = string
 type PhoneNumberValue = string
-type LookupValue = Array<string | number | boolean>
```
madebyisaacr, line 57: Removed the lookup type because it doesn't exist according to the Airtable API docs: https://airtable.com/developers/web/api/field-model#lookup

"multipleLookupValues" is the only lookup type.

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -274,10 +282,13 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
                 .filter(field => !ignoredFieldIds.has(field.id))
-                .map(field => ({
-                    ...field,
-                    name: field.name.trim() || field.id,
-                }))
+                .map(field => {
+                    const originalFieldName = dataSource.fields.find(sourceField => sourceField.id === field.id)?.name
+                    return {
+                        ...field,
+                        name: field.name.trim() || originalFieldName || field.id,
```
niekert: possibly negligible due to small size of the fields array but creating a map keyed by field ID for faster lookup could prevent O(n×m) complexity

plugins/airtable/src/data.ts
```diff
@@ -122,6 +128,74 @@ function getFieldDataEntryForFieldSchema(fieldSchema: PossibleField, value: unkn
...
+                case "barcode": {
+                    if (!value || typeof value !== "object" || !("text" in value) || typeof value.text !== "string")
+                        return null
+                    return {
+                        value: value.text,
+                        type: "string",
+                    }
+                }
+                case "aiText": {
+                    if (!value || typeof value !== "object" || !("value" in value) || typeof value.value !== "string")
```
niekert: nit: maybe coudl create some type guards for these? 

Is it expected that one is `value.text` and the other becomes `value.value`?

plugins/airtable/src/FieldMapping.tsx
```diff
@@ -234,8 +234,16 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
...
+                    case "number":
+                        return { ...field, type: "number" } as PossibleField
+                    case "boolean":
+                        return { ...field, type: "boolean" } as PossibleField
                     case "color":
                         return { ...field, type: "color" } as PossibleField
+                    case "date":
+                        return { ...field, type: "date" } as PossibleField
+                    case "enum":
+                        return { ...field, type: "enum" } as PossibleField
```
niekert, line 252: nit: i know its existing code but are all these `as` casts needed here?

plugins/airtable/src/data.ts
```diff
@@ -122,6 +128,74 @@ function getFieldDataEntryForFieldSchema(fieldSchema: PossibleField, value: unkn
...
+
+                    return {
+                        value: result,
+                        type: "string",
+                    }
+                }
+                case "singleCollaborator":
+                case "createdBy":
+                case "lastModifiedBy": {
+                    if (!value || typeof value !== "object" || !("name" in value) || typeof value.name !== "string")
```
niekert: maybe also nice to separate into a typeguard fn

plugins/airtable/src/data.ts
```diff
@@ -356,13 +436,26 @@ export async function syncExistingCollection(
...
+        // Use properly inferred fields instead of existing collection fields
+        const inferredFields = await inferFields(collection, table)
+
+        // Filter fields to match the format expected by syncCollection
+        const fieldsToSync = inferredFields.filter(
+            field =>
+                existingFields.find(existingField => existingField.id === field.id) &&
+                field.type !== "unsupported" &&
+                ((field.type !== "collectionReference" && field.type !== "multiCollectionReference") ||
+                    field.collectionId !== "")
```
niekert: nit: kinda hard to read this one

plugins/airtable/src/fields.ts
```diff
@@ -160,7 +201,102 @@ function inferDateField(
...
+    return {
+        id: fieldSchema.id,
+        name: fieldSchema.name,
+        userEditable: false,
+        airtableType: fieldSchema.type,
+        type: "string",
+        allowedTypes: ["string"],
+        ...createFieldMetadata(fieldSchema),
+    }
+}
```
niekert: These all result in the same string `PossibleField`. Are you keeping them separate in case they're gonna support different field types in the future?

plugins/airtable/src/fields.ts
```diff
@@ -160,7 +201,102 @@ function inferDateField(
...
+    if (!options.result || typeof options.result !== "object" || !options.result.type) {
+        return createUnsupportedField(fieldSchema)
     }
+
+    // Use the result object as the field schema for recursive inference
+    const resultSchema = {
+        type: options.result.type as AirtableFieldSchema["type"],
+        id: fieldSchema.id,
+        name: fieldSchema.name,
+        options: options.result.options as AirtableFieldSchema["options"],
```
niekert: Since we are maintaining the types, does it need the typecast? Can't we align the type here to be `AirTsbleFieldSchema["options"]` if that is what it is?

plugins/airtable/src/fields.ts
```diff
@@ -160,7 +201,102 @@ function inferDateField(
...
+    } as AirtableFieldSchema
+
+    // Use the helper functions to infer the appropriate type based on the result
+    const inferredField = await inferFieldByType(resultSchema, collection, tableIdBeingLinkedTo, 1)
+
+    // Return the inferred field with the lookup field metadata
+    return { ...inferredField, originalAirtableType: "multipleLookupValues" } as PossibleField
```
niekert: don't need the typecast I think?

plugins/airtable/src/fields.ts
```diff
@@ -160,7 +201,102 @@ function inferDateField(
...
+        id: fieldSchema.id,
+        name: fieldSchema.name,
+        type: options.result.type as AirtableFieldSchema["type"],
+        options: options.result.options as AirtableFieldSchema["options"],
+    } as AirtableFieldSchema
+
+    // Use the helper functions to infer the appropriate type based on the result
+    const inferredField = await inferFieldByType(resultSchema, collection, tableIdBeingLinkedTo, 1)
+
+    // Return the inferred field with the rollup field metadata
```
niekert: supernit: this comment rather states the obvious, I see you AI hehe

niekert (commented):
code LGTM besides some nits. Not too familiar with Airtable, any easy ways to QA? Since the PR description doesn't have those yet

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: It gets mad when you remove them.

image

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: Done - added a map of field ids to names and replaced find with it.

plugins/airtable/src/data.ts
madebyisaacr: What does a type guard look like?

Yes, Airtable returns `value.text` for barcode and `value.value` for AI text.

madebyisaacr: Nevermind about the type guard functions.

plugins/airtable/src/fields.ts
madebyisaacr: No that was an oversight, not intentional. Fixed.

plugins/airtable/src/fields.ts
madebyisaacr: It works without the type cast. Removed `as` here and a few other places where it wasn't needed.

plugins/airtable/src/fields.ts
```diff
@@ -160,9 +201,74 @@ function inferDateField(
...
+        options: options.result.options,
+    } as AirtableFieldSchema
+
+    const inferredField = await inferFieldByType(resultSchema, collection, tableIdBeingLinkedTo, 1)
+
+    return { ...inferredField, originalAirtableType: "multipleLookupValues" }
+}
+
+async function inferRollupField(
+    fieldSchema: AirtableFieldSchema & { type: "rollup" },
```
niekert: nitpick (feel free to ignore): we could define a utility type somewhere
```
type FieldSchema<T extends AirtableFieldSchema["type"]> = Extract<AirtableFieldSchema, { type: T }>
```

and use it here and in those other places like so

```
fieldSchema: FieldSchema<"rollup">
```

The nice thing is autocomplete for the `type`, it won't allow values besides the possible types.

plugins/airtable/src/api.ts
```diff
@@ -197,7 +195,10 @@ interface FormulaOption {
 interface RollupOption {
     fieldIdInLinkedTable?: string
     recordLinkFieldId?: string
-    result?: AirtableFieldType | null
+    result?: {
+        type: AirtableFieldType
+        options?: AirtableFieldOptions[AirtableFieldType]
+    } | null
```
niekert: You could define this as the following so that the `options` are inferred correctly based on the `type`

```
type AirtableField<T extends AirtableFieldType> = Extract<AirtableFieldSchema, { type: T }>
type AirtableFieldResult = {
    [K in AirtableFieldType]: Pick<AirtableField<K>, "type" | "options">
}[AirtableFieldType]
```

niekert: Also, why is `options` optional here? Because in `inferLookupField` we perform a type cast which assumes the options are not there

plugins/airtable/src/fields.ts
```diff
@@ -160,9 +201,74 @@ function inferDateField(
...
+    if (!options.result || typeof options.result !== "object" || !options.result.type) {
+        return createUnsupportedField(fieldSchema)
+    }
+
+    const resultSchema = {
+        type: options.result.type,
+        id: fieldSchema.id,
+        name: fieldSchema.name,
+        options: options.result.options,
+    } as AirtableFieldSchema
```
niekert: I know its an existing pattern in the Plugin, but the type cast is bothering me a bit because it feels unsafe. As it now assumes that `options` is always defined. 

You could fix this by making sure we return an unsupported field when options is empty, as the types reflect that it can be but its not checked here.

```
function hasOptions(
    result: AirtableComputedFieldResult
): result is AirtableComputedFieldResult & { options: NonNullable<AirtableComputedFieldResult["options"]> } {
    return result.options !== undefined && result.options !== null
}

if (!hasOptions(options.result)) {
  return createUnsupportedField(fieldSchema)
}

const resultSchema: AirtableFieldSchema = {
        ...options.result,
        id: fieldSchema.id,
        name: fieldSchema.name,
    }
```

plugins/airtable/src/data.ts
```diff
@@ -372,3 +469,19 @@ export async function syncExistingCollection(
...
+
+function isBarcodeValue(value: unknown): value is { text: string } {
+    return !!(value && typeof value === "object" && "text" in value && typeof value.text === "string")
+}
+
+function isAiTextValue(value: unknown): value is { value: string } {
+    return !!(value && typeof value === "object" && "value" in value && typeof value.value === "string")
+}
+
+function isDurationValue(value: unknown): value is number {
```
niekert: nit: this is just an `isNumber()` type guard

plugins/airtable/src/data.ts
```diff
@@ -372,3 +469,19 @@ export async function syncExistingCollection(
         return { didSync: false }
     }
 }
+
+function isBarcodeValue(value: unknown): value is { text: string } {
```
niekert: Instead of doing `is { text: string }` can we guard that this is `value is BarcodeValue` and `value is AiTextValue` from the types in `api.ts`?

plugins/airtable/src/FieldMapping.tsx
niekert: ok! ya that type is a a little crazy. lets keep it like that then since this is the existing pattern.

niekert (commented):
Just few small nits on types - I know this is existing code but I think its good to make it a little safer for the future.

plugins/airtable/src/FieldMapping.tsx
triozer: when I'll have a bit of time, I'll look into the CMS Starter and some ways of fixing this ugly `as`

plugins/airtable/src/fields.ts
niekert: @madebyisaacr I see other comments were addressed and would love to land this one. Was this one possible to address or did you run into type errors?

madebyisaacr: @niekert done

niekert (approved):
amazing 🚀 

I QA'd the PR also by syncing the product catalog you shared in the comment - all good

PR #272, opened 2025-07-03
Add Code Versions Plugin

elmarburke:
### Description

This PR adds a new Code-Versions plugin that allows users to browse and restore previous versions of Code Files using the Code File API.

**Key Features:**
- Browse version history of code components
- View diff between current and selected versions
- Restore previous versions with one click

**Technical Implementation:**
- Uses the Code File API (`CodeFile.getVersions()`, `CodeFileVersion.getContent()`, etc)
- State-driven architecture with React hooks and reducers
- Abort controllers for async operations

### Testing

- [ ] **Basic Version Browsing Canvas**
  - Open plugin with a code component selected from the canvas
  - Verify version history loads and displays correctly
  - Select different versions and verify content loads
  - Verify diff view shows differences between versions

- [ ] **Basic Version Browsing Code**
  - Open plugin with an open code file
  - Verify version history loads and displays correctly
  - Select different versions and verify content loads
  - Verify diff view shows differences between versions

- [ ] **Version Restoration**
  - To see the reset best, open a code file in the code editor
  - Select a previous version that differs from current
  - Click "Restore" button
  - Verify file content is updated in the code editor
  - Verify plugin closes automatically after successful restore


plugins/code-versions/vite.config.ts
alecmev, line 1: Not needed 😌

turbo.json
```diff
@@ -30,6 +30,10 @@
             "dependsOn": ["build"],
             "cache": false,
             "persistent": true
+        },
+        "test": {
+            "dependsOn": ["build"],
+            "cache": false
```
alecmev: I think this, no? Vitest will compile stuff on its own, and it's okay to cache, as long as the inputs are correct (which is usually the deps and the sources).

```suggestion
```

plugins/code-versions/package.json
```diff
@@ -0,0 +1,41 @@
...
+        "@radix-ui/react-tooltip": "^1.2.7",
+        "@tailwindcss/vite": "^4.1.11",
+        "clsx": "^2.1.1",
+        "diff": "^8.0.2",
+        "framer-plugin": "^3.3.3-alpha.2",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
+        "tailwind-merge": "^3.3.1",
+        "tailwindcss": "^4.1.11",
+        "ts-pattern": "^5.0.5"
```
alecmev, line 24: Pattern-matching is cool, but, this being a "one of many" library, is it worth the cognitive overhead for potential contributors here? I'd wager many more know how `switch-case` + `satisfies never` works.

`InlineDiffs` does a bit more than just a `switch`, but, at a glance, I feel like it could easily be rewritten as a couple of conditions and might actually end up being more succinct.

Don't mean to be "no fun allowed" 😄 A little on the fence about this. @framer/plugins?

plugins/code-versions/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "dependencies": {
+        "@radix-ui/react-tooltip": "^1.2.7",
+        "@tailwindcss/vite": "^4.1.11",
+        "clsx": "^2.1.1",
+        "diff": "^8.0.2",
+        "framer-plugin": "^3.3.3-alpha.2",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
+        "tailwind-merge": "^3.3.1",
+        "tailwindcss": "^4.1.11",
```
alecmev: Not needed, I think!

plugins/code-versions/package.json
```diff
@@ -0,0 +1,41 @@
...
+        "preview": "run g:preview",
+        "pack": "npx framer-plugin-tools@latest pack",
+        "check-typescript": "run g:check-typescript",
+        "test": "vitest",
+        "test:ui": "vitest --ui",
+        "test:run": "vitest run"
+    },
+    "dependencies": {
+        "@radix-ui/react-tooltip": "^1.2.7",
+        "@tailwindcss/vite": "^4.1.11",
```
alecmev: Neither 😛

plugins/code-versions/package.json
```diff
@@ -0,0 +1,41 @@
...
+        "check-eslint": "run g:check-eslint",
+        "preview": "run g:preview",
+        "pack": "npx framer-plugin-tools@latest pack",
+        "check-typescript": "run g:check-typescript",
+        "test": "vitest",
+        "test:ui": "vitest --ui",
+        "test:run": "vitest run"
+    },
+    "dependencies": {
+        "@radix-ui/react-tooltip": "^1.2.7",
```
alecmev, line 17: Consider [React Aria](https://react-spectrum.adobe.com/react-aria/Tooltip.html)! It's better-maintained, and I also noticed that we started using it in FramerStudio.

plugins/code-versions/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Code Versions Plugin</title>
```
alecmev: I think we started dropping the "Plugin" here. I think?

plugins/code-versions/index.html
alecmev, line 1: Link `logo.svg`?

package-lock.json
alecmev, line 1: `rm package-lock.json` 😛

plugins/code-versions/src/styles.css
```diff
@@ -0,0 +1,34 @@
+@import "framer-plugin/framer.css";
```
alecmev: Hmm, I think this won't work well, because unlayered rules take precedence over layered ones, which is all of Tailwind. Take a look at `globals.css` in HubSpot, for inspiration.

Or is that why Tailwind has `important`?

yarn.lock
```diff
@@ -43,17 +68,17 @@ __metadata:
...
+  resolution: "@biomejs/biome@npm:2.0.6"
+  dependencies:
+    "@biomejs/cli-darwin-arm64": "npm:2.0.6"
+    "@biomejs/cli-darwin-x64": "npm:2.0.6"
+    "@biomejs/cli-linux-arm64": "npm:2.0.6"
+    "@biomejs/cli-linux-arm64-musl": "npm:2.0.6"
+    "@biomejs/cli-linux-x64": "npm:2.0.6"
+    "@biomejs/cli-linux-x64-musl": "npm:2.0.6"
+    "@biomejs/cli-win32-arm64": "npm:2.0.6"
+    "@biomejs/cli-win32-x64": "npm:2.0.6"
```
alecmev: Did you update the deps at some point? 🤔 If so, not through `yarn upgrade-interactive`? Because I see that `package.json` remained untouched. Maybe drop these for now (by reverting this file to `main` and then running `yarn` again), unless there's something broken? To make this PR a little bit more focused.

plugins/code-versions/src/main.tsx
```diff
@@ -0,0 +1,16 @@
+import "framer-plugin/framer.css"
```
alecmev: Imported here and in `styles.css`?

alecmev (commented):
Some non-meat-and-gravy-related comments 😌

alecmev:
Big W re the introduction of tests, by the way! Been having shower thoughts about the lack of them for a while.

plugins/code-versions/src/App.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    const { state: fileStatus } = useSelectedCodeFile()
+    const { state, selectVersion, restoreVersion, clearErrors } = useCodeFileVersions()
+
+    // Close plugin when restore is completed successfully
+    useEffect(() => {
+        if (state.restoreCompleted) {
+            framer.closePlugin()
+        }
+    }, [state.restoreCompleted])
```
kaloyanvi, line 29: is that the intended behaviour?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,162 @@
...
+export enum StatusTypes {
+    INITIAL = "initial",
+    SELECTED_CODE_FILE = "selectedCodeFile",
+    NO_SELECTION = "noSelection",
+    ERROR = "error",
+}
+
+type InitialState = {
+    type: StatusTypes.INITIAL
+}
```
kaloyanvi: Conceptually, what is the difference between `StatusTypes.INITIAL` and `StatusTypes.NO_SELECTION`?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,162 @@
...
+
+type ErrorState = {
+    type: StatusTypes.ERROR
+    error: string
+}
+
+type State = InitialState | CodeFileState | NoSelectionState | ErrorState
+
+// Hook to handle Framer selection changes
+export function useSelectedCodeFile() {
```
kaloyanvi: can we add return type since this is exported

plugins/code-versions/vite.config.ts
elmarburke: You mean the file? I need it for the tailwind plugin, no?

alecmev: Nope, the shared Vite config already has Tailwind!

plugins/code-versions/package.json
elmarburke: I was considering React Aria, but Radix' Tooltop is already used in the examples here.

package-lock.json
elmarburke: 🙈

plugins/code-versions/package.json
alecmev: Yeah, but not Tooltip, and I'd migrate the rest of them instead, rather than doubling down 😛

plugins/code-versions/src/styles.css
elmarburke: Exactly that's why, yes. I considering dropping this one completely, but decided it's better to have some defaults in, while maybe some rules need to be overwritten gloabaly.

yarn.lock
elmarburke: Updated!

plugins/code-versions/src/App.tsx
elmarburke: Yes, we want to close the plugin once it's restored

plugins/code-versions/src/styles.css
alecmev: Checked locally, yeah, this doesn't work. Please, borrow the approach from HubSpot, that puts all Framer styles in the base layer, so Tailwind can override them 😉

plugins/code-versions/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+    "extends": "../../tsconfig.json",
+    "include": ["src/**/*"],
+    "exclude": ["node_modules", "dist"]
+}
```
alecmev:
```suggestion
{
    "extends": "../../tsconfig.json",
    "include": ["src", "*"]
}
```

plugins/code-versions/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "vitest/config"
+
+export default defineConfig({
+    test: {
+        environment: "happy-dom",
+        setupFiles: ["./src/test-setup.ts"],
```
alecmev: With the `tsconfig.json` change, `test-setup.ts` can live in the root of the package, since it's not really plugin source code.

plugins/code-versions/package.json
```diff
@@ -0,0 +1,39 @@
...
+        "dev": "run g:dev",
+        "build": "run g:build",
+        "check-biome": "run g:check-biome",
+        "check-eslint": "run g:check-eslint",
+        "preview": "run g:preview",
+        "pack": "npx framer-plugin-tools@latest pack",
+        "check-typescript": "run g:check-typescript",
+        "test": "vitest",
+        "test:ui": "vitest --ui",
+        "test:run": "vitest run"
```
alecmev: In general, fewer scripts = less "rotten" commands and less cognitive overhead, and, Yarn allows you do stuff like `yarn test --ui`, which is equivalent to `yarn vitest --ui` and `yarn test:ui` above. Thoughts?

```suggestion
        "test": "vitest"
```

plugins/code-versions/package.json
```diff
@@ -0,0 +1,39 @@
...
+        "ts-pattern": "^5.0.5"
+    },
+    "devDependencies": {
+        "@testing-library/dom": "^10.4.0",
+        "@testing-library/jest-dom": "^6.6.3",
+        "@testing-library/react": "^16.3.0",
+        "@testing-library/user-event": "^14.6.1",
+        "@types/react": "^18.3.23",
+        "@types/react-dom": "^18.3.7",
+        "@vitest/ui": "^3.2.4",
```
alecmev, line 33: Do you prefer this over their VS Code integration?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,162 @@
...
+    const [selectedCodeFile, setSelectedCodeFile] = useState<State>({
+        type: StatusTypes.INITIAL,
+    })
```
kaloyanvi: I think calling this `selectedCodeFile` is a bit misleading because it does more than that, perhaps something like `codeFileState` and `setCodeFileState`?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,162 @@
...
+                      }
+            )
+        })
+
+        // Subscribe to selection changes in Canvas Mode
+        const unsubscribeSelection = framer.subscribeToSelection(async nodes => {
+            if (framer.mode !== "canvas") return
+
+            // If there's no selection or multiple selections, clear the selection
+            const firstNode = nodes[0]
```
kaloyanvi: we dont need to define this const until after the next if block

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,162 @@
...
+                    const componentExport = componentExports[0]
+                    if (
+                        componentExports.length !== 1 ||
+                        (componentExport && componentExport.name !== firstNode.componentName)
+                    ) {
+                        setSelectedCodeFile({
+                            type: StatusTypes.NO_SELECTION,
+                        })
+                        return
+                    }
```
kaloyanvi: Does that mean that if a code file has more than 1 export, we cannot use the plugin?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,162 @@
...
+        return () => {
+            unsubscribeOpenCodeFile()
+            unsubscribeSelection()
+            unsubscribeCodeFile()
+        }
+    }, [selectedCodeFile.type, codeFileId])
+
+    const clearSelection = useCallback(() => {
+        try {
+            framer.setSelection([])
```
kaloyanvi: nit: can we define an empty array outside of react rendering and use that array here?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
elmarburke: It's pretty much knowing "nothing has been set up yet" only, but since you're asking why, I'll remove it as there are no benefits right now.

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
elmarburke: That rule is not needed her, good catch!

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
elmarburke: I went with `selectionState`

plugins/code-versions/src/App.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <div className="flex items-center justify-center h-full p-6">
+                <div className="text-center">
+                    <h2 className="text-lg font-semibold mb-2 text-red-600">Error</h2>
+                    <p className="text-sm text-gray-600 mb-4">{fileStatus.error}</p>
+                </div>
+            </div>
+        )
+    }
+
+    return state.codeFile ? (
```
kaloyanvi, line 43: nit: imo ternaries for jsx which is this long just makes it more difficult to read, itd be clearer:

```tsx
if (fileStatus.type === StatusTypes.ERROR) // render error state

if (!state.codeFile) // render empty state

return // render code file state

```

elmarburke: this already changed while updating the designs.

plugins/code-versions/package.json
```diff
@@ -0,0 +1,39 @@
...
+        "pack": "npx framer-plugin-tools@latest pack",
+        "check-typescript": "run g:check-typescript",
+        "test": "vitest",
+        "test:ui": "vitest --ui",
+        "test:run": "vitest run"
+    },
+    "dependencies": {
+        "@radix-ui/react-tooltip": "^1.2.7",
+        "clsx": "^2.1.1",
+        "diff": "^8.0.2",
```
alecmev, line 19: So, Myers is a safe default, but, we could try to do better 😛 I haven't been able to find a JavaScript implementation of Histogram, but there's Patience here: https://www.npmjs.com/package/@lowlighter/diff (though needs a patch that exports the non-stringified version) And the repo it's borrowed from: https://github.com/jonTrent/PatienceDiff Thoughts? Not worth it?

plugins/code-versions/src/components/CodeFileView.tsx
```diff
@@ -0,0 +1,44 @@
...
+                selectedId={state.selectedVersionId}
+                onSelect={selectVersion}
+                isLoading={state.versionsLoading === LoadingState.Initial}
+            />
+            <div className="bg-bg-secondary overflow-hidden relative">
+                {state.contentLoading === LoadingState.Initial ? null : (
+                    <FileDiff original={state.versionContent ?? ""} revised={currentContent ?? ""} />
+                )}
+            </div>
+            {state.versionContent !== currentContent && canRestoreVersion ? (
```
kaloyanvi: why are we comparing with version content instead of id? 🤔

plugins/code-versions/src/components/CodeFileView.tsx
```diff
@@ -0,0 +1,44 @@
...
+                // FIXME: currently button is hidden when it's the same as the current content
+                // hide it only when it's the same as the current version
+                <button
+                    className="px-6 py-2 rounded bg-tint text-black font-semibold hover:bg-tint-dark transition disabled:opacity-50 disabled:cursor-not-allowed m-3 w-full"
+                    onClick={restoreVersion}
+                    disabled={state.restoreLoading === LoadingState.Initial}
+                >
+                    {state.restoreLoading === LoadingState.Initial ? "Restoring..." : "Restore"}
```
kaloyanvi:
```suggestion
                    {state.restoreLoading === LoadingState.Initial ? "Restoring…" : "Restore"}
```

plugins/code-versions/src/hooks/useCodeFileVersions.ts
```diff
@@ -0,0 +1,365 @@
...
+}
+
+const initialState: VersionsState = {
+    codeFile: undefined,
+    versions: [],
+    selectedVersionId: undefined,
+    versionContent: undefined,
+    versionsLoading: LoadingState.Idle,
+    contentLoading: LoadingState.Idle,
+    restoreLoading: LoadingState.Idle,
```
kaloyanvi, line 363: `restoreLoading` reads a bit weird because "restore" is a verb and if you don't know the details of the code can be confusing. Unfortunately I dont have a better idea atm either other than `restorationLoading` but not great either

kaloyanvi, line 363: Is there a case where the three loading states occur simultaneously in UI?

plugins/code-versions/package.json
elmarburke: Valid point, but it's easier to "explore" the runner with the UI. But with `yarn` not needing the `--` I have no strong preference for one or the other, and I went with this...

plugins/code-versions/package.json
elmarburke: Me personally, yes.

plugins/code-versions/src/hooks/useCodeFileVersions.ts
```diff
@@ -0,0 +1,365 @@
...
+export function useCodeFileVersions(): CodeFileVersionsState {
+    const { state: fileStatus } = useSelectedCodeFile()
+    const [state, dispatch] = useReducer(versionsReducer, initialState)
+
+    // Refs to track current operations and abort controllers
+    const versionsAbortControllerRef = useRef<AbortController | null>(null)
+    const contentAbortControllerRef = useRef<AbortController | null>(null)
+
+    // Derived state
+    const selectedVersion = state.versions.find(version => version.id === state.selectedVersionId)
```
kaloyanvi, line 16: This will be recomputed on every render right? Should we add `useMemo`?

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: I struggled with exactly that, and just `restoring` removed the keyword Loading what I wanted to keep for easier scanning the code.

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: It's on every render but I recon the operation is rather fast, and `useMemo` isn't free. Even on files with lot's of changes, this might not be a bottleneck. You think it's worth it?

plugins/code-versions/src/hooks/useCodeFileVersions.ts
```diff
@@ -0,0 +1,365 @@
...
+            versionsAbortControllerRef.current?.abort()
+            versionsAbortControllerRef.current = null
+            contentAbortControllerRef.current?.abort()
+            contentAbortControllerRef.current = null
+
+            dispatch({ type: VersionsActionType.StateResetCompleted })
+            return
+        }
+
+        dispatch({ type: VersionsActionType.CodeFileSelected, payload: { codeFile: fileStatus.codeFile } })
```
kaloyanvi, line 31: do we need to check for `state.codeFile?.id !== fileStatus.codeFile.id` before running the dispatch?

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: Actually, no, not at the same time. The first one happen together (with the UI fixes), and the last one on restore.

plugins/code-versions/src/hooks/useCodeFileVersions.ts
```diff
@@ -0,0 +1,365 @@
...
+}
+
+export const useCanRestoreVersion = () => {
+    return useIsAllowedTo("CodeFile.setFileContent")
+}
+
+export enum LoadingState {
+    Idle,
+    Initial,
+    Refreshing,
```
kaloyanvi, line 131: what do these mean? 🤔 Normally loading state is binary - true or false, why do we have 3 here?

plugins/code-versions/src/hooks/useCodeFileVersions.ts
```diff
@@ -0,0 +1,365 @@
...
+
+const initialState: VersionsState = {
+    codeFile: undefined,
+    versions: [],
+    selectedVersionId: undefined,
+    versionContent: undefined,
+    versionsLoading: LoadingState.Idle,
+    contentLoading: LoadingState.Idle,
+    restoreLoading: LoadingState.Idle,
+    restoreCompleted: false,
```
kaloyanvi, line 364: I find this state property strange, can we not derive it or use the loading states for that? 🤔

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: When you have a new version saved in the editor, we need to refresh the list (to have the new version you just saved). We don't want to show a loading indicator then. While it's not idle, it's also not doing the initial fetch.

plugins/code-versions/src/hooks/useCodeFileVersions.ts
kaloyanvi: in that case, could we simplify by having a single loading state? Something like:

`loading: "versions" | "content" | "restore | undefined`? Note that im suggesting that because Im not fully sure what the `Idle` and `Refreshing` on the loading state are for 😆

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: Great question! That check is done in the reducer to be sure nobody forgets about that, making dispatching the action safe.

plugins/code-versions/src/hooks/useCodeFileVersions.ts
kaloyanvi: As a rule of thumb I always use it on `array.find` and `array.filter` 🤷🏻‍♂️  But im ok if we dont, in that case I think we should leave a comment that its a conscious decision and the reasoning.

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: Thought about that as well, namely splitting out the LoadingState to a MutationState and having a `Done`. However, that felt more complicated/bloated than this. What do you thing?

plugins/code-versions/src/hooks/useCodeFileVersions.ts
elmarburke: With that, managing the loading states becomes cumbersome. what happens in the (theoretical) situation you have multiple loading states.

Those states are heavily inspired by React Query: `idle` when nothing is fetching and `refreshing` is Query's `isRefetching`

plugins/code-versions/src/components/FileDiff.tsx
```diff
@@ -0,0 +1,153 @@
...
+    return <LineNumberCell lineNumber={lineNumber} className="text-red-600 whitespace-nowrap" prefix={"-\u00A0"} />
+}
+
+function ContentCell({ children, className = "text-gray-600" }: { children: React.ReactNode; className?: string }) {
+    return <td className={cn("whitespace-pre px-2 py-1", className)}>{children}</td>
+}
+
+function InlineDiffs({ parts, type }: { parts: readonly InlineDiff[]; type: "add" | "remove" }) {
+    return (
+        <>
```
kaloyanvi: no need for fragment here, we can directly `return parts.map(() => {})`

plugins/code-versions/src/components/FileDiff.tsx
```diff
@@ -0,0 +1,153 @@
+import { match, P } from "ts-pattern"
```
kaloyanvi, line 1: Very nice and clean structure in this file! ❤

plugins/code-versions/package.json
elmarburke: I had a great experience with Radix before, really.

plugins/code-versions/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export type TooltipProps = {
+    content: React.ReactNode
+    children: React.ReactNode
+    delayDuration?: number
+    side?: "top" | "right" | "bottom" | "left"
+    align?: "start" | "center" | "end"
+}
+
+export function Tooltip({ content, children, delayDuration = 300, side = "top", align = "center" }: TooltipProps) {
```
kaloyanvi, line 12: I thought we were not doing tooltip for V1

plugins/code-versions/package.json
elmarburke: Maybe something to explore later.
I was also looking into other algorithms but non of them looked well used, that's why I went with a good old, boring, Myers.

plugins/code-versions/src/components/Tooltip.tsx
elmarburke: We are, to show the absolute date

plugins/code-versions/src/components/FileDiff.tsx
elmarburke: Right, this ain't a `FunctionComponent` type 😅

plugins/code-versions/src/components/VersionsSidebar.tsx
```diff
@@ -0,0 +1,147 @@
...
+    children,
+}: {
+    version: CodeFileVersion
+    isSelected: boolean
+    onSelect: (id: string) => void
+    children: React.ReactNode
+}) {
+    return (
+        <Tooltip content={formatFull(version.createdAt)} side="bottom" align="center">
+            <div
```
kaloyanvi, line 16: Since we have `onClick` this should be a button right?

elmarburke: That's changed in the upcoming UI update already, yes.

plugins/code-versions/src/components/VersionsSidebar.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+function Version({
+    version,
+    isSelected,
+    onSelect,
+    children,
+}: {
+    version: CodeFileVersion
+    isSelected: boolean
+    onSelect: (id: string) => void
```
kaloyanvi, line 10: this bit of the props is very similar across most/all components in thsi file, can we create a shared type and reuse?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -0,0 +1,156 @@
...
+        return () => {
+            unsubscribeOpenCodeFile()
+            unsubscribeSelection()
+            unsubscribeCodeFile()
+        }
+    }, [selectionState.type, codeFileId])
+
+    const clearSelection = useCallback(() => {
+        try {
+            framer.setSelection(noSelection)
```
kaloyanvi: isnt that `async`?

plugins/code-versions/package.json
alecmev: I'd err on the side of less, if you don't mind! And can document `--ui` and other tips in a README instead.

plugins/code-versions/src/utils/date.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Uses the native Intl.RelativeTimeFormat API for proper localization and pluralization
+ */
+export function formatRelative(from: Date | string, to: Date | string, locales?: Intl.LocalesArgument): string {
+    const fromDate = typeof from === "string" ? new Date(from) : from
+    const toDate = typeof to === "string" ? new Date(to) : to
+
+    const diff = fromDate.getTime() - toDate.getTime()
+
+    const seconds = Math.floor(diff / 1000)
+    if (seconds < 60) return `now`
```
kaloyanvi:
```suggestion
    if (seconds < 60) return "now"
```

plugins/code-versions/package.json
alecmev: I mean, it's not a bad library, just poorly maintained, some discussion: https://github.com/shadcn-ui/ui/issues/1622

plugins/code-versions/src/utils/diff/context-limiter.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (changeIndices.length === 0) return diffs
+
+    return markIncludedLines(diffs, changeIndices, CONTEXT_LINES)
+        .map((include, lineIndex) => {
+            if (!include) return null
+            const diff = diffs[lineIndex]
+            if (diff) return diff
+            return createDividerLine([lineIndex, lineIndex + 1])
+        })
+        .filter((line): line is LineDiff => line !== null)
```
kaloyanvi: This can be done in a single pass

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
elmarburke: I realise, this actions is unused. Removing!

plugins/code-versions/src/utils/diff/context-limiter.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function isChange(d: LineDiff): boolean {
+    return d.type === "change" || d.type === "add" || d.type === "remove"
+}
+
+function getChangeIndices(diffs: LineDiff[]): number[] {
+    return diffs.map((d, i) => (isChange(d) ? i : -1)).filter(i => i !== -1)
```
kaloyanvi: This can also be done in a single pass

plugins/code-versions/src/utils/diff/context-limiter.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function markIncludedLines(diffs: LineDiff[], changeIndices: number[], maxContextLines: number): boolean[] {
+    return diffs.map((d, i) =>
+        changeIndices.some(idx => Math.abs(i - idx) <= maxContextLines && (d.type === "context" || isChange(d)))
```
kaloyanvi: Also can be done in one loop

plugins/code-versions/src/utils/diff/context-limiter.ts
elmarburke: Yes, in this file, I opted-in to reduce the complexity of the code and functions, rather than maximal performance. This logic is already rather complex.

plugins/code-versions/src/utils/date.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Formats a date as a relative time string (e.g., "2 minutes ago", "yesterday")
+ * Uses the native Intl.RelativeTimeFormat API for proper localization and pluralization
+ */
+export function formatRelative(from: Date | string, to: Date | string, locales?: Intl.LocalesArgument): string {
+    const fromDate = typeof from === "string" ? new Date(from) : from
+    const toDate = typeof to === "string" ? new Date(to) : to
```
alecmev: Looks like this will work? I don't see string inputs.

```suggestion
export function formatRelative(from: Date, to: Date, locales?: Intl.LocalesArgument): string {
```

plugins/code-versions/src/utils/date.ts
```diff
@@ -0,0 +1,45 @@
...
+        day: "2-digit",
+    })
+    const timePart = dateObj
+        .toLocaleTimeString(locales, {
+            hour: "2-digit",
+            minute: "2-digit",
+        })
+        .toLowerCase()
+
+    return `${datePart} \u2022 ${timePart}`
```
alecmev, line 41: Thoughts on relaxing the density on this one? Since it's in a tooltip, where space is of no concern? Slack for inspiration, formats like this: `Mar 14th, 2017 at 13:14`

plugins/code-versions/src/utils/date.ts
```diff
@@ -0,0 +1,45 @@
...
+    const minutes = Math.floor(seconds / 60)
+    if (minutes < 60) return `${Math.abs(minutes)}m ago`
+
+    const hours = Math.floor(minutes / 60)
+    if (hours < 24) return `${Math.abs(hours)}h ago`
+
+    const days = Math.floor(hours / 24)
+    if (days < 7) return `${Math.abs(days)}d ago`
+
+    return toDate.toLocaleDateString(locales)
```
alecmev: Could also follow Slack here, maybe? They do `Xm ago` up to 59 minutes, and then switch to `Today at 14:11`, `Yesterday at 13:20`, `Tuesday at 14:10`, `Jun 26th at 12:00`, and finally `Mar 14th, 2017 at 13:14`. Having a long list of `1d ago` just feels less useful than `hh:mm`. Could in theory omit the time from 30+ days, since it's unlikely to matter, but, since there are already longer strings, like `Yesterday xx:yy`, then might as well keep them?

Or are you following the design strictly, so this is out of scope for now?

plugins/code-versions/src/utils/diff/context-limiter.ts
```diff
@@ -0,0 +1,55 @@
...
+                if (diff) return diff
+                return createDividerLine([lineIndex, lineIndex + 1])
+            })
+            .filter(isLineDiff)
+    )
+}
+
+function isLineDiff(diff: LineDiff | null): diff is LineDiff {
+    return diff !== null
+}
```
alecmev: Could this not be a generic `isNotNull`?

plugins/code-versions/src/utils/diff/diff-blocks.ts
```diff
@@ -0,0 +1,89 @@
...
+     * to avoid double-processing the same content.
+     */
+    skipNext: boolean
+}
+
+function splitLinesAndRemoveTrailingEmpty(value: string): string[] {
+    const lines = value.split("\n")
+    if (lines[lines.length - 1] === "") lines.pop()
+    return lines
+}
```
alecmev, line 24: You're interested specifically only in trailing new-lines? What about two trailing new-lines? And a third line with a space in it? And one leading new line, and then a space? What happens without this preprocessing?

plugins/code-versions/src/components/VersionsSidebar.tsx
```diff
@@ -0,0 +1,127 @@
...
+    return (
+        <Tooltip content={formatFull(version.createdAt)} side="bottom" align="center">
+            <div
+                className={cn(
+                    "flex items-center flex-nowrap gap-2 px-2 py-[8px] cursor-pointer select-none group relative w-full",
+                    isSelected ? "bg-gray-100 rounded-xl" : "",
+                    "transition-colors"
+                )}
+                tabIndex={0}
+                onClick={() => onSelect(version.id)}
```
kaloyanvi, line 23: if its currently selected we shouldn't do the `onClick` action here

plugins/code-versions/src/utils/diff/diff-blocks.ts
```diff
@@ -0,0 +1,89 @@
...
+            currentNewLine++
+        }
+    }
+
+    return { diffs, oldLine: currentOldLine, newLine: currentNewLine }
+}
+
+export function convertDiffBlockToLineDiffs(
+    cur: ReturnType<typeof diffLines>[number],
+    next: ReturnType<typeof diffLines>[number] | undefined,
```
alecmev: Nit: Maybe?

```suggestion
    cur: ChangeObject<string>,
    next: ChangeObject<string> | undefined,
```

plugins/code-versions/src/utils/diff/inline-diff.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export function getInlineDiff(oldLine: string, newLine: string): InlineDiff[] {
+    return diffWords(oldLine, newLine).map(part =>
+        part.added
+            ? { type: "add", value: part.value }
+            : part.removed
+              ? { type: "remove", value: part.value }
+              : { type: "unchanged", value: part.value }
+    )
+}
```
alecmev, line 15: Do we need this conversion? Is `ChangeObject<string>` too awkward to operate on directly?

plugins/code-versions/src/utils/diff/line-diff.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export function getLineDiff(oldStr: string, newStr: string): LineDiff[] {
+    const rawDiffs = diffLines(oldStr, newStr)
+    const result: LineDiff[] = []
+    let oldLine = 1
+    let newLine = 1
+
+    for (let i = 0; i < rawDiffs.length; i++) {
+        const currentDiff = rawDiffs[i]
+        if (!currentDiff) continue
```
alecmev: Should be an assert, must never happen.

plugins/code-versions/src/utils/diff/line-diff.ts
```diff
@@ -0,0 +1,143 @@
...
+            skipNext,
+        } = convertDiffBlockToLineDiffs(currentDiff, nextDiff, oldLine, newLine)
+
+        result.push(...diffs)
+        oldLine = updatedOldLine
+        newLine = updatedNewLine
+        // If skipNext is true, it means we processed both current and next diff blocks together
+        // as a paired remove/add operation, so we need to skip the next iteration to avoid
+        // processing the same block twice.
+        if (skipNext) i++
```
alecmev: Nit: I think we have a soft rule that only return/continue/break/throw can be inline, the rest in a block. Also, I'd encourage `i += 1`, slightly less error-prone 😛

plugins/code-versions/src/utils/diff/line-diff.ts
```diff
@@ -0,0 +1,143 @@
...
+ * @param newStr - The revised string to compare against
+ * @returns An array of LineDiff objects representing the differences
+ */
+export function getLineDiff(oldStr: string, newStr: string): LineDiff[] {
+    const rawDiffs = diffLines(oldStr, newStr)
+    const result: LineDiff[] = []
+    let oldLine = 1
+    let newLine = 1
+
+    for (let i = 0; i < rawDiffs.length; i++) {
```
alecmev: Nit: (https://eslint.org/docs/latest/rules/no-plusplus)

```suggestion
    for (let i = 0; i < rawDiffs.length; i += 1) {
```

plugins/code-versions/src/utils/diff/line-diff.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Diff block processing for converting raw diff output into structured LineDiff objects.
+ */
+
+function splitLinesAndRemoveTrailingEmpty(value: string): string[] {
+    const lines = value.split("\n")
+    if (lines[lines.length - 1] === "") lines.pop()
+    return lines
+}
```
alecmev, line 71: There are two of these, just in case!

plugins/code-versions/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+    "extends": "../../tsconfig.json",
+    "include": ["*"]
```
alecmev, line 3: What happened to `"src"`? 😁

alecmev (approved):
Nit galore! The core is good to go, nice stuff! 😎

plugins/code-versions/src/utils/date.ts
elmarburke: I'm following the design spec here, we still can adjust that later!

plugins/code-versions/src/utils/diff/diff-blocks.ts
elmarburke: I'm only interested in  trailing new-lines, yes.

What it does:

```
"hello\nworld\n" → ["hello", "world"]
"hello\nworld\n\n" → ["hello", "world", ""]
"hello\nworld\n " → ["hello", "world", " "]
"\nhello\nworld" → ["", "hello", "world"]
```

Why it does, what it does:
Without it, file endings would have a strange new line on the end, which is normally not a useful change to show. I followed what github is doing here, so I left out their icon when there is *no* new line.

plugins/code-versions/src/components/VersionsSidebar.tsx
elmarburke: I'll handle that in the UI update with the change to the button.

plugins/code-versions/src/utils/diff/inline-diff.ts
elmarburke: `ChangeObject` can have `part.added` and `part.removed`, and I didn't wanted to have this in our component.

plugins/code-versions/src/utils/diff/line-diff.ts
elmarburke: yeah, went with the "rule of two" here, didn't wanted to have a utils in diff itself...

plugins/code-versions/tsconfig.json
elmarburke: removed it so the test-setup can be in the root

alecmev: Wait, no, but `"*"` includes it; without `"src"` however the whole plugin is type-unchecked 😅

plugins/code-versions/src/utils/diff/line-diff.ts
alecmev: Hmm, I feel like that's a slightly too literal of interpretation of that rule, haha. Two 100% identical pieces of code surely deserve to have a single source, no refactor involved, no? But, that's a nit, okay as-is! 👌

plugins/code-versions/src/utils/diff/inline-diff.ts
alecmev: > I didn't wanted to have this in our component

That's the part I'm curious about, why? 😁 (just for info)

PR #273, opened 2025-07-03
Code Versions UI Adjustments

elmarburke:
### Description

Adjusting the UI of the Code Versions plugin.

This covers all the UI elements.

This does **not** cover loading states and interactions.

image

image

image

- re-introduces the divider line
- brings "marks" to the changed line in-line, removing irrelevant whitespace changes
- `CurrentCode` component for highlighted code when there is no change. This uses PrismJS for tokenisation.

### Testing

- [ ] Version sidebar styles
- [ ] Restore button looks right
- [ ] FileDiff, including the "edge border" between blocks,
- [ ] Current Code preview looks great in the two-colours-of-grey ⚫ ⚪ ✔️ 
- [ ] Dark Mode works
- [ ] it just feels right™ and passes the vibe test™


plugins/code-versions/src/utils/diff/inline-diff.ts
```diff
@@ -3,13 +3,83 @@ import type { InlineDiff } from "./types"
```
elmarburke, line 4: The changes here come from adding the "marks" on changes, rather than highlighting the text. That resulted in having to much highlighted before (leading spaces for example)

plugins/code-versions/src/App.tsx
```diff
@@ -24,33 +22,54 @@ export default function App() {
...
+                variant: "success",
+            })
         }
     }, [state.restoreCompleted])
 
+    useEffect(() => {
+        if (state.restoreLoading === MutationState.Mutating) {
+            framer.hideUI()
+        }
+    }, [state.restoreLoading])
```
kaloyanvi, line 35: I think this highlights the issues with how the loading states are set up. These two loading states are mutually exclusive right?

Two `useEffects` shouldn't be needed to handle a loading state which could be represented on the same const. I know this PR is mainly focusing on the UI but we should address that in follow-up later on.

plugins/code-versions/src/components/CodeFileView.tsx
```diff
@@ -24,20 +29,51 @@ export default function CodeFileView({ state, selectVersion, restoreVersion }: C
...
+    original: string
+    revised: string
+    isCurrentVersion: boolean
+}) {
+    if (original === revised) {
+        // Show the full file if there are no changes
+        return <CurrentCode code={original} />
+    }
+
+    return isCurrentVersion ? <CurrentCode code={original} /> : <FileDiff original={original} revised={revised} />
```
kaloyanvi: nit: can we extract in a const the condition for when we are showing the `<CurrentCode/>` and combine it?

elmarburke: Not a fan of that, but I merge both in the same if

plugins/code-versions/src/components/FileDiff.tsx
```diff
@@ -63,75 +64,110 @@ function ContextRow({ line }: { line: LineDiff & { type: "context" } }) {
...
+
+function RemoveRowLineNumberCell({ lineNumber, className }: { lineNumber: number | undefined; className?: string }) {
+    return (
+        <LineNumberCell
+            lineNumber={lineNumber}
+            className={cn("text-diff-remove dark:text-diff-remove whitespace-nowrap", className)}
+            prefix="-"
+        />
+    )
+}
```
kaloyanvi: nit: imo this logic would be nicer as a prop (`type: "add" | "remove"`) on `<LineNumberCell/>` and handling the custom classes and prefix inside of the component.

plugins/code-versions/src/components/CodeFileView.tsx
```diff
@@ -24,20 +29,50 @@ export default function CodeFileView({ state, selectVersion, restoreVersion }: C
...
-                {state.contentLoading === LoadingState.Initial ? null : (
-                    <FileDiff original={state.versionContent ?? ""} revised={currentContent ?? ""} />
-                )}
+            <div className="bg-code-area-light dark:bg-code-area-dark overflow-y-auto relative scrollbar-hidden">
+                <div className="absolute inset-0 ms-3 me-4 mt-3">
+                    {/* The overflow-x-auto here ensures scrollbars appear in the correct position when enabled by user styles */}
+                    <div className="overflow-x-auto scrollbar-hidden">
+                        {state.contentLoading === LoadingState.Initial ||
+                        state.versionContent === undefined ||
+                        currentContent === undefined ? null : (
```
kaloyanvi, line 38: would be nice to split this into a const wdyt?

plugins/code-versions/src/components/VersionsSidebar.tsx
```diff
@@ -92,36 +80,37 @@ export default function VersionsSidebar({
...
-                    <>
-                        {currentVersion && (
+        <aside
+            className={cn(
+                "relative flex flex-col h-full border-r border-framer-divider",
+                "after:absolute after:inset-x-0 w-versions after:bottom-0 after:h-6 after:pointer-events-none after:bg-gradient-to-t after:from-framer-bg-base after:to-transparent",
+                className
+            )}
+        >
+            {isLoading ? null : (
```
kaloyanvi, line 90: whats the difference of doing instead `{!isLoading && ()}` ?

plugins/code-versions/src/utils/diff/context-limiter.ts
```diff
@@ -12,19 +11,29 @@ export function addContextLimitingAndDividers(diffs: LineDiff[]): LineDiff[] {
...
-                if (!include) return null
-                const diff = diffs[lineIndex]
-                if (diff) return diff
-                return createDividerLine([lineIndex, lineIndex + 1])
-            })
-            .filter(isNotNull)
-    )
+    const result = markIncludedLines(diffs, changeIndices, CONTEXT_LINES)
+        // looping over the diffs multiple times is not ideal, but here, it's a trade-off between performance and readability
+        // and we're going for readability here
```
kaloyanvi:
```suggestion
        // and we're going for readability in this particular case
```

plugins/code-versions/src/utils/diff/context-limiter.ts
```diff
@@ -12,19 +11,29 @@ export function addContextLimitingAndDividers(diffs: LineDiff[]): LineDiff[] {
...
+            if (!diff) return false
+            const next = array[index + 1]
+            if (!next) return true
+            if (isDivider(diff) && isDivider(next)) return false
+            return true
+        })
+
+    // remove a divider if it's the first or last item
+    if (isDivider(result[0])) result.shift()
+    if (isDivider(result[result.length - 1])) result.pop()
```
kaloyanvi, line 38: we are already using the index in the `array.filter()`, why don't we handle this there as well?

plugins/code-versions/src/utils/diff/inline-diff.ts
```diff
@@ -3,13 +3,83 @@ import type { InlineDiff } from "./types"
...
+    const { leading, trailing, originalCore, revisedCore } = getWhitespaceAndCoreParts(originalLine, revisedLine)
+    const result: InlineDiff[] = []
+    if (leading) result.push({ type: "unchanged" as const, value: leading })
+    result.push(
+        ...diffWords(originalCore, revisedCore).map(part =>
+            part.added
+                ? { type: "add" as const, value: part.value }
+                : part.removed
+                  ? { type: "remove" as const, value: part.value }
+                  : { type: "unchanged" as const, value: part.value }
```
kaloyanvi: can we split up the nested ternaries to if statements for more readbility?

kaloyanvi (approved):
Left few minor comments, lgtm! Would be nice to get another review also on the styles by someone else.

plugins/code-versions/index.html
```diff
@@ -3,6 +3,8 @@
     <head>
         <meta charset="UTF-8" />
         <link rel="icon" type="image/svg+xml" href="/icon.svg" />
+        <link rel="stylesheet" href="https://use.typekit.net/dba5ktr.css" /> <!-- input mono -->
+        <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap" />
```
alecmev: Why not in `styles.css`?

plugins/code-versions/src/components/VersionsSidebar.tsx
elmarburke: Me, personally, I like to show explicitly "there is nothing when loading"

plugins/code-versions/src/components/CodeFileView.tsx
elmarburke: I was thinking of the same, but typescript didn't liked that idea, and introducing a type guard for that felt overkill

plugins/code-versions/src/components/VersionsSidebar.tsx
kaloyanvi: Me too but not with a ternary 😆 Normally:

```tsx
if (isLoading) return null
return <Content />
```

but here thats not possible unless we split into separate component. No strong preference, up to you :)

plugins/code-versions/src/components/CodeFileView.tsx
kaloyanvi: why is TS not happy?

plugins/code-versions/package.json
```diff
@@ -29,6 +29,7 @@
         "@testing-library/jest-dom": "^6.6.3",
         "@testing-library/react": "^16.3.0",
         "@testing-library/user-event": "^14.6.1",
+        "@types/prismjs": "^1",
```
alecmev, line 32: Nit: Exact version?

plugins/code-versions/src/components/VersionsSidebar.tsx
elmarburke: Normally a HUGE fan of early returns, but not here, yeah. I'll leave it in 🤓

plugins/code-versions/src/utils/diff/types.ts
```diff
@@ -22,3 +31,5 @@ export interface InlineDiff {
     type: "unchanged" | "add" | "remove"
     value: string
 }
+
+export const isDivider = (diff: LineDiff | undefined | null): diff is Divider => diff?.type === "divider"
```
alecmev: Nit: Doesn't really belong in "types" 😛

plugins/code-versions/src/components/CurrentCode.css
alecmev, line 1: Should this be centralized in `styles.css` maybe? And could the two not be a single class?

plugins/code-versions/src/components/CurrentCode.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }, [code])
+
+    return (
+        <div className="current-code line-numbers">
+            <pre className="font-mono text-code-size">
+                <code ref={codeRef} className="language-tsx">
+                    {code}
+                </code>
+            </pre>
+        </div>
```
alecmev, line 33: Maybe `pre` could be the top-level component, is `div` needed?

plugins/code-versions/index.html
elmarburke: no real reason, moving it back to the css!

plugins/code-versions/package.json
elmarburke: for the types, as they don't have the same version as main package, I'm not interested in an exact match, really.

plugins/code-versions/src/components/CurrentCode.css
elmarburke: I wanted to keep it seperate to be lazy loadable later

elmarburke: an the single class: the line-number plugin expect it to be "one above" the main styles to apply. didn't wanted to mix in a third class name to unify

plugins/code-versions/src/components/CurrentCode.tsx
elmarburke: The `line-numbers` plugin needs it, yes

plugins/code-versions/src/utils/diff/types.ts
elmarburke: Well, it's a type guard. If you have a better place in mind, that isn't `utils`, I'm happy to take suggestions 💅 😅

plugins/code-versions/package.json
alecmev: Exact match - no, but exact version - yes. Same as any package, no?

plugins/code-versions/src/styles.css
```diff
@@ -1,4 +1,6 @@
 @import "tailwindcss";
+@import url("https://use.typekit.net/dba5ktr.css"); /* input mono */
+@import url("https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap");
```
alecmev, line 2: AFAIK, they need to come before Tailwind to work properly.

plugins/code-versions/package.json
elmarburke: it's `1.26.5` for `@types/prismjs` but `"prismjs": "^1.30.0",`, to avoid confusion, I picked just `^1`.

plugins/code-versions/src/styles.css
elmarburke: looked okay, but still, not risking anything ✅

plugins/code-versions/src/utils/diff/types.ts
alecmev: `typeChecks.ts`? 😏 (like in Framer)

alecmev (approved):
🚢

PR #277, opened 2025-07-04
Notion: sync page cover image

madebyisaacr:
### Description

This pull request adds support for importing page cover images from Notion.

Cover Image

image

### Testing

The teams database in this page has a "Gallery" view with cover images:

- [x] Add and remove cover images from pages and re-sync to check that the changes are synced properly.
- [x] Test syncing with the Cover field enabled and disabled.

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -149,10 +149,15 @@ export async function syncCollection(
                 let coverValue: string | null = null
 
                 if (item.cover) {
-                    if (item.cover.type === "external") {
-                        coverValue = item.cover.external.url
-                    } else if (item.cover.type === "file") {
-                        coverValue = item.cover.file.url
+                    switch (item.cover.type) {
+                        case "external":
+                            coverValue = item.cover.external.url
+                            break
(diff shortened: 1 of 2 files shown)


plugins/notion/src/data.ts
```diff
@@ -144,6 +145,20 @@ export async function syncCollection(
...
+            if (fieldsById.has(pageCoverProperty.id)) {
+                let coverValue: string | null = null
+
+                if (item.cover) {
+                    if (item.cover.type === "external") {
+                        coverValue = item.cover.external.url
+                    } else if (item.cover.type === "file") {
+                        coverValue = item.cover.file.url
+                    }
+                }
```
niekert: nit: for exhaustive check

```suggestion
                if (item.cover) {
                    switch (item.cover.type) {
                        case "external":
                            coverValue = item.cover.external.url
                            break
                        case "file":
                            coverValue = item.cover.file.url
                            break
                        default:
                            item.cover satisfies never
                    }
                }
```

niekert (approved):
Awesome 👏

PR #278, opened 2025-07-04
Notion: sync phone number property

madebyisaacr:
### Description

This pull request adds support for syncing the phone number property type from Notion. Previously, phone number properties were marked as "unsupported".

Phone numbers can be imported as string or link ("tel:" link).

### Testing

The "Kitchen Sink" database has a phone number property:

- [ ] Sync a Notion database with a phone number property
- [ ] Test the Plain Text and Link types.
- [ ] Add and remove values in Notion and re-sync the collection


niekert (approved):
Great 👏

PR #279, opened 2025-07-04
Notion: improve code block language detection

madebyisaacr:
### Description

This pull request improves syncing Notion code blocks. 

- Adds a map of Notion code languages (from [API docs](https://developers.notion.com/reference/block#code)) to their equivalent Framer code languages (from code block component). 
- Changes the default language for unsupported languages from JavaScript to Markdown, because Markdown is the closest to having no formatting. Syncing non-JS code blocks as JS results in random words such as "break" and "function" being highlighted, so Markdown is a better fit to use as the fallback language.
- Uses `pre data-language=""` to set the language, which is more up-to-date than the class attribute method it was using before.

### Testing

The "First" page in the Kitchen Sink database here has some code blocks.

- [x] Add and remove code blocks in a Notion page.
- [x] Change the language of code blocks.


plugins/notion/src/blocksToHtml.ts
```diff
@@ -141,3 +143,78 @@ export function blocksToHtml(blocks: BlockObjectResponse[]) {
 
     return htmlContent
 }
+
+const CODE_LANGUAGE_MAP: Record<string, string | null> = {
```
niekert:
```suggestion
type NotionCodeLanguage = CodeBlockObjectResponse["code"]["language"]
const CODE_LANGUAGE_MAP: Record<CodeLanguage, string | null> = {
```

can we enforce the possible types from the notion API here?

PR #281, opened 2025-07-04
Notion: add color fields

madebyisaacr:
### Description

This pull request adds support for importing colors from Notion databases. Notion does not have a color property type, so instead you can use a regular text property and write colors in hex, rgb, rgba, hsl etc. format and choose to import them as colors in Framer.

image
image
<br>
image

### Testing

This Notion database has a text property named "Color"
https://www.notion.so/framer/224adf6e8c9680eca17fda0e00267005?v=224adf6e8c9680899fb5000c4fa462a1

- [ ] Change the field type of a text field to "Color"
- [ ] Test adding colors in different formats such as hex, rgb and rgba.
- [ ] Test empty color values, which import into Framer as the default color #0099FF

### DIFF
--- a/plugins/notion/src/api.ts
+++ b/plugins/notion/src/api.ts
@@ -58,7 +58,7 @@ export const supportedCMSTypeByNotionPropertyType = {
     date: ["date"],
     number: ["number"],
     title: ["string"],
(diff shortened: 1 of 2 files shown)


niekert (approved):
🔥

PR #282, opened 2025-07-04
Unsplash: allow resizing the plugin window

madebyisaacr:
### Description

This pull request adds support for resizing the Unsplash plugin window to see more images at a time and larger previews - useful for Framer users with large screens.

The default/min width and height are the same, and there's a max width of 750 which is 6 columns.

image
image

### Testing

- [ ] Resize the Unsplash plugin in canvas mode.


niekert:
When resizing the height from small to large it sometimes doesn't trigger loading the next page.

I also think the resizing is a little weird when in the `editImage/image` modes. Its very constrained. wdyt?

niekert (approved):
Nice one 👏

PR #283, opened 2025-07-04
Unsplash: add alt text + fix drag and drop

madebyisaacr:
### Description

This pull request adds two improvements to the Unsplash plugin:
* Fixes drag-and-drop. It already uses the `<Draggable>` component, but it doesn't work because it's nested inside a `<button>`. This PR moves the `<Draggable>` outside the button, making drag-and-drop work again.
* Adds alt text to the inserted images.
* Fixes a React duplicate key error in Placeholders.

### Testing

- [x] Try dragging an image in the production version of Unsplash and in this version.
- [x] Check the alt text of images after inserting.

### Notes
Alt text on drag and drop images does not work because of this bug: https://github.com/framer/plugins/issues/284
However the correct `altText` attribute has been added to Draggable in this PR, so it will work once that's fixed.

### DIFF
--- a/plugins/unsplash/src/App.tsx
+++ b/plugins/unsplash/src/App.tsx
@@ -354,8 +354,8 @@ const Placeholders = ({ index }: { index: number }) => {
     const heights = placeholderHeights[index % placeholderHeights.length]
     if (!heights) return null
 
-    return heights.map(height => (
-        <div key={height} className="animate-pulse">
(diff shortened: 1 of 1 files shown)


niekert (approved):
Amazing. Thank you! The alt text fix is tested on beta. Will be released to stable soon. (after which I'll release a new version)

PR #288, opened 2025-07-07
Code Versions: Fade in after load

elmarburke:
### Description

This introduces loading animation and also includes #287

### Testing

- [x] Switching between versions really fast doesn't mix up requests
  - [x] Click on different versions fast
  - [x] when stoped, you see the right version in the code view
- [x] Versions fade in (after switching components)
- [x] Code fades in after selecting a version
- [x] Empty State fade in

### DIFF
--- a/plugins/code-versions/src/hooks/useCodeFileVersions.ts
+++ b/plugins/code-versions/src/hooks/useCodeFileVersions.ts
@@ -46,8 +46,9 @@ export function useCodeFileVersions(): CodeFileVersionsState {
 
         // Cleanup function
         return () => {
+            abortController.abort()
+            // Only clear the ref if it still points to this controller
             if (versionsAbortControllerRef.current === abortController) {
-                abortController.abort()
                 versionsAbortControllerRef.current = null
             }
         }
@@ -71,8 +72,9 @@ export function useCodeFileVersions(): CodeFileVersionsState {
 
         // Cleanup function
         return () => {
+            abortController.abort()
+            // Only clear the ref if it still points to this controller
             if (contentAbortControllerRef.current === abortController) {
-                abortController.abort()
                 contentAbortControllerRef.current = null
             }
         }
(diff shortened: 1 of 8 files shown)


plugins/code-versions/src/hooks/useCodeFileVersions.ts
```diff
@@ -46,8 +46,9 @@ export function useCodeFileVersions(): CodeFileVersionsState {
 
         // Cleanup function
         return () => {
+            abortController.abort()
+            // Only clear the ref if it still points to this controller
             if (versionsAbortControllerRef.current === abortController) {
```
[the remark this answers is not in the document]
elmarburke: this is always the same instance, false command. Bad robot, bad AI.

plugins/code-versions/src/utils/shared-styles.ts
```diff
@@ -0,0 +1,8 @@
+import { cn } from "../utils"
+
+/**
+ * One-time fade in animation that runs only once when the element is first rendered
+ *
+ * The animation starts with opacity 0 and animates to opacity 1 over 150ms
+ */
+export const fadeInAnimationClassName = cn("animate-[fadeIn_150ms_forwards]")
```
alecmev, line 8: Hmm, is there no Tailwind-native way to do this? The `cn()` hack feels easy to forget 🙈

elmarburke: You don't need the `cn` for it to work. This is only for the tailwind plugin to pick it up properly and get autocompletion.

alecmev (approved):
Nice, good to go otherwise! 👌

plugins/code-versions/src/utils/shared-styles.ts
alecmev: Yeah, I understand why `cn` is there, that's what I meant by "for it to work". But there must be some `@variant` or something you can do? Then you don't need to import, don't need to wrap other `className` in `cn`, more idiomatic.

elmarburke: as discussed offline: at a later stage, we should collect the arbitrary values. Then, this constant becomes obsolete.

PR #292, opened 2025-07-08
Code Versions: Empty State fade in delay

elmarburke:
### Description

Adding a little delay before the EmptyState becomes visible, to be sure the iframe has resized to the plugin size already.

- 🧹 refactor the EmptyState into into it's own component, move App into components as well
- 🧰 adding a animation-delay tailwind util
- ⏲️ use the delay in the EmptyState


niekert (approved):
👏 personally feel like a little faster could feel good too

PR #297, opened 2025-07-08
Airtable: add gallery support

madebyisaacr:
### Description

This pull request adds support for importing attachment columns from Airtable as galleries.

Other:
* Removed "link" type from attachments. It doesn't work properly because Airtable file URLs expire 2 hours after fetching them from the API, so the links quit working 2 hours after syncing.
* Filter out non-image files by checking the mime types, which Airtable includes for each file. This avoids uploading txt, pdf, json etc. files as broken images. Using this list:
```js
["image/jpeg", "image/png", "image/gif", "image/apng", "image/webp", "image/svg+xml"]
```
* Fixed an error caused by missing `allowedFileTypes` in file fields when re-syncing an existing collection in manage mode.

### Testing

- [x] Add an attachment column to your Airtable base.
- [x] Add multiple image and non-image files to the items.
- [x] Test importing with image, file, and gallery types. Expected behavior:
  * Image: imports the first image file
  * File: imports the first file
  * Gallery: imports all image files, ignoring non-image files

### DIFF
--- a/plugins/airtable/src/data.ts
+++ b/plugins/airtable/src/data.ts
@@ -107,15 +107,18 @@ function getFieldDataEntryForFieldSchema(fieldSchema: PossibleField, value: unkn
                 return null
             }
 
-            const firstItem = value[0]
+            let selectedItem = value[0]
 
-            // Filter out non-image files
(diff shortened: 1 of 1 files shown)


madebyisaacr:
@tom-james-watson fixed! It now imports the first image file in the array instead of only checking the first file.

PR #309, opened 2025-07-14
Add Notion menu

tom-james-watson:
### Description

Adds a plugin menu to Notion. Moves "View in Notion" to this menu, and also adds a new "Log Out" button.

### Testing

- [ ] View in Notion
  - Log in to Notion and select a database
  - Click the menu and select "View in Notion"
  - The database should open in a new tab
- [ ] Log Out
  - Log Out while selecting a database
  - Log Out while mapping fields
  - In both scenarios, the plugin menu should be cleared and the window resized to the correct size


plugins/notion/src/auth.ts
```diff
@@ -1,7 +1,9 @@
+import { framer } from "framer-plugin"
 import { API_BASE_URL, PLUGIN_KEYS } from "./api"
+import { showLoginUI } from "./ui"
 import { generateRandomId } from "./utils"
 
-type Tokens = {
+export type Tokens = {
```
elmarburke: The exported value isn't used in this change outside of this module


```suggestion
type Tokens = {
```

plugins/notion/src/App.tsx
```diff
@@ -40,29 +42,17 @@ export function App({
...
+            if (hasAccessError) {
+                await showAccessErrorUI()
+            } else if (dataSource || isLoadingDataSource) {
+                await showFieldMappingUI()
+            } else {
+                await showLoginUI()
+            }
         }
+
+        showUI().catch(error => console.error(error))
```
elmarburke: Nit: you're using async/await here, consider sticking with one


```suggestion
        try {
          await showUI()
        } catch (error) {
          console.error(error)
        }
```

tom-james-watson: We can't because effects can't be async. But could just move the try-catch inside `showUI`

plugins/notion/src/FieldMapping.tsx
```diff
@@ -274,11 +274,6 @@ export function FieldMapping({
                 <label className="slug-field" htmlFor="slugField">
                     <div className="heading-row">
                         <span>Slug Field</span>
```
madebyisaacr:
```suggestion
                    <span>Slug Field</span>
```

Also remove `.heading-row` and `.heading-link` from App.css as they are no longer used.

plugins/notion/src/App.tsx
madebyisaacr: Would also recommend adding this:

```tsx
framer.notify(`Error opening plugin. Check the logs for more details.`, { variant: "error" })
```

plugins/notion/src/FieldMapping.tsx
madebyisaacr: `.mapping .heading-link` can also be removed.

Screenshot because GitHub won't let me add a comment:

image

tom-james-watson: Thanks!

PR #312, opened 2025-07-15
Notion: add formula support

madebyisaacr:
### Description

This pull request adds support for syncing formula fields from Notion, and improves importing null values.

### Testing

This database has one of each type of formula.
https://www.notion.so/framer/231adf6e8c96808eb382f0d0f53ec6d7?v=231adf6e8c9680269e43000c4105e5da&source=copy_link

- [ ] Select a field type for each formula. All formulas have plain text as the default field type, but you can select number, toggle, color, link, or date.
- [ ] Test changing values and field types for each formula.

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -130,9 +130,53 @@ export async function syncCollection(
                 if (fieldEntry) {
                     fieldData[field.id] = fieldEntry
                 } else {
-                    console.warn(
-                        `Skipping item at index ${index} because it doesn't have a valid value for field ${field.name}`
-                    )
+                    switch (field.type) {
+                        case "string":
+                        case "formattedText":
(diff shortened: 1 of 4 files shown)


plugins/notion/src/data.ts
```diff
@@ -402,21 +446,13 @@ export function getFieldDataEntryForProperty(
...
             return { type: "enum", value: property.select.id }
         }
         case "status": {
             if (field.type !== "enum") return null
-
-            if (!property.status) {
-                const firstCase = field.cases?.[0]?.id
-                return firstCase ? { type: "enum", value: firstCase } : null
-            }
+            if (!property.status) return null
```
madebyisaacr, line 461: The logic for selecting the first case if none is found in Notion was moved to data.ts line 141-154.

PR #313, opened 2025-07-16
Notion: add single reference support

madebyisaacr:
### Description

This pull request adds the ability to sync relations as reference fields. Before this, you can only sync relations as multi-references.

Notion has a "Limit: 1" option on relation properties to limit it to 1 selected item, essentially turning it into a single reference, but it's not exposed to the API.

### Testing

The Users and Kitchen Sink databases here have relation properties:

- [ ] Sync a relation as multi-reference.
- [ ] Sync a relation as reference.

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -349,7 +349,8 @@ export async function fieldsInfoToCollectionFields(
                 })
                 break
             }
-            case "multiCollectionReference": {
+            case "multiCollectionReference":
+            case "collectionReference": {
                 assertFieldTypeMatchesPropertyType(property.type, fieldType)
 
                 if (property.type === "relation") {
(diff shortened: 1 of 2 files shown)


plugins/notion/src/data.ts
```diff
@@ -375,7 +376,7 @@ export function fieldsInfoToCollectionFields(
                 break
             }
             default:
-                throw new Error(`Unsupported field type: ${fieldType}`)
+                throw new Error(`Unsupported field type: ${fieldType as string}`)
```
madebyisaacr, line 379: I know we prefer not to use `as`, but I could not figure out another way to make eslint not complain about this. If you remove `as string` it has this error:

```
379:60  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
```

PR #319, opened 2025-07-16
Enable `strictTypeChecked` and `stylisticTypeChecked`, fix all `any`-related issues

alecmev:
### Description

Addresses: https://github.com/framer/company/issues/33149

### Testing

- [ ] No way around it, need to smoke-test every touched plugin.


plugins/airtable/package.json
```diff
@@ -15,7 +15,8 @@
     "dependencies": {
         "framer-plugin": "^3.4.0",
         "react": "^18.3.1",
-        "react-dom": "^18.3.1"
+        "react-dom": "^18.3.1",
+        "valibot": "^1.1.0"
```
alecmev, line 19: Looked at `zod/mini`, but Valibot is just... still smaller 🤷

plugins/airtable/src/data.ts
```diff
@@ -64,6 +65,11 @@ export async function getTables(baseId: string, signal: AbortSignal): Promise<Ai
 const EMAIL_REGEX = /\S[^\s@]*@\S+\.\S+/
 const PHONE_REGEX = /^(\+?[0-9])[0-9]{7,14}$/
 
+const ArrayLikeSchema = v.pipe(
+    v.array(v.object({ type: v.optional(v.string()), id: v.string(), url: v.string() })),
+    v.minLength(1)
+)
```
alecmev: Better name welcome! Didn't want to think about it for too long.

plugins/dither/src/App.tsx
```diff
@@ -87,10 +87,7 @@ function DitherImage({ image }: { image: ImageAsset | null }) {
         if (!resolution[0] || !resolution[1]) return
 
         renderer.setSize(resolution[0], resolution[1])
-
-        // @ts-expect-error - TODO: fix this
-        program?.setResolution?.(resolution[0], resolution[1])
```
alecmev, line 92: There's no mention of `setResolution` in `ogl`. Feels vibe-coded 😄

plugins/google-search-console/src/screens/Performance.tsx
```diff
@@ -25,7 +26,7 @@ interface QueriesTableProps {
     queries: GoogleQueryResult
 }
 
-const CustomYAxisTick = ({ x, y, payload }: { x: any; y: any; payload: any }) => {
+const CustomYAxisTick = ({ x, y, payload }: { x: number; y: number; payload: { value: number } }) => {
```
alecmev, line 28: `recharts` typing leaves a lot to be desired...

plugins/google-search-console/src/utils.ts
```diff
@@ -27,11 +27,11 @@ function* chunks<T>(arr: T[], n: number): Generator<T[], void> {
 
 const batchLimit = 900
 
-export async function batchGoogleApiCall<T, P extends BatchGoogleApiCallRequest>(
+export async function batchGoogleApiCall<P extends BatchGoogleApiCallRequest>(
     token: string,
     refresh: () => Promise<GoogleToken | null>,
     parts: P[]
-): Promise<({ request: P; response: T } | null)[] | null> {
+): Promise<({ request: P; response: unknown } | null)[] | null> {
```
alecmev, line 34: `T` is a "hidden cast", better return `unknown` and do an `as T` explicitly (if you don't have a validator, which would be nice, but just infeasible for the types which aren't ours).

packages/eslint-config/src/index.ts
```diff
@@ -31,10 +31,49 @@ export default tseslint.config(
...
     {
         files: JS_AND_TS_GLOBS,
-        extends: [tseslint.configs.recommended],
+        extends: [tseslint.configs.strictTypeChecked, tseslint.configs.stylisticTypeChecked],
+        languageOptions: { parserOptions: { projectService: true } },
         rules: {
             // Covered by TypeScript
             "@typescript-eslint/no-unused-vars": 0,
+
+            // TODO: Enable
```
niekert, line 40: mans on a mission

plugins/airtable/src/data.ts
niekert: huh, its a little weird in the function where it is used it receives `unknown` as the incoming type. But at all call-sites the type is known: `AirtableFieldValue`.

Which reveals that this type should be MultipleAttachmentsSchema, probably?

niekert: Also are you sure about the min length? Imo an attachments array is valid if it is `0` in size and that kind of reads as if its not that type

plugins/airtable/src/data.ts
```diff
@@ -123,16 +128,18 @@ function getFieldDataEntryForFieldSchema(fieldSchema: PossibleField, value: unkn
             }
         }
 
-        case "collectionReference":
-            if (!Array.isArray(value)) return null
-            if (value.length === 0) return null
+        case "collectionReference": {
+            if (!v.is(v.pipe(v.array(v.string()), v.minLength(1)), value)) return null
```
niekert: supernit but I find this hard to read haha. i'd prefer the schema declared outside

plugins/cms-export/src/csv.ts
```diff
@@ -151,7 +150,7 @@ export function getDataForCSV(slugFieldName: string | null, fields: Field[], ite
                 }
 
                 default: {
-                    shouldBeNever(fieldData)
+                    fieldData satisfies never
```
niekert, line 153: nice

plugins/google-search-console/src/App.tsx
```diff
@@ -134,15 +137,13 @@ export function App() {
...
-                        <GoogleLogin
-                            loading={loading}
-                            hasError
-                            errorMessage={e.error.name !== "GoogleError" ? e.error.message || "" : ""}
-                            login={login}
-                        />
-                    )
+                FallbackComponent={({ error }: { error: unknown }) => {
+                    const errorMessage =
+                        v.is(v.object({ name: v.string(), message: v.string() }), error) && error.name !== "GoogleError"
```
niekert: nit: error object schema could be lifted up purely for readbility

plugins/google-search-console/src/utils.ts
```diff
@@ -89,26 +89,21 @@ ${JSON.stringify(part.body)}
...
-            }[]
+            const textParts = text.split("--batch_").map(part => {
+                try {
+                    const indexId = part.match(/<response-request-([0-9]+)>/)?.[1]
+                    if (!indexId) return null
+
+                    const request = currParts[Number(indexId)]
+                    if (!request) return null
+
+                    const response = JSON.parse(part.slice(part.indexOf("{"))) as unknown
```
niekert, line 100: yeaaah.. best leave that part untouched 😬

plugins/hubspot/src/pages/canvas/Chat.tsx
```diff
@@ -110,15 +100,23 @@ export default function ChatPage() {
...
+                        })
+                    }
                     disabled={!isAllowedToSetCustomCode}
                     title={isAllowedToSetCustomCode ? undefined : "Insufficient permissions"}
                     className={isAllowedToSetCustomCode ? undefined : "opacity-50"}
                 >
                     <option value="true">Enabled</option>
                     <option value="false">Disabled</option>
-                    <option value="ON_EXIT_INTENT">On Exit Intent</option>
+                    <option value='"ON_EXIT_INTENT"'>On Exit Intent</option>
```
niekert, line 120: This is not intentional right? doesn't this now include the `""` in the string?

plugins/notion/src/data.ts
```diff
@@ -227,9 +230,9 @@ export async function syncExistingCollection(
             return { didSync: false }
         }
 
-        const ignoredFieldIds: Set<string> = previousIgnoredFieldIds
-            ? new Set(JSON.parse(previousIgnoredFieldIds))
-            : new Set()
+        const ignoredFieldIds = previousIgnoredFieldIds
+            ? new Set(v.parse(IgnoredFieldIdsSchema, JSON.parse(previousIgnoredFieldIds)))
+            : new Set<string>()
```
niekert: nit: could be made DRY from `FieldMapping.tsx`, if you're touching the code anyway. but feel free to ignore.

plugins/dither/src/App.tsx
niekert: hmmm are you sure? I think this breaks Dither (works on `main`)

plugins/airtable/src/data.ts
alecmev: It replaces `if (value.length === 0) return null` in both spots below, is that not right?

Re `AirtableFieldValue`, I actually had that initially, then rebased on top of array changes, and it came with a bunch of new loose assumptions about the shape of `value`, so I decided to call a spade a spade and keep `unknown` for now 😅 Should I revisit it?

plugins/airtable/src/data.ts
alecmev: Nah, good call, was lazy, fixed this everywhere 😉

plugins/dither/src/App.tsx
alecmev: I... am not sure 🙈 Fixed!

plugins/google-search-console/src/utils.ts
alecmev: 😅

plugins/hubspot/src/pages/canvas/Chat.tsx
alecmev: The annoying part about this thing was that it's a union of `boolean` and a string literal. Rather than re-interpreting `"false"` and `"true"` I just decided to encode them with `JSON.stringify`, so that's why there are these extra quotes, which go away when you `JSON.parse`. Does this make sense?

plugins/notion/src/data.ts
alecmev: Did notice this, and went "eh, there's so much more to do...", but, with a fresh mind, why not! 😄

plugins/airtable/src/data.ts
alecmev: Renamed to `NonEmptyArrayOfAttachmentsSchema`.

niekert: I'm fine to keep it `unknown`. PR is large enough as is! 👍

plugins/hubspot/src/pages/canvas/Chat.tsx
niekert: mmhmmm it does make sense now that you explain it but does smell a bit. feels like a bit of implicit knowledge thats bound to break when context is lost. But its not a blocker for me.

plugins/airtable/src/data.ts
niekert: > It replaces if (value.length === 0) return null in both spots below, is that not right?

ah yes true. I mean it more in the sense that this is a schema for validating API structure. And if you'd use it to validate an entire API response you wouldn't want it to be invalid without a length. 

But thats theoritical and a nitpick and I should have marked as such. More as a comment because it can be an easy trap to use the same schema for full API response validation and then it'd be incorrect.

niekert (approved):
bad day to be `any`

plugins/airtable/src/data.ts
alecmev: Ah, yeah, that's why I went extra verbose with the schema names, hope that's enough of a deterrent! 🤞

plugins/hubspot/src/pages/canvas/Chat.tsx
alecmev: Added a clarifying comment!

This case is smelly no matter what, unfortunately, as it involves working with DOM's stringly-typed stuff. Could be helped a bit by encapsulating this in a reusable component, but not worth a refactor, IMO.

PR #320, opened 2025-07-16
Google Sheets: add plugin menu

madebyisaacr:
### Description

This pull request adds a plugin menu to Google Sheets.

image

Closes https://github.com/framer/plugins/issues/318


plugins/google-sheets/src/auth.ts
```diff
@@ -28,13 +31,15 @@ class Auth {
     storedTokens?: StoredTokens | null
 
     constructor() {
-        this.AUTH_URI = location.hostname.includes("localhost")
-            ? "https://localhost:8787"
-            : "https://oauth.framer.wtf/google-sheets-plugin"
+        this.AUTH_URI = "https://oauth.framer.wtf/google-sheets-plugin"
```
tom-james-watson: This localhost is for using local oauth worker. I guess you made this change because you don't have access to https://github.com/framer/FramerPluginWorkers/? Regardless, we need to undo this change. We can maybe look into tweaking the build tooling to make using the local oauth an option somehow, but that should be a separate PR.

plugins/google-sheets/src/App.tsx
```diff
@@ -86,19 +88,41 @@ export function AuthenticatedApp({ pluginContext, setContext }: AuthenticatedApp
...
+            }
+        }
+
+        showUI()
+    }, [sheetTitle, hasSheet, isSheetPending])
+
+    useEffect(() => {
+        framer.setMenu([
+            {
+                label: "View in Google Sheets",
```
tom-james-watson: Should add the sheet name here like "View MySheet in Google Sheets"

plugins/google-sheets/src/auth.ts
madebyisaacr: Yes, that was a mistake. I forgot to uncheck this change when committing the changes this time.

Making it an option, maybe as an env variable, is a good idea. I make this exact same change every time I start a new branch on one of the CMS plugins, and anyone outside of Framer will have the same problem as they don't have access to the workers repo. Will open an issue for this.

tom-james-watson: Yeah - I think an env var to use the local worker makes sense, if that works! Something like `LOCAL_OAUTH=true yarn dev`

tom-james-watson (approved):

PR #328, opened 2025-07-17
Add Airtable menu

tom-james-watson:
### Description

Adds a plugin menu to Airtable. Moves "View X in Airtable" to this menu, and also adds a new "Log Out" button.

Closes https://github.com/framer/plugins/issues/317

Note that this shouldn't be merged until 3.5.0 is stable.

### Testing

- [x] View X in Airtable
  - Log in to Airtable and select a database
  - Click the menu and select "View X in Airtable"
  - The database should open in a new tab
- [x] Log Out
  - Log Out while selecting a database
  - Log Out while mapping fields
  - In both scenarios, the plugin menu should be cleared and the window resized to the correct size


plugins/airtable/src/NoAccess.tsx
```diff
@@ -1,6 +1,7 @@
 import { framer } from "framer-plugin"
 import { useLayoutEffect, useState } from "react"
 import auth from "./auth"
+import { showAccessErrorUI as showNoTableAccessUI } from "./ui"
```
madebyisaacr: Why not use the original name? This function is only used in one place.

```suggestion
import { showAccessErrorUI } from "./ui"
```

plugins/airtable/src/App.tsx
```diff
@@ -63,6 +61,26 @@ export function App({ collection, previousBaseId, previousTableId, previousSlugF
             })
     }, [collection, previousBaseId, previousTableId])
 
+    useEffect(() => {
+        framer.setMenu([
+            {
+                label: `View ${dataSource?.tableName} in Notion`,
```
madebyisaacr:
```suggestion
                label: `View ${dataSource?.tableName} in Airtable`,
```

madebyisaacr (changes requested):
Looks good aside from these two things!

plugins/airtable/src/NoAccess.tsx
tom-james-watson: Failed rename - will fix!

plugins/airtable/src/App.tsx
tom-james-watson: 😬

plugins/airtable/src/ui.ts
```diff
@@ -0,0 +1,33 @@
...
+        height: 345,
+    })
+}
+
+export async function showLoginUI() {
+    framer.showUI({
+        width: 320,
+        height: 340,
+    })
+}
```
madebyisaacr: If it doesn't reset `resizable` to `false`, the window keeps its size when going back to the login UI but only if you've ever resized the plugin before. 

Adding `resizable: false` fixes this.

image


```suggestion
export async function showDataSourceSelectionUI() {
    framer.showUI({
        width: 320,
        height: 345,
        resizable: false,
    })
}

export async function showLoginUI() {
    framer.showUI({
        width: 320,
        height: 340,
        resizable: false,
    })
}
```

madebyisaacr (changes requested):

plugins/airtable/src/ui.ts
tom-james-watson: Nice one - thanks

madebyisaacr (approved):
LGTM

plugins/airtable/src/ui.ts
kaloyanvi, line 1: Love this ❤️

PR #329, opened 2025-07-17
Fix `no-unsafe-x` in Greenhouse

alecmev:
### Description

Addresses: https://github.com/framer/company/issues/33149

### Testing

- [x] Smoke-test the plugin.

### DIFF
--- a/plugins/greenhouse/src/api-types.ts
+++ b/plugins/greenhouse/src/api-types.ts
@@ -1,204 +1,67 @@
-interface Location {
-    name: string
-}
+import * as v from "valibot"
 
-export interface Department {
-    id: number
-    name: string
-    jobs: Job[]
-}
+const DepartmentLeafSchema = v.object({
+    id: v.number(),
+    name: v.string(),
+})
 
-export interface Office {
-    id: number
-    name: string
-    location: string | null
-    departments: Department[]
-}
+const OfficeLeafSchema = v.object({
+    id: v.number(),
+    name: v.string(),
+    location: v.nullable(v.string()),
+})
 
-export interface Job {
-    id: number
-    internal_job_id: number
-    title: string
-    updated_at: string
-    requisition_id: string | null
-    absolute_url: string
-    location: Location
-    content: string
-    departments: Department[]
-    offices: Office[]
-}
+const JobLeafSchema = v.object({
+    id: v.number(),
+    internal_job_id: v.number(),
+    title: v.string(),
+    updated_at: v.string(),
+    requisition_id: v.nullable(v.string()),
+    absolute_url: v.string(),
+    location: v.object({ name: v.string() }),
+})
 
-export interface Education {
-    id: number
-    text: string
-}
+export const DepartmentSchema = v.object({
+    ...DepartmentLeafSchema.entries,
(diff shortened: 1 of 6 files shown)


plugins/greenhouse/src/api-types.ts
```diff
@@ -1,204 +1,60 @@
...
-            throw new Error("Expected section to have numeric id")
-        }
-        if (typeof item.name !== "string") {
-            throw new Error("Expected section to have string name")
-        }
-        if (!Array.isArray(item.jobs)) {
-            throw new Error("Expected section to have jobs array")
-        }
-    }
-}
```
niekert, line 204: love to see it

plugins/greenhouse/src/dataSources.ts
```diff
@@ -160,15 +172,17 @@ const officesDataSource = createDataSource(
...
     {
         name: jobsDataSourceName,
         apiPath: "jobs?content=true",
-        fetch: async (boardToken: string) => {
+        fetch: async (boardToken: string): Promise<Job[]> => {
             const url = `https://boards-api.greenhouse.io/v1/boards/${boardToken}/jobs?content=true`
-            const items = await fetchGreenhouseData(url, "jobs")
-            validateJobs(items)
-            return items
+            const items = v.parse(JobPagesSchema, await fetchGreenhousePages(url))
```
niekert, line 177: The schema validation fails here on the API response :(

```
FieldMapping.tsx:208 ValiError: Invalid key: Expected "jobs" but received undefined
    at Module.parse (index.js:6837:11)
    at Object.fetch (dataSources.ts:184:29)
    at async getItems (data.ts:108:23)
    at async syncCollection (data.ts:234:19)
    at async handleSubmit (FieldMapping.tsx:205:13)
```

repro:

1. Open plugin
2. Use `perplexityai` as a board key
3. Select jobs for the select
4. try to sync

alecmev: Thanks, yeah, the API returns shallow non-top-level resources! Adjusted. Actually makes typing even nicer, as there are no circular references anymore.

tom-james-watson:
> Collection references don't seem to get hooked up properly, at least for perplexityai, but I don't think it's a regression, and fixing it is out of scope for this PR.

This does look like a regression because it just worked when I tested it on the production plugin

plugins/greenhouse/src/data.ts
```diff
@@ -172,13 +176,13 @@ async function getItems(
...
-                        value: value ? String(value) : null,
+                        value: v.is(StringifiableSchema, value) ? String(value) : null,
                         type: field.type,
                     }
                     break
                 case "multiCollectionReference": {
-                    const ids: string[] = []
-                    if (Array.isArray(value)) {
+                    const ids = []
+                    if (v.is(ArrayWithIdsSchema, value)) {
```
tom-james-watson, line 185: I guess it's this that isn't quite right

tom-james-watson (commented):
Otherwise looks ok

plugins/greenhouse/src/data.ts
alecmev: Good catch, the ids are numbers, not strings 👌

PR #330, opened 2025-07-17
Make local oauth worker optional for Search Console

tom-james-watson:
### Description

It's not possible to run this plugin as an external contributor otherwise. And most of the time we run this as Framer devs we aren't also making changes to the oauth worker.

Is this reasonable? We can do this for all the other plugins too if so.

### Testing

- [ ] `yarn dev` uses production oauth
- [ ] `VITE_LOCAL_OAUTH=true yarn dev` uses local oauth worker

### DIFF
--- a/plugins/google-search-console/.env.example
+++ b/plugins/google-search-console/.env.example
@@ -1,1 +1,0 @@
-VITE_OAUTH_API_DOMAIN=https://localhost:8787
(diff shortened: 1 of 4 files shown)


tom-james-watson:
Mm as these endpoints are in the codebase already, not much difference really?

Issue #323, opened 2025-07-17
Add menu to Hubspot


tom-james-watson:

madebyisaacr:
@tom-james-watson what options do you think the HubSpot menu should have aside from a "Log Out" button?

PR #332, opened 2025-07-18
Hubspot: add plugin menu

madebyisaacr:
### Description

This pull request adds a plugin menu to the HubSpot plugin with a Log Out button.

Closes https://github.com/framer/plugins/issues/323

Other changes:
* Fixed plugin having wrong height in canvas view, resulting in the bottom buttons being clipped.
Pasted Graphic embed widgets and much more

* Fixed content not being centered in sign in view.
HubSpot (Development) HubSpot (Development)

### Testing

- [x] When not signed in, there is no menu
- [ ] When signed in, there is a menu with a Log Out button
- [ ] Test in canvas and managed collection modes


madebyisaacr:
Should we make logging out keep the plugin open and go back to the log in screen, instead of closing it?

plugins/hubspot/src/App.tsx
```diff
@@ -186,6 +186,19 @@ export function App() {
                 return
             }
 
+            framer.setMenu([
+                {
+                    label: "Log Out",
+                    visible: true,
+                    onAction: () => {
+                        auth.logout()
```
kaloyanvi, line 194: shouldn't that be `async` and then `await auth.logout()`?

elmarburke: `auth.logout` is synchronious here. If that ever changes in the future, we now have eslint in place to make sure we don't forget to `await`.

kaloyanvi: yep, I realized in the notion plugin

PR #334, opened 2025-07-18
Re-enable the rest of `strictTypeChecked` and `stylisticTypeChecked`

alecmev:
### Description

WIP

### Testing

- [ ] WIP


plugins/airtable/src/Login.tsx
```diff
@@ -24,32 +24,37 @@ export function Authenticate({ onAuthenticated }: AuthenticationProps) {
         }
 
         return new Promise(resolve => {
-            pollInterval.current = setInterval(() => {
-                auth.fetchTokens(readKey).then(tokens => {
-                    clearInterval(pollInterval.current)
-                    resolve(tokens)
-                })
-            }, 2500)
+            const task = async () => {
```
niekert, line 27: just to check: requiring `await` and disallowing `.then` like this is not linted on, right, feels like a stylistic choice? that feels a little like refactoring for the sake of refactoring (imo)

niekert:
@droid review

plugins/cms-export/src/App.tsx
```diff
@@ -13,47 +13,47 @@ export function App() {
...
 
-        const csv = await convertCollectionToCSV(selectedCollection)
+        const task = async () => {
+            const csv = await convertCollectionToCSV(selectedCollection)
 
-        try {
-            if (navigator.clipboard && navigator.clipboard.writeText) {
+            try {
                 await navigator.clipboard.writeText(csv)
-            } else {
```
niekert, line 43: No longer necessary?

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
```diff
@@ -67,40 +67,44 @@ export function useSelectedCodeFile(): UseSelectedCodeFile {
...
-                    return
-                } catch (error) {
-                    const errorMessage = error instanceof Error ? error.message : "Failed to get code file"
-                    setSelectionState({
-                        type: StatusTypes.ERROR,
-                        error: errorMessage,
-                    })
-                    return
                 }
+
```
niekert, line 95: The behavior changed here now right? Because you are no longer `await`-ing the stuff inside `if(isComponentInstanceNode)` before updating state

Not sure if that leads to problems or UX flashing if the selection state is set before that work is done.

plugins/concentric/src/App.tsx
```diff
@@ -128,39 +129,31 @@ export function App() {
...
                 borderRadius: `${innerValue}px`,
             })
         }
     }, [outerValue, innerValue])
 
     function handleOuterSliderChange(value: number[]) {
         if (!state) return
-        const radii = calculateRadius(true, value[0], state.childRect)
-
-        if (radii) {
```
niekert, line 144: rip beautiful word

plugins/csv-import/src/App.tsx
```diff
@@ -63,15 +63,16 @@ function ManageConflicts({ records, onAllConflictsResolved }: ManageConflictsPro
             }
 
             let current = currentRecord
-            do {
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Intentional
+            while (true) {
```
niekert, line 67: nit: we don't allow the `do` expression anymore?

plugins/csv-import/src/csv.ts
```diff
@@ -40,7 +40,7 @@ export async function parseCSV(data: string): Promise<CSVRecord[]> {
     const { parse } = await import("csv-parse/browser/esm/sync")
 
     let records: CSVRecord[] = []
-    let error: unknown
+    let error
```
niekert, line 43:
```suggestion
    let error: Error | undefined
```


no `any` welcome here

plugins/csv-import/src/csv.ts
```diff
@@ -392,19 +390,23 @@ export async function importCSV(collection: Collection, result: ImportResult) {
...
-                      }
-                    : {
-                          id: item.id!,
-                          fieldData: item.fieldData,
-                          draft: item.draft,
-                      }
-            )
+            .map(item => {
+                if (item.action === "add") {
+                    assert(item.slug !== undefined)
```
niekert: nit (feel free to ignore)
```suggestion
                    assert(item.slug !== undefined, "Item requires a slug")
```

plugins/csv-import/src/csv.ts
```diff
@@ -392,19 +390,23 @@ export async function importCSV(collection: Collection, result: ImportResult) {
...
+                    return {
+                        slug: item.slug,
+                        fieldData: item.fieldData,
+                        draft: item.draft,
+                    }
+                }
+
+                assert(item.id !== undefined)
```
niekert: nit (feel free to ignore)

```suggestion
                assert(item.id !== undefined, "Item requires an id")
```

plugins/hubspot/src/hubdb.ts
```diff
@@ -298,17 +289,14 @@ async function processRow({ row, columns, fieldsById, slugFieldId, status, unsyn
     }
 }
 
-async function processAllRows(rows: HubDbTableRowV3[], processRowParams: Omit<ProcessRowParams, "row" | "status">) {
-    const limit = pLimit(CONCURRENCY_LIMIT)
```
niekert, line 302: This is not necessary anymore?

plugins/hubspot/src/router.tsx
```diff
@@ -56,17 +56,9 @@ function useRoutes({ routes, hubDbPluginContext, blogPluginContext }: UseRoutesP
         setIsFirstPage(false)
     }, [])
 
-    useEffect(() => {
-        const originalHistoryBack = history.back
-
-        history.back = () => {
```
niekert, line 62: wow lol, thats wild

plugins/locale-sync/src/xliff.ts
```diff
@@ -28,13 +28,10 @@ function statusToXliffState(status: LocalizedValueStatus): XliffState | undefine
         case "warning":
         case "done":
             return "final"
-        default:
-            status satisfies never
```
niekert, line 32: accidentally remove exhaustive check? and on the function below too


actually makes me wonder if thats lintable (for another PR ofc)

plugins/airtable/src/Login.tsx
alecmev: This is [no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises/). IMO, pretty important! Being able to pass `() => Promise<whatever>` to `someFunction(x: () => void)` can give a false sense of "ah, `someFunction` will wait for `x` to finish". It's common knowledge that `setInterval` will not wait for it, which we internalized by now, but there are many many situations where you aren't familiar with the interface, and would have to look at the type signature to understand if it's going to await or not. You might do that when writing code, but it's very unlikely that you'll do it when reading it. This rule forces you to never return promises from `() => void`, making it crystal clear that it won't be awaited.

plugins/cms-export/src/App.tsx
alecmev: Yup! https://caniuse.com/mdn-api_clipboard_writetext

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
alecmev: Oh, make sure to enable "Hide whitespace", there it's visible that the code is unchanged, and just the part of the callback containing `await` is now wrapped in a task.

plugins/concentric/src/App.tsx
alecmev: 🫡

plugins/csv-import/src/App.tsx
alecmev: We do! But I feel like this is clearer, as `current` is always true.

plugins/csv-import/src/csv.ts
alecmev: It's not really! 😛 Relying on inference and CFA here. If you try to do `const foo = error + 1` on the next line, you'll get:

```
- Unsafe assignment of an error typed value.
- Invalid operand for a '+' operation. Operands must each be a number or string. Got `undefined`.
- 'error' is possibly 'undefined'.
```

plugins/hubspot/src/hubdb.ts
alecmev: So, I don't see a need for it, but maybe I missed something? The code below does `await fetchTableRows` which calls the API and fetches ALL the rows in a single query, and then the limiter was being applied to local processing, which doesn't have concurrency constraints. Even the constant has this comment:

```
// Public HubSpot apps have a max of 100 requests / 10s
```

Would be a different story if the limit were applied to all interactions with the API.

plugins/hubspot/src/router.tsx
alecmev: Yeah 😅

plugins/csv-import/src/csv.ts
niekert: Oh, but now it allows you to assign anything to `error` right?

plugins/hubspot/src/hubdb.ts
niekert: ah yes you're right, they probably took it from notion where every row would do a call to the API. Still feels a little unrelated to remove this PR but all g

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
niekert: ahh mb I missed the brace after `setSelectionState`. Thought that was called before the `void task()` now

plugins/locale-sync/src/xliff.ts
alecmev: Sooo, the exhaustiveness is enforced by TypeScript, as long as there's no `undefined` in the return type. However, having looked at this again, the presence of `undefined` in the return type is an intentional defensive measure by Tom, I think. I re-added it, with an explanation.

plugins/airtable/src/Login.tsx
niekert: But just doing `void auth.fetchTokens()` would be the same right? Cause we're not returning a promise in `setInterval`. I may be misunderstanding

alecmev: Ah, yes, in this case that would be fine! Depending on the mood I just sometimes refactored non-trivial `.then.catch` chains into `await/try/catch`. My brain is programmed by years of [this rule](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md) 😄

niekert: got it! I was worried we'd have a rule like that enabled :P

plugins/csv-import/src/csv.ts
alecmev: Yes, but at the end, when you consume it, you have to deal with the whole bag of types assigned to it, it won't be `any`. This is similar to implicit vs. explicit return types in functions, but, IMO, within the bounds of a single function inference is fine, since it's within your brain context. Thoughts?

plugins/hubspot/src/hubdb.ts
alecmev: It was related because `limit` was async-poisoning a few functions 🥲 Had to fix it anyway.

plugins/airtable/src/Login.tsx
alecmev: > I was worried we'd have a rule like that enabled

(should we not? haha 😈 )

plugins/csv-import/src/csv.ts
niekert: I don't feel strongly enough, feels safer to narrow down what you can assign it to, could accidentally assign it to some random thing otherwise. but nbd!

plugins/airtable/src/Login.tsx
niekert: that's where I draw the line of ESLint becomes a hassle to guard against code style choices instead of a being a tool to help catch bugs haha

plugins/csv-import/src/csv.ts
alecmev: It's a balancing act! We already depend on a lot of inference 😛 Every `const x = ...` without an explicit type is inference, every `foo.map(x => ...)`, etc. We can decide that a "naked" let is a step too far! Should we? To me, it's comparable to a chain of a few `const` assignments depending on other inferred `const` assignments.

triozer:
ASCII fill control colors are off image

PR #344, opened 2025-07-28
PR descriptions should be required.

melissarofman:


.github/workflows/shippy.yml
```diff
@@ -0,0 +1,33 @@
+name: Shippy
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - ready_for_review
+      - synchronize
+  workflow_dispatch:
```
melissarofman, line 10: This should provide a "retry" button, but it seems like that only works if the code is in `main`. So I need to merge this PR to fully test it.

PR #354, opened 2025-08-01
HubSpot: fix dates and alt text in blog CMS

madebyisaacr:
### Description

This pull request fixes importing dates (Publish Date, Created, and Updated fields) and featured image alt text in the HubSpot plugin's blog CMS sync feature.

Previously these three dates were not imported due to a bug in date handling.

image

Featured Image already included the alt text as a separate text field, but it was not also imported as alt text on the image itself. Now alt text is imported in both places.

image

### Testing

- [ ] Test syncing blog content from HubSpot and view the Publish Date, Created, and Updated fields in Framer
- [ ] Test syncing an article with a featured image with alt text and check the Feature Image field's alt text

Slack thread: https://framer-team.slack.com/archives/C06L5H5ADK2/p1754046054072969

### DIFF
--- a/plugins/hubspot/src/blog.ts
+++ b/plugins/hubspot/src/blog.ts
@@ -88,7 +88,11 @@ export async function getBlogPluginContext(): Promise<BlogPluginContext> {
     }
 }
 
-function getFieldDataEntryInput(field: ManagedCollectionFieldInput, value: unknown): FieldDataEntryInput | undefined {
+function getFieldDataEntryInput(
+    field: ManagedCollectionFieldInput,
+    value: unknown,
+    post?: BlogPost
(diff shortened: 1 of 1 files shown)


plugins/hubspot/src/blog.ts
```diff
@@ -101,8 +105,8 @@ function getFieldDataEntryInput(field: ManagedCollectionFieldInput, value: unkno
         }
 
         case "date": {
-            if (typeof value !== "number") return undefined
-            return { type: "date", value: new Date(value).toUTCString() }
+            if (typeof value !== "string") return undefined
+            return { type: "date", value }
```
madebyisaacr, line 109: Dates in the HubSpot API are strings in ISO format so we don't need to convert them.

Issue #303, opened 2025-07-14
Convert Renamer from Svelte to React


niekert:
Death by a thousand cuts:

- It's the only non-React plugin in the repo, necessitating a bunch of dependencies.
- None of us do Svelte, which complicates maintenance/development (like #304).
- `tsc` can't handle `*.svelte`, so it needs to go through `check-svelte` instead.
-  ESLint needs a whole plugin for it to be useful on `*.svelte` files.

Issue #356, opened 2025-08-01
framer.getNodesWithType() type for VectorSetNode and VectorSetItemNode is missing


madebyisaacr:
**Describe the bug**
These types are missing in framer-plugin. It works as expected when you run `framer.getNodesWithType("VectorSetNode")` and `framer.getNodesWithType("VectorSetItemNode")`, so it's just a missing type, not an issue with the actual code.

```ts
getNodesWithType(type: "VectorSetNode"): Promise<VectorSetNode[]>;
getNodesWithType(type: "VectorSetItemNode"): Promise<VectorSetItemNode[]>;
```

It should be added in this section of `index.d.ts`

```ts
    /** Get all nodes of a certain class. */
    getNodesWithType(type: "FrameNode"): Promise<FrameNode[]>;
    getNodesWithType(type: "TextNode"): Promise<TextNode[]>;
    getNodesWithType(type: "SVGNode"): Promise<SVGNode[]>;
    getNodesWithType(type: "ComponentInstanceNode"): Promise<ComponentInstanceNode[]>;
    getNodesWithType(type: "WebPageNode"): Promise<WebPageNode[]>;
    getNodesWithType(type: "ComponentNode"): Promise<ComponentNode[]>;
```

**To Reproduce**
1. In a Framer plugin, add this code:

```ts
const nodes = await framer.getNodesWithType("VectorSetNode")
const items = await framer.getNodesWithType("VectorSetItemNode")
console.log(nodes, items)
```

3. See that there is a red underline under `"VectorSetNode"` and `"VectorSetItemNode"`
4. Open the plugin in a project with vector sets
5. Check the console to see the array of vector sets

PR #357, opened 2025-08-01
Convert Renamer to React

madebyisaacr:
### Description

This pull request converts the Renamer plugin from Svelte to React.

It's the only plugin made with Svelte, and no one on the plugins team knows Svelte, so this helps make it easier to maintain and removes Svelte as a dependency.

Closes https://github.com/framer/plugins/issues/303

**Updates**
* Converted from Svelte to React (Cursor basically one shotted it in 5 min 🤯)
* Fixed text nodes - previously text nodes did not appear in the results list
* Added support for renaming layers inside components - previously only layers in pages could be renamed
* Added success and error notifications
image

* Replica layers inside variants/breakpoints are now filtered out. Because all replicas of a layer share the same name, there is no benefit to showing each replicated layer in the list.
Before & after for a component with 3 variants:
image

**UI changes**
* Updated component and text icon colors to match layers panel
    * Vector set instances use the component icon. See https://github.com/framer/plugins/issues/358
* Reduced padding around arrow icon in layers list

### Testing

None yet

### Testing

- [x] Rename layers in search mode
- [x] Rename layers in clean mode
- [x] Switch between multiple pages and components and ensure the content in the results list changes
- [x] Rename layers inside a page
- [x] Rename layers inside a component
- [x] Make sure frames, text layers, and component instances are all renamed as expected.
- [x] To test permissions, open the plugin in a project where you are a viewer and do not have edit access. You should be able to search layers, but the rename button should be disabled.
- [x] Test the same processes in the current production version of Renamer and this PR version, and make sure they work exactly the same (or better).


plugins/renamer/src/search/indexer.ts
```diff
@@ -97,15 +104,21 @@ export class Indexer {
...
         if (this.scope === "page") {
             return [root]
         }
 
-        return await framer.getNodesWithType("WebPageNode")
+        const [webPages, componentNodes, vectorSetNodes] = await Promise.all([
+            framer.getNodesWithType("WebPageNode"),
+            framer.getNodesWithType("ComponentNode"),
+            // @ts-expect-error - See https://github.com/framer/plugins/issues/356
+            framer.getNodesWithType("VectorSetNode"),
```
madebyisaacr: The type for`framer.getNodesWithType("VectorSetNode")` is missing, but it still works as expected.

See https://github.com/framer/plugins/issues/356

plugins/renamer/src/components/SearchReplace.tsx
```diff
@@ -0,0 +1,67 @@
...
+    query: string
+    setQuery: (query: string) => void
+    replacement: string
+    setReplacement: (replacement: string) => void
+    loading: boolean
+    disableAction: boolean
+    actionLabel: string
+    showReplacement: boolean
+    isAllowed: boolean
+    onRenameClick?: () => void
```
elmarburke: This property is always set


```suggestion
    onRenameClick: () => void
```

plugins/renamer/src/App.tsx
```diff
@@ -0,0 +1,261 @@
...
+    position: "top right",
+    width: 260,
+    height: 450,
+    minWidth: 260,
+    minHeight: 450,
+    resizable: true,
+})
+
+export function App() {
+    const isAllowedToSetAttributes = useIsAllowedTo("Node.setAttributes")
```
elmarburke, line 28: Consider using `useIsAllowedTo` here.

plugins/renamer/src/components/SearchIcon.tsx
```diff
@@ -0,0 +1,10 @@
+export default function SearchIcon() {
+    return (
+        <svg xmlns="http://www.w3.org/2000/svg" width="11.384" height="11.134">
```
madebyisaacr, line 3: Search icon copied from here:

image

plugins/renamer/src/App.tsx
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    return (
+        <div className="app">
+            <Tabs items={tabItems} />
+
+            <div className="results">
+                {!textSearchFilter.query ? (
+                    <div className="empty-state">
+                        <img className="light" src={starsLightImage} alt="Stars" draggable={false} />
```
elmarburke: useCallback?

plugins/renamer/src/App.tsx
```diff
@@ -0,0 +1,261 @@
...
+    ]
+
+    const getTextAfterRename = useCallback(
+        (result: Result) => {
+            switch (currentMode) {
+                case "search":
+                    return renameResult(result, replacement)
+
+                case "clean":
+                    return cleanUpResult(result)
```
elmarburke: useMemo?

plugins/renamer/src/App.tsx
```diff
@@ -0,0 +1,261 @@
...
+                }}
+                replacement={replacement}
+                setReplacement={setReplacementValue}
+                loading={replacing}
+                disableAction={(currentMode === "search" && !replacement) || noResults}
+                isAllowed={isAllowedToSetAttributes}
+                showReplacement={currentMode === "search"}
+                actionLabel={currentMode === "search" ? "Rename" : "Clean Up"}
+                onRenameClick={renameResults}
+            />
```
elmarburke, line 271: consider moving this up and with a `useCallback` as well, as the many others above

plugins/renamer/src/App.css
elmarburke, line 1: In Svelte, the ‎`<style>` block is scoped to the current file, which helps prevent style collisions. By moving everything into a single large CSS file, we lose that benefit.

To address this, consider using CSS Modules to emulate Svelte’s scoping, or adopt BEM naming conventions to provide more context about where each component belongs.

Right now, we have multiple ‎`.empty-state` and `.results` classes, which probably conflict.

plugins/renamer/src/search/indexer.ts
```diff
@@ -97,15 +102,19 @@ export class Indexer {
         }
     }
 
-    private async getPages(): Promise<WebPageNode[]> {
+    private async getPages(): Promise<CanvasRootNode[]> {
```
elmarburke: Should we update the method name here as well?

plugins/renamer/src/search/batch_process_results.ts
```diff
@@ -58,29 +58,34 @@ export class BatchProcessResults {
     }
 
     async start(results: Result[]) {
-        if (this.started || !this.ready) return
+        if (this.started) {
```
elmarburke: Are we no longer handling ‎`!this.ready`?

plugins/renamer/src/search/batch_process_results.ts
```diff
@@ -58,29 +58,34 @@ export class BatchProcessResults {
...
+            framer.notify("A layer rename operation is already in progress", { variant: "error" })
```
elmarburke: In my opinion, the UI should never be able to reach this state. If that's the case, we don't need to add extra handling here. Also, this should be handled in the caller to keep the functionality transparent.

plugins/renamer/src/search/batch_process_results.ts
```diff
@@ -58,29 +58,34 @@ export class BatchProcessResults {
...
         let index = 0
+        let renamedCount = 0
 
         for (const batch of this.batchProcess(results)) {
             for (const result of batch) {
                 const node = await framer.getNode(result.id)
                 if (!isCanvasNode(node)) continue
 
-                await this.process(result, node, index)
+                const changed = await this.process(result, node, index)
```
elmarburke: We should keep the`renamedCount` and `changed` logic inside ‎`process` to maintain isolation here. 

This `BatchProcessResults` should maintain to be context-free, it just manages task execution and shouldn’t know about the larger workflow. Keeping responsibilities separated like this makes the code easier to reuse and reason about, and helps prevent accidental coupling as things evolve.

plugins/renamer/src/components/virtual_list.svelte
elmarburke, line 1: I didn't saw a virtual list anywhere anymore. Do large sets of results not cause issues now?

plugins/renamer/src/components/TextField.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    const focusInput = () => {
+        inputRef.current?.focus()
+    }
+
+    useEffect(() => {
+        if (focused) {
+            inputRef.current?.focus()
+        }
+    }, [focused])
```
elmarburke: duplication?

plugins/renamer/src/components/SearchReplace.tsx
```diff
@@ -0,0 +1,67 @@
...
+    onRenameClick,
+}: Props) {
+    const handleTextFieldKeyDown = (event: React.KeyboardEvent) => {
+        if (event.key === "Enter" && onRenameClick) {
+            onRenameClick()
+        }
+    }
+
+    return (
+        <div className="search-replace">
```
elmarburke: Since this is a larger refactoring already: maybe change this to a `form` and have it handle submit natively?

plugins/renamer/src/App.tsx
madebyisaacr: It already does?

elmarburke: GH maybe did GH things to me 😅  It was a `framer.isAllowedTo` with a `useState` before...

plugins/renamer/src/App.css
madebyisaacr: To solve this, I renamed the duplicate `.empty-state` and `.results` classes. App.css is only 330 lines long, which is not enough to require splitting it into multiple files.

plugins/renamer/src/search/batch_process_results.ts
madebyisaacr: Checking `this.ready` is handled by `await this.waitForReady()` a few lines down. `waitForReady` already existed, but it never did anything because if it wasn't ready, it would early return here.

Removing the `!this.ready` check makes it wait for it to be ready instead of cancelling the rename process entirely.

The reason to do this is because ready is false while it's indexing layers, so if you clicked the rename button while indexing it would fail.

```typescript
    async start(results: Result[]) {
        if (this.started) {
            framer.notify("A layer rename operation is already in progress", { variant: "error" })
            return
        }

        this.started = true
        this.onStarted()

        await this.waitForReady() 
```

plugins/renamer/src/components/TextField.tsx
madebyisaacr: Fixed by replacing the `<div>` with a `<label>`

plugins/renamer/src/search/indexer.ts
madebyisaacr: Done - renamed all references to `pages` in indexer.ts to `rootNodes`

plugins/renamer/src/search/batch_process_results.ts
madebyisaacr: Removed the notification. It can't get into this state because it shows a loading spinner while running. I originally added that before removing the `!this.ready` check which was possible to get to.

plugins/renamer/src/search/batch_process_results.ts
madebyisaacr: Moved the rename counter logic out of `BatchProcessResults` to App.tsx

plugins/renamer/src/App.tsx
```diff
@@ -129,10 +132,17 @@ export function App() {
...
 
-    const renameResults = useCallback(() => {
-        if (!isAllowedToSetAttributes) return
-        void resultsRenamer.start(results)
-    }, [isAllowedToSetAttributes, resultsRenamer, results])
+    const renameResults = useCallback(
+        (e?: React.FormEvent) => {
+            if (e) {
+                e.preventDefault()
+            }
```
elmarburke: nit:

```suggestion
              e?.preventDefault()
```

elmarburke:
image

I think we can improve how the plugin communicates insufficient edit rights. In other cases, hiding action elements makes sense, but since renaming is the core function here, that approach doesn’t fit. Maybe we could display a banner instead?

madebyisaacr:
> I think we can improve how the plugin communicates insufficient edit rights. In other cases, hiding action elements makes sense, but since renaming is the core function here, that approach doesn’t fit. Maybe we could display a banner instead?

Disabling the button and adding an "Insufficient permissions" tooltip is how all of our other plugins handle permissions. It already shows this banner if you don't have edit access, so does it really need a second banner?

image

plugins/renamer/src/components/virtual_list.svelte
madebyisaacr: Good catch, I missed that part.

In most cases it will show a lot less results because replica layers are filtered out now. For example on a page with 3 breakpoints, it will show about 1/3 as many results as before.

Yesterday I tried converting the virtual list to React with Cursor. It failed spectacularly (so many bugs), so I'm going to try again later without AI.

madebyisaacr: @elmarburke I added a virtual list. It's in `Results.tsx` now instead of a separate component.

plugins/renamer/src/components/Results.tsx
```diff
@@ -15,61 +15,110 @@ interface Props {
...
+                <div
+                    className="results-list"
+                    style={{
+                        height: results.length * ITEM_HEIGHT,
+                        paddingTop: startIndex * ITEM_HEIGHT,
+                    }}
+                >
+                    {visibleItems.map((result, index) => (
+                        <RenameComparison
+                            key={`${result.title}-${startIndex + index}`}
```
elmarburke: Consider using result's id over the constructed strucutre.

elmarburke (approved):
Looks good to me, great work 👍 

QA passed ✅

PR #360, opened 2025-08-02
Code Versions: improve padding

madebyisaacr:
### Description

This pull request moves the padding in the code viewer inside the scrollable area, allowing the code to be visible to the edges.

| Before | After |
| --- | --- |
| Code is clipped by 15px padding on top and sides | Padding is inside the scrollable area |
| Pasted Graphic 8 | Pasted Graphic 6 |
| Pasted Graphic 1 | Pasted Graphic 7 |


elmarburke:
The current horizontal padding is designed to align the button even when the line is long. That’s why the one at the top stays the same, so it behaves the same way. I think the one at the top might also work when scrolling away.

madebyisaacr:
@elmarburke ok, so do you think we should close this or do you like it? It's a pretty small change so not a big deal either way.

elmarburke (approved):
Thanks!! 👏

PR #362, opened 2025-08-04
Init 3rd-party-optimizer plugin

kurtextrem:
### Description

This pull request introduces the 3rd-party-optimizer plugin.

### Testing

- [ ] Open the plugin, click the button: It should add the snippet
- [ ] Open the plugin again, the button should show "Remove": Clicking it should remove the snippet.
- [ ] Add the snippet, disable the snippet from the settings: The plugin should tell you it's disabled and let you know how to enable it again.


.vscode/settings.json
```diff
@@ -48,5 +48,10 @@
     },
     "workbench.editorAssociations": {
         "*.svg": "default"
-    }
+    },
+    "eslint.workingDirectories": [
+        "./starters/cms",
+        { "directory": "./plugins/*", "!cwd": true },
+        { "pattern": "./packages/*", "!cwd": true }
+    ]
```
kurtextrem, line 56: **note:** this ensures eslint picks up the dictionaries correctly.

plugins/3rd-party-optimizer/src/App.tsx
```diff
@@ -0,0 +1,86 @@
...
+const subscribe = (callback: () => void) => {
+    return framer.subscribeToCustomCode(customCode => {
+        currentCustomCode = customCode
+        callback()
+    })
+}
+const getSnapshot = () => currentCustomCode
+
+export function App() {
+    const customCode = useSyncExternalStore(subscribe, getSnapshot)
```
kurtextrem, line 21: **note:** TBH I think this pattern is cleaner than `useEffect` + `setState` and probably more efficient (cuz it avoids wasted re-renders in case the callback is fast enough)

starters/cms/eslint.config.ts
```diff
@@ -16,6 +16,10 @@ export default tseslint.config(
         ],
         languageOptions: {
             globals: globals.browser,
+            parserOptions: {
+                tsconfigRootDir: __dirname,
+                projectService: true,
+            },
```
alecmev, line 22: Does this have an effect? Asking because the recommended config doesn't have any type-aware rules.

starters/cms/src/App.tsx
```diff
@@ -19,7 +19,7 @@ export function App({ collection, previousDataSourceId, previousSlugFieldId }: A
     useLayoutEffect(() => {
         const hasDataSourceSelected = Boolean(dataSource)
 
-        framer.showUI({
+        void framer.showUI({
```
alecmev, line 22: Were these reported by CMS Starter's ESLint config? These are good changes, but this plugin has its own (weaker) set of rules for now, which hasn't changed for a while 🤔

starters/cms/src/data.ts
```diff
@@ -42,7 +42,7 @@ export const dataSourceOptions = [
 export async function getDataSource(dataSourceId: string, abortSignal?: AbortSignal): Promise<DataSource> {
     // Fetch from your data source
     const dataSourceResponse = await fetch(`/data/${dataSourceId}.json`, { signal: abortSignal })
-    const dataSource = await dataSourceResponse.json()
+    const dataSource = (await dataSourceResponse.json()) as DataSource
```
alecmev, line 45: Oh, I'd leave it be for now, if there's no error. Ideally this should be Valibotted.

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
alecmev, line 1: Whoa, what's happening here?

plugins/csv-import/src/csv.ts
```diff
@@ -207,6 +207,7 @@ function getFieldDataEntryInputForField(
 
         case "divider":
         case "unsupported":
+        default:
```
alecmev, line 210: Mm, an explicit (missing) case, maybe? For clarity of what's not supported.

plugins/csv-import/src/csv.ts
```diff
@@ -445,9 +446,7 @@ export async function importCSV(collection: Collection, result: ImportResult) {
     }
 
     const finalMessage = messages.join(". ")
-    await framer.closePlugin(
-        messages.length > 1 ? finalMessage + "." : finalMessage || "Successfully imported Collection"
-    )
+    framer.closePlugin(messages.length > 1 ? finalMessage + "." : finalMessage || "Successfully imported Collection")
```
alecmev, line 449: Was this reported without updating `framer-plugin`?

plugins/flip-image/src/utils.tsx
```diff
@@ -28,7 +28,7 @@ export function bytesFromCanvas(canvas: HTMLCanvasElement): Promise<Uint8Array |
  * Utility to create a HTMLImageElement from the image data returned by `getData`
  */
 export async function loadBitmap(bytes: Uint8Array): Promise<ImageBitmap> {
-    const url = URL.createObjectURL(new Blob([bytes]))
+    const url = URL.createObjectURL(new Blob([bytes as unknown as ArrayBuffer]))
```
alecmev, line 31: Why?

.vscode/settings.json
alecmev: What do you mean by dictionaries?

plugins/3rd-party-optimizer/src/yieldGTMCalls.ts
```diff
@@ -0,0 +1,372 @@
...
+                writable: false, // Needed to prevent write, else this gets overridden when dispatching.
+                value: event[key as keyof Event],
+            })
+        }
+    }
+
+    newEvent.__originalEvent = event
+
+    if (DEBUG) {
+        // biome-ignore lint/suspicious/noConsole: debug
```
alecmev, line 136: Biome is currently only a formatter in this repo, this shouldn't be needed, no?

.vscode/settings.json
kurtextrem: directories* 😄

starters/cms/eslint.config.ts
kurtextrem: It's the recommended to turn this on by default by now unless you hit issues, so I figured why not. 
The tsconfigRootDir is needed yeah

starters/cms/src/App.tsx
kurtextrem: yeah it was reported by the pipeline

starters/cms/src/data.ts
kurtextrem: Was also reported by CI

plugins/csv-import/src/csv.ts
kurtextrem: Not sure what you mean; the error was the function doesn't always return (so return type didn't match)

plugins/csv-import/src/csv.ts
kurtextrem: No idea, it was an error right away

plugins/flip-image/src/utils.tsx
kurtextrem: TS 5.9 changed the type

plugins/3rd-party-optimizer/src/yieldGTMCalls.ts
kurtextrem: The code is from our internal repo; I didn't really do a pass on removing outdated ignore rules

plugins/code-versions/src/hooks/useSelectedCodeFile.ts
kurtextrem: All new errors from the TS & ts-eslint upgrade (and framer-plugin?). As the name is "unstable" I didn't spend the time to fix types.

starters/cms/eslint.config.ts
alecmev: > The tsconfigRootDir is needed yeah

Hmm, from [`projectService`](https://typescript-eslint.io/packages/parser/#projectservice):

> It will automatically use the nearest `tsconfig.json` for each file

And re this:

> It's the recommended to turn this on by default by now

Can you link me to that, please? I mean, surely doesn't hurt 😄 But we try to keep this specific config (in starters) as minimal/short as possible.

starters/cms/src/App.tsx
alecmev: Need to investigate, means the CI was missing this? 🤔

plugins/csv-import/src/csv.ts
alecmev: It's probably because a new field type was added, no? Does adding that uncovered type as a case not fix it?

plugins/csv-import/src/csv.ts
alecmev: Okay, yeah, `framer-plugin@3.5` satisfies `^3.3`, so all plugins now get `3.5`, even if not set as such in `package.json`. Can you maybe upgrade using `yarn upgrade-interactive`, please? So that all versions get bumped? Adding Yarn constraints is on the todo list 😅

.vscode/settings.json
alecmev: Also confusing that this is needed 🥲

plugins/3rd-party-optimizer/src/yieldGTMCalls.ts
kurtextrem: removed

PR #364, opened 2025-08-05
Airtable: ensure there is only one protocol when sync email/phone fields

triozer:
### Description

This pull request adds unit tests for the Airtable plugin's email and phone number field processing. It improves the handling of `mailto:` and `tel:` prefixes by normalizing them to ensure proper formatting regardless of input format. The implementation now correctly handles:

- Adding appropriate prefixes when missing
- Preserving single prefixes
- Normalizing multiple prefixes to a single one
- Supporting case-insensitive prefixes
- Detecting email and phone formats via regex

The PR also sets up the testing infrastructure with Vitest and Happy-DOM for the Airtable plugin.

See https://framer-team.slack.com/archives/C06L5H5ADK2/p1754387147011229

### QA

- [x] Email field processing
  - Verify `mailto:` prefix is added to emails without prefix
  - Verify single `mailto:` prefix is preserved
  - Verify multiple `mailto:` prefixes are normalized to one
  - Verify case-insensitive `mailto:` prefix handling works
  - Verify email detection by regex works for non-email field types
- [x] Phone field processing
  - Verify `tel:` prefix is added to phone numbers without prefix
  - Verify single `tel:` prefix is preserved
  - Verify multiple `tel:` prefixes are normalized to one
  - Verify case-insensitive `tel:` prefix handling works
  - Verify phone detection by regex works for non-phone field types

### DIFF
--- a/plugins/airtable/src/data.test.ts
+++ b/plugins/airtable/src/data.test.ts
@@ -7,127 +7,67 @@ describe("getFieldDataEntryForFieldSchema", () => {
         vi.clearAllMocks()
     })
 
-    const createEmailField = (): PossibleField => ({
-        id: "email_field",
-        name: "Email",
-        type: "link",
-        userEditable: true,
-        airtableType: "email",
-    })
+    const createEmailField = () =>
+        ({
+            id: "email_field",
+            name: "Email",
+            type: "link",
+            airtableType: "email",
+        }) as const satisfies PossibleField
 
-    const createPhoneField = (): PossibleField => ({
-        id: "phone_field",
-        name: "Phone",
-        type: "link",
-        userEditable: true,
-        airtableType: "phoneNumber",
-    })
+    const createPhoneField = () =>
+        ({
+            id: "phone_field",
+            name: "Phone",
+            type: "link",
+            airtableType: "phoneNumber",
+        }) as const satisfies PossibleField
 
     describe("Email field processing", () => {
-        it("should add mailto: prefix to email without prefix", () => {
-            const result = getFieldDataEntryForFieldSchema(createEmailField(), "user@example.com")
+        it.each<[PossibleField, string]>([
+            [createEmailField(), "user@example.com"],
+            [createEmailField(), "mailto:user@example.com"],
+            [createEmailField(), "mailto:mailto:user@example.com"],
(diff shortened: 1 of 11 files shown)


triozer:
* **#364** <a href="https://app.graphite.dev/github/pr/framer/plugins/364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/framer/plugins/364?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

plugins/airtable/src/data.ts
```diff
@@ -94,14 +103,14 @@ function getFieldDataEntryForFieldSchema(fieldSchema: PossibleField, value: unkn
             if (typeof value === "string") {
                 if (fieldSchema.airtableType === "email" || EMAIL_REGEX.test(value)) {
                     return {
-                        value: `mailto:${value}`,
+                        value: `${MAILTO_PREFIX}${value.replace(MAILTO_REGEX, "")}`,
```
kaloyanvi: Might be nice to create a helper for this which accepts the value and the prefix and constructs the regex in the helper on demand.

kaloyanvi: Something like this:

```ts
function prependField(value: string, prefix: string) {
	const regex = new RegExp(prefix, "gi")
	const result = value.replace(regex, "") // remove any existing prefix which might be in the value (1 or more)
	return `${prefix}${result}`
}
```

triozer: Done!

plugins/airtable/src/data.test.ts
```diff
@@ -0,0 +1,135 @@
...
+            const result = getFieldDataEntryForFieldSchema(linkField, "test@domain.co.uk")
+
+            expect(result).toEqual({
+                value: "mailto:test@domain.co.uk",
+                type: "link",
+            })
+        })
+    })
+
+    describe("Phone field processing", () => {
```
kaloyanvi, line 51: nit: instead of repeating the tests we can do a loop

triozer: Updated test too, we now use `it.each` ;)

PR #369, opened 2025-08-07
Airtable: Add manual refinement to marked HTML

triozer:
### Description

This pull request adds support for formatted text fields in the Airtable plugin. It introduces a new `refineMarkdownHTML` function that post-processes HTML generated by marked.parse() to handle edge cases and ensure consistent formatting for complex nested structures.

The implementation includes comprehensive test cases for formatted text field processing, covering various scenarios such as:
- Nested formatting in lists
- Mixed inline formatting (bold, italic, code)
- Empty elements and spacing
- Deeply nested lists with mixed formatting
- HTML tags mixed with markdown
- Special characters and escaping
- Edge cases with consecutive formatting marks
- Complex nested structures with paragraph breaks
- Links within formatted text

### Testing

- [ ] Test formatted text field processing
  - [ ] Verify nested formatting in lists works correctly
  - [ ] Check mixed inline formatting (bold, italic, code)
  - [ ] Test handling of deeply nested lists with mixed formatting
  - [ ] Verify HTML tags mixed with markdown are processed correctly
  - [ ] Test links within formatted text
- [ ] Verify edge cases
  - [ ] Test special characters and escaping
  - [ ] Check handling of consecutive formatting marks
  - [ ] Verify complex nested structures with paragraph breaks


triozer:
* **#369** <a href="https://app.graphite.dev/github/pr/framer/plugins/369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/framer/plugins/369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

plugins/airtable/src/utils.ts
```diff
@@ -17,18 +17,71 @@ export function assert(condition: boolean, message?: string): asserts condition
...
+        .split("\n\n") // Assume two newlines is a new paragraph
+        .map(paragraph => {
+            if (!/^<.*>.*<\/.*>$/.test(paragraph)) {
+                return `<p>${paragraph}</p>`
+            }
+            return paragraph
+        })
+        .join("")
+
+    return html
```
alecmev, line 68: If Marked doesn't do this and can't be configured to do this for us, then we need a new library, no? 😄 The @unifiedjs ecosystem has been extremely solid for me over the years, give [Remark](https://github.com/remarkjs/remark) a go maybe?

madebyisaacr: Marked does do all of that, but it breaks when the text contains `<p><br></p>` which was causing someone's content formatting to be broken.

So the solution was to add back the original markdown parsing code that didn't have this problem and also use Marked, but switching to a better library sounds like a better way to fix it. I'll try replacing it with Remark.

madebyisaacr: Ok, I tried using Remark and it works well. But it has the same problem where as soon as it hits `\n<p><br></p>`, it completely stops formatting anything after. Both Remark and Marked have this problem.

Maybe we should add a regex for this specific case?

madebyisaacr: This fixes it, although I hate to add a workaround for one edge case.

```tsx
let result = cellValue.replace(/\n<p><br><\/p>/g, "<br>")
```

alecmev: I'm on it! 👌

PR #370, opened 2025-08-07
Global search: Indexer

elmarburke:
### Description

This introduces the indexer. In order for the search to function, a `filterer` will follow after this. This is the good old Map/Reduce. 

- Adding the indexer for nodes on `WebPageNode`s and `ComponentNode`s. For the nodes it indexes, see the `IndexedNode` type.
- Introduces a DevTool to see what's in the index
- Sets React to 18 again, to only have one react version mounted (in test)

### Testing

- [x] Indexes canvas nodes
- [x] Indexes collection items
- [x] Indexes managed collection items (As `CollectionItem` and the fields as `TextNode` when they exist on a page)


plugins/global-search/src/components/DevToolsScene.tsx
elmarburke, line 1: This is just a DevTool that will get removed again once we have filters and the result list in place.

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    useEffect(() => {
+        indexer.start()
+
+        const onUpsert = ({ entry }: Events["upsert"]) => {
+            setIndex(prev => ({ ...prev, [entry.id]: entry }))
+        }
+
+        const onStarted = () => {
+            setIndex({})
```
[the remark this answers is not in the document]
elmarburke: That's something to look after later. I think using a immutable type here could lead to performance issues down the line anyway.

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const indexer = useMemo(() => new GlobalSearchIndexer(), [])
+    const [isIndexing, setIsIndexing] = useState(false)
+    const [index, setIndex] = useState<Record<string, IndexEntry>>({})
+
+    useEffect(() => {
+        indexer.start()
+
+        const onUpsert = ({ entry }: Events["upsert"]) => {
+            setIndex(prev => ({ ...prev, [entry.id]: entry }))
+        }
```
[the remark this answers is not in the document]
elmarburke: See https://github.com/framer/plugins/pull/370#discussion_r2260509163

plugins/global-search/src/components/DevToolsScene.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { useIndexer } from "../utils/indexer/useIndexer"
+
+export function DevToolsScene() {
+    const { index, isIndexing, indexerInstance } = useIndexer()
+
+    const [selectedEntry, setSelectedEntry] = useState<IndexEntry | null>(null)
+
+    const [filterQuery, setFilterQuery] = useState("")
+
+    const entries = useMemo(() => Object.values(index), [index])
```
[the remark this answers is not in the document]
elmarburke: That's a mutable object that is mutated.

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -0,0 +1,183 @@
...
+                    rootNodeName: collection.name,
+                    rootNodeType: "Collection",
+                    fields,
+                    slug: item.slug,
+                })
+            }
+
+            if (batch.length === this.batchSize) {
+                yield batch
+                batch = []
```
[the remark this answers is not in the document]
elmarburke: They are then included in the next iteration of data. This is fine.

plugins/global-search/src/components/DevToolsScene.tsx
```diff
@@ -0,0 +1,187 @@
...
+                    return acc
+                },
+                {} as Record<string, number>
+            ),
+        }),
+        [entries]
+    )
+
+    return (
+        <div className="flex flex-col h-full">
```
niekert, line 46: I know its just temporary / for debugging but its a lil completely broken on dark mode :D


image

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    if (isTextNode(node)) {
+        return (await node.getText()) ?? null
+    }
+
+    return null
+}
+
+export class GlobalSearchIndexer {
+    private eventEmitter = new TypedEventEmitter()
```
niekert: nitpick: `TypedEventEmitter` to me sounds like its a generic event emitter that supports types. Maybe rename `IndexerEventEmitter`?

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+export function IndexerProvider({ children }: { children: React.ReactNode }) {
+    const indexer = useMemo(() => new GlobalSearchIndexer(), [])
+    const [isIndexing, setIsIndexing] = useState(false)
+    const [index, setIndex] = useState<Record<string, IndexEntry>>({})
+
+    useEffect(() => {
+        indexer.start()
+
+        const onUpsert = ({ entry }: Events["upsert"]) => {
+            setIndex(prev => ({ ...prev, [entry.id]: entry }))
```
niekert: This seems like a non-urgent update that may be expensive, I wonder if it would be nice to wrap that in `startTransition`

niekert: Also just curious if there's any reason to prefer syncing the entries from the Indexer with React in a `useEffect` over `useSyncExternalStore` (atp I'm not sure anymore what is the recommended approach, just curious)

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -0,0 +1,183 @@
...
+            this.eventEmitter.emit("started", undefined)
+
+            for await (const batch of this.crawlNodes([...pages, ...components])) {
+                if (this.abortRequested) break
+                this.upsertEntries(batch)
+            }
+
+            const collections = await framer.getCollections()
+
+            for await (const batch of this.crawlCollections(collections)) {
```
niekert, line 166: I think out of scope for this PR, but I wonder if you are considering something to make this not freeze the UI on huge projects? 

Idk maybe like something with `requestIdleCallback` or some kind of scheduling? A worker probs will be more overhead because everything has to go through `framer` anyway

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -0,0 +1,183 @@
...
+    private async *crawlCollections(collections: readonly Collection[]): AsyncGenerator<IndexEntry[]> {
+        let batch: IndexEntry[] = []
+
+        for (const collection of collections) {
+            const [fieldNames, items] = await Promise.all([collection.getFields(), collection.getItems()])
+            const fieldNameMap = new Map(fieldNames.map(f => [f.id, f.name]))
+
+            for (const item of items) {
+                const fields = Object.fromEntries(
+                    Object.entries(item.fieldData).flatMap(([key, field]) => {
```
niekert, line 114: nit: I find this quite hard to read with the nesting. I feel like a regular `for ... of` loop would be more readable and consistent with code style in FramerStudio (but am nitpicking so feel free to ignore)

plugins/global-search/src/utils/indexer/strip-markup.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    })
+
+    it("should handle complex HTML structures (fallback to DOMParser)", () => {
+        expect(stripMarkup("<div><p>Only road <span>I've ever <em>been down</em></span></p></div>")).toBe(
+            "Only road I've ever been down"
+        )
+    })
+
+    it("should handle self-closing tags", () => {
+        expect(stripMarkup("Try to make<br/>ends meet")).toBe("Try to makeends meet")
```
niekert: Wouldn't thisbreak a "Match word" filter option like in VSCode? (not sure if planned)

plugins/global-search/src/utils/indexer/types.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export function shouldIndexNode(value: AnyNode): boolean {
+    if (isFrameNode(value)) return true
+    if (isComponentInstanceNode(value)) return true
+    if (isTextNode(value)) return true
+    if (isSVGNode(value)) return true
+    if (isWebPageNode(value)) return true
+    if (isComponentNode(value)) return true
+
+    return false
```
niekert, line 58: nit: kinda feel like it would be nice to somehow make this exhaustive so we get notified when we add support for new node types while upgrading the plugin version

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
elmarburke: `startTransition`: Yes! That's a great idea, didn't thought about using it yet, as I don't show the data 😅 

`useSyncExternalStore `: Thought about it, that's why I added the event emitter already. But I'm not entirely sure how the filterer will work and what API it will have, and how the two might be connected.

plugins/global-search/src/utils/indexer/indexer.ts
elmarburke: I thought about a worker at first, but yeah, sending data back and forth might be to much. The nodes should be chunked already, giving time in between for other tasks. For Collections however, the process loop might be to long, but that could be batched as well...

But, since plugins run in iframes, we don't share an event loop with `framer.com`, giving a bit more head room...

plugins/global-search/src/utils/indexer/strip-markup.test.ts
elmarburke: Good point! This should probably add a `" "` there instead...

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
elmarburke: `useSyncExternalStore` might be good for a next iteration. It requires to maintain a immutable state and I don't want to bloat this PR to much for this.

plugins/global-search/src/components/DevToolsScene.tsx
elmarburke: I'm not investing time into this for now, as it stays only for a matter of days...

plugins/global-search/src/utils/indexer/types.ts
elmarburke: will do that and make the types as narrow as they need to be once the results are visible and it's easier to rule out additional, no needed data as well

plugins/global-search/src/utils/indexer/indexer.ts
elmarburke: 🙉 🙈 🙊

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+import type { IndexerEvents } from "./indexer"
+import { GlobalSearchIndexer } from "./indexer"
+import type { IndexEntry } from "./types"
+
+/**
+ * Creates an indexer instance and provides it to the children.
+ * Manages the index in a React state, so that it can use reactivity to show the results
+ */
+export function IndexerProvider({ children }: { children: React.ReactNode }) {
+    const indexer = useMemo(() => new GlobalSearchIndexer(), [])
```
niekert: should this be `useState()` too?

> However, since useMemo is performance optimization, not a semantic guarantee, React may throw away the cached value if [there is a specific reason to do that](https://react.dev/reference/react/useMemo#caveats).

niekert (approved):
Nice one 👏

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
elmarburke: It can even be a `useRef` here, there is no tracking needed.

Issue #366, opened 2025-08-05
Airtable base selection page has no error handling


madebyisaacr:
**Describe the bug**
In the Airtable plugin, if there's an API error on the base selection UI, it says "Loading..." forever.

**To Reproduce**
The problem is that my Airtable base got corrupted (I can't access it from Airtable's dashboard either), so I can't exactly give steps to reproduce. However you can simulate an API error by adding `throw new Error()` right before the API call in `fetchTables` in `api.ts`.

**Expected behavior**
It should have an error state and show a notification when there's an API error.

**Screenshots**
<img width="525" height="173" alt="Image" src="https://github.com/user-attachments/assets/0bd8e93e-2fc1-46d7-a51c-9830a1bc6d08" />

<img width="329" height="411" alt="Image" src="https://github.com/user-attachments/assets/8c22245b-c1ec-4334-9094-873798da023b" />

PR #372, opened 2025-08-09
Airtable: add error handling to base selection page

madebyisaacr:
### Description

This pull request adds error handling for API calls to the base selection page.

Closes https://github.com/framer/plugins/issues/366

The Airtable base I was using for testing somehow became corrupted, and it won't open in Airtable. API calls to get the base info also fail, which is how I found out that if there's an API error, it says "Loading..." forever and doesn't tell you there was an error.

**Changes:**
* Show a notification and show "Error" if the tables or bases API call fails.
* Reduced the spacing between the select rows from 15px to 10px.

Left: fetching list of all bases fails
Right: fetching info for the selected base fails
image

image

### Testing

To simulate an API error, add `throw new Error()` to the top of the `getUserBases` and `getTables` functions (not at the same time). This has the same effect as it failing to fetch from the API.

- [x] Test bases failing to load
- [x] Test tables failing to load

### DIFF
--- a/plugins/airtable/src/SelectDataSource.tsx
+++ b/plugins/airtable/src/SelectDataSource.tsx
@@ -10,7 +10,9 @@ interface SelectDataSourceProps {
 }
 
 export function SelectDataSource({ collection, onSelectDataSource }: SelectDataSourceProps) {
-    const [status, setStatus] = useState<"loading-bases" | "loading-tables" | "ready" | "error">("loading-bases")
+    const [status, setStatus] = useState<"loading-bases" | "loading-tables" | "ready" | "error-bases" | "error-tables">(
+        "loading-bases"
+    )
     const [bases, setBases] = useState<AirtableBase[]>([])
     const [tables, setTables] = useState<AirtableTable[]>([])
 
@@ -22,9 +24,15 @@ export function SelectDataSource({ collection, onSelectDataSource }: SelectDataS
         setStatus("loading-bases")
 
         const task = async () => {
-            const bases = await getUserBases()
-            setBases(bases)
-            setSelectedBaseId(bases[0]?.id ?? "")
+            try {
+                const bases = await getUserBases()
+                setBases(bases)
+                setSelectedBaseId(bases[0]?.id ?? "")
+            } catch (error) {
+                console.error(error)
+                setStatus("error-bases")
+                framer.notify("Failed to load bases. Check the logs for more details.", { variant: "error" })
+            }
         }
 
         void task()
@@ -39,12 +47,22 @@ export function SelectDataSource({ collection, onSelectDataSource }: SelectDataS
             setSelectedTableId("")
 
             const task = async () => {
-                const tables = await getTables(selectedBaseId, abortController.signal)
-                if (abortController.signal.aborted) return
+                try {
+                    const tables = await getTables(selectedBaseId, abortController.signal)
+                    if (abortController.signal.aborted) return
 
-                setTables(tables)
-                setStatus("ready")
-                setSelectedTableId(tables[0]?.id ?? "")
+                    setTables(tables)
+                    setStatus("ready")
+                    setSelectedTableId(tables[0]?.id ?? "")
+                } catch (error) {
+                    console.error(error)
+                    setStatus("error-tables")
+
+                    const baseName = bases.find(base => base.id === selectedBaseId)?.name ?? selectedBaseId
+                    framer.notify(`Failed to load tables for base "${baseName}". Check the logs for more details.`, {
+                        variant: "error",
+                    })
+                }
             }
 
             void task()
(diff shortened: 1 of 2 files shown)


plugins/airtable/src/SelectDataSource.tsx
```diff
@@ -96,47 +116,58 @@ export function SelectDataSource({ collection, onSelectDataSource }: SelectDataS
...
+                            status === "error-tables" ||
+                            status === "error-bases"
+                        }
+                    >
+                        <option value="" disabled>
+                            {status === "loading-tables"
+                                ? "Loading…"
+                                : status === "error-tables" || status === "error-bases"
+                                  ? "Error"
+                                  : "Choose…"}
```
elmarburke: Try to avoid nested ternary operators because they can be difficult to follow and understand. Instead, consider wrapping decision tree into a helper function that uses a switch/case statement.

plugins/airtable/src/SelectDataSource.tsx
```diff
@@ -96,47 +116,58 @@ export function SelectDataSource({ collection, onSelectDataSource }: SelectDataS
...
+                    <select
+                        id="base"
+                        onChange={event => {
+                            setSelectedBaseId(event.target.value)
+                        }}
+                        value={selectedBaseId}
+                        disabled={status === "loading-bases" || status === "error-bases"}
+                    >
+                        <option value="" disabled>
+                            {status === "loading-bases" ? "Loading…" : status === "error-bases" ? "Error" : "Choose…"}
```
elmarburke: As in the previous case, try to avoid nested ternary expressions for readability.

plugins/airtable/src/SelectDataSource.tsx
```diff
@@ -39,12 +49,22 @@ export function SelectDataSource({ collection, onSelectDataSource }: SelectDataS
...
+
+                    setTables(tables)
+                    setStatus("ready")
+                    setSelectedTableId(tables[0]?.id ?? "")
+                } catch (error) {
+                    console.error(error)
+                    setStatus("error-tables")
+
+                    const baseName = selectedBase?.name ?? selectedBaseId
+                    framer.notify(`Failed to load tables for base "${baseName}". Check the logs for more details.`, {
```
elmarburke, line 80: I’m not sure about the user experience here. We shouldn’t ask users to check their browser log for details. 

Instead, we could cover some common cases by inspecting the error message (such as a network issue) and for all other cases, display the error message prominently in the plugin window. Since such an error prevents the plugin from working, taking over the entire plugin window with the details could work.

This is how it looks like in Framer:

Image

plugins/airtable/src/SelectDataSource.tsx
```diff
@@ -10,21 +10,31 @@ interface SelectDataSourceProps {
 }
 
 export function SelectDataSource({ collection, onSelectDataSource }: SelectDataSourceProps) {
-    const [status, setStatus] = useState<"loading-bases" | "loading-tables" | "ready" | "error">("loading-bases")
+    const [status, setStatus] = useState<"loading-bases" | "loading-tables" | "ready" | "error-bases" | "error-tables">(
```
elmarburke, line 22: While it’s okay to have error states, since the Plugin won’t be functional in those cases, consider throwing in error cases and letting it run into an error boundary. That way, the happy path components will remain happy flow only. However, this approach would be different if some functions remain usable in error cases.

madebyisaacr: In the case where loading a specific base fails, we shouldn't throw because you can still select a different base. But if loading the list of all bases fails, it makes sense to have an error boundary for that.

plugins/airtable/src/SelectDataSource.tsx
madebyisaacr: 100% agree!

Do you know if the [Alerts API](https://github.com/framer/company/issues/33603) is planned? If so, we should wait for that to be released instead of building an error UX now and then having to replace it later.

elmarburke: For now, the idea is captured but implementation is not planned yet.

plugins/airtable/src/SelectDataSource.tsx
elmarburke: That makes sense 👍

plugins/airtable/src/SelectDataSource.tsx
madebyisaacr: Ok, I think this is out of scope for this PR. Because if we're going to make an error message UI, we should design something that we can reuse and implement it for all errors in all of the CMS plugins instead of making something just for this specific case.

elmarburke: I think error handling can be in a simple error component with a retry/recover functionality.

madebyisaacr: Added a simple error state.

image

plugins/airtable/src/SelectDataSource.tsx
```diff
@@ -110,6 +117,57 @@ export function SelectDataSource({ collection, onSelectDataSource }: SelectDataS
...
+                tablesText = "Loading…"
+                break
+            case "error-bases":
+            case "error-tables":
+                tablesText = "Error"
+                break
+        }
+
+        return [basesText, tablesText]
+    }, [status])
```
elmarburke: nit:
```suggestion
// this defined outside of the render flow
const statusMap = {
  "loading-bases": { bases: "Loading…", tables: "Choose…" },
  "error-bases":   { bases: "Error", tables: "Error" },
  "loading-tables":{ bases: "Choose…", tables: "Loading…" },
  "error-tables":  { bases: "Choose…", tables: "Error" },
}

// this inside
const { bases, tables } = statusMap[status] ?? { bases: "Choose…", tables: "Choose…" }
```

plugins/airtable/src/SelectDataSource.tsx
```diff
@@ -10,41 +11,67 @@ interface SelectDataSourceProps {
...
     const [bases, setBases] = useState<AirtableBase[]>([])
     const [tables, setTables] = useState<AirtableTable[]>([])
 
     const [selectedBaseId, setSelectedBaseId] = useState<string>("")
     const [selectedTableId, setSelectedTableId] = useState<string>("")
     const [isLoading, setIsLoading] = useState(false)
 
-    useEffect(() => {
+    const basesLoadedRef = useRef(false)
+    const lastBaseIdRef = useRef<string>("")
```
elmarburke, line 33: I'm unsure about the refs - couldn't you depend on the state from above? That would make the decision tree easier.
Also, now if loading fails, you can't recover at all, but you have the `error-bases` status. That's maybe an other indicator this should be a single thing

madebyisaacr: By "this should be a single thing" do you mean there should be one error state for both types of errors? That's not really possible, because if loading a single base errors, you should be able to select a different base. But if loading all bases fails, there needs to be a different error state.

PR #373, opened 2025-08-12
Global search: adding filterer

elmarburke:
### Description

This PR will introduce the filter for indexed results, in addition to some rudimentary UI to show the results.
The results are grouped so they can be shown in context later on.

Note that the amber UI elements are not matching design specs. This PR is about the filter functionality, and the introduced elements are just enabling easy access to the functionality.

This filterer is not chunking the work nor is it caching results. If they seem to be needed and beneficial, this functionality comes in a follow-up.

image

image

### Testing

- [x] Filter by text works
  - Enter text in the search box, and you see matching results
- [x] Match ranges are correctly
  - Matching text is highlighted in the result
- [x] Root Node filters work
  - Hit the `...` and un-select some; those root nodes are no longer in the result list.


plugins/global-search/src/utils/filter/execute-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+                field,
+                text,
+                ranges,
+                entry,
+                type: ResultType.CollectionItem,
+            }
+        }
+    }
+
+    return false
```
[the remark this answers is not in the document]
elmarburke: As stated above, this should be handled by the indexer, creating multiple items for the entry.

plugins/global-search/src/utils/filter/ranges.ts
```diff
@@ -0,0 +1,32 @@
...
+    const needle = isCaseSensitive ? query : query.toLowerCase()
+
+    const indices = indicesOf(haystack, needle)
+    if (indices.length === 0) return []
+
+    return indices.map((start): Range => [start, start + needle.length])
+}
+
+export function randomID() {
+    return `${Date.now()}-${Math.floor(Math.random() * 100)}`
```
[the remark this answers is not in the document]
elmarburke: removed the function

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -0,0 +1,203 @@
...
+        startTransition(() => {
+            setQuery(event.target.value)
+        })
+    }, [])
+
+    const hasResults = useMemo(() => {
+        for (const [, resultForRootNodeType] of entries(results)) {
+            if (!resultForRootNodeType) continue
+
+            for (const [, resultsForRootId] of entries(resultForRootNodeType)) {
```
niekert: kinda nitpicky but destructuring every entry just to get the value in an object that can be huge feels like unneccesary overhead imo.

I think in general `Object.entries()` is quite a lot slower than `Object.keys()`, and its also used in the rendering

plugins/global-search/src/utils/filter/execute-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    for (const entry of index) {
+        let include = true
+        let result: Result | undefined
+
+        for (const filter of filters) {
+            const filterResult = executeFilter(filter, entry)
+
+            if (typeof filterResult !== "boolean") {
+                result = filterResult
```
niekert: I think I find it a little confusing that filters can either return `boolean` or a `FilterResult`. 

What if a third filter is added whose implementation returns a `FilterResult` instead of a boolean? it would overwrite the result, I think?

Could it be reflected in the types that there can only be one filter that produces a result when there's other filters to filter things out / if items should be included?

plugins/global-search/src/utils/filter/execute-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+        id: entry.id,
+        text: text,
+        ranges,
+        entry,
+        type: ResultType.Node,
+    }
+}
+
+function executeTextFilterForCollectionItems(filter: TextFilter, entry: IndexCollectionItemEntry): FilterResult {
+    // FIXME: This only returns the first matching field
```
niekert, line 70: Just to confirm this FIXME will be done in a followup? Up to you but it feels related to filtering logic so just making sure you didn't miss it

plugins/global-search/src/utils/filter/ranges.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    while (index !== -1) {
+        indices.push(index)
+        index = haystack.indexOf(needle, index + needle.length)
+    }
+
+    return indices
+}
+
+export function findRanges(text: string, query: string, isCaseSensitive: boolean): readonly Range[] {
```
niekert, line 18: nit: nice one to have some self documenting unit tests for

plugins/global-search/src/utils/filter/types.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type Filter = TextFilter | RootNodesFilter
+
+export function isTextFilter(filter: Filter | undefined): filter is TextFilter {
+    return typeof filter !== "undefined" && filter.type === FilterType.Text
+}
+
+export function isRootNodesFilter(filter: Filter | undefined): filter is RootNodesFilter {
+    return typeof filter !== "undefined" && filter.type === FilterType.RootNodes
+}
```
niekert: these are unused I think?

plugins/global-search/src/utils/filter/types.ts
```diff
@@ -0,0 +1,71 @@
...
+    return result.type === type
+}
+
+export function resultContainsCollectionItem(result: Result): result is CollectionItemResult {
+    return isResultType(result, ResultType.CollectionItem)
+}
+
+export function resultContainsNode(result: Result): result is NodeResult {
+    return isResultType(result, ResultType.Node)
+}
```
niekert: also unused

niekert (commented):
Nice one 👏 

> This filterer is not chunking the work nor is it caching results. If they seem to be needed and beneficial, this functionality comes in a follow-up.

Just FYI on a large project the Plugin freezes when you type a single character, so I think it would need something like that soon 🙏

plugins/global-search/src/utils/filter/execute-filter.ts
elmarburke: This comes separately, wanted this PR to remain atomic(ish). Same for the async filter...

plugins/global-search/src/components/SearchScene.tsx
elmarburke: I think you mean `Object.values` here, don't you? Will use that instead. Good point.

plugins/global-search/src/utils/filter/execute-filter.ts
elmarburke: I was first thinking on having two different types here, one as a `ResultQuery` outputting a `Result` and then a `PredicateFilter`, where it's only a `boolean`. But, since we only have these two filter types, and I have a hard time to think on any other, I kept it like this.
Do you think the code would benefit from such a split?

niekert: I did get a little confused about the potential for the result being overwritten here while trying to understand the code. So to me separating a PredicateFilter and ResultQuery kinda makes more sense, at makes it more difficult to make mistake should somebody introduce another filter in the future (imo)

plugins/global-search/src/utils/filter/types.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type Result = CollectionItemResult | NodeResult
+
+function isResultType<T extends ResultType>(result: Result, type: T): result is Result & { type: T } {
+    return result.type === type
+}
+
+export function resultContainsCollectionItem(result: Result): result is CollectionItemResult {
+    return isResultType(result, ResultType.CollectionItem)
+}
```
niekert: nit: oops my last comment was on single line but these are still unused

niekert (approved):
Code LGTM. QA too 🙌 

During QA I noticed:

- When you search for "Desktop" it matches a Frame Node named Desktop but it doesn't display the node in the result, so you see an empty header of the WebPage
- You have to restart the plugin for nodes added in the current scope while the plugin is active to be indexed

Not sure if already aware, can be followups ofc!

image

plugins/global-search/src/utils/filter/execute-filter.ts
niekert: I like this split 🙏

plugins/global-search/src/utils/filter/types.ts
elmarburke: Oh, I thought I was using this one now (I brought them back after deleting it first). Anyhow, it's gone now 💥

Issue #345, opened 2025-07-28
Keep HubSpot plugin open after logging out


madebyisaacr:
When you log out of the HubSpot plugin, it closes the plugin and tells you to reopen it to sign back in. It should keep the plugin open and reload it like other plugins with a log out option do.

https://github.com/framer/plugins/pull/332#issuecomment-3091367043

PR #376, opened 2025-08-12
HubSpot: keep plugin open after logging out

madebyisaacr:
### Description

This pull request improves the log out UX in the HubSpot plugin.

Instead of closing the plugin, it reloads and shows a notification. This is consistent with all of our other plugins that have log out functionality.

Closes https://github.com/framer/plugins/issues/345

### Testing

- [ ] Sign in with HubSpot
- [ ] Test log out button in menu
- [ ] Test log out button on account page

### DIFF
--- a/plugins/hubspot/src/App.tsx
+++ b/plugins/hubspot/src/App.tsx
@@ -182,6 +182,7 @@ export function App() {
             const isInCMSModes = mode === "syncManagedCollection" || mode === "configureManagedCollection"
 
             if (!isAuthenticated) {
+                void framer.setMenu([])
                 navigate("/")
                 return
             }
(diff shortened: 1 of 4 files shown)


plugins/hubspot/src/App.tsx
```diff
@@ -182,6 +182,7 @@ export function App() {
             const isInCMSModes = mode === "syncManagedCollection" || mode === "configureManagedCollection"
 
             if (!isAuthenticated) {
+                void framer.setMenu([])
```
elmarburke: Could you achieve this by using an `useEffect` to reactively update the `framer.setMenu(...)` based on `isAuthenticated`? This way, the `getContexts` will only perform the actions you expect it to when viewed from the outside.

madebyisaacr: Yes, that's a better solution. Done.

PR #383, opened 2025-08-14
chore(code-versions): update framer-plugin to version 3.6.0-alpha.1

elmarburke:
### Description

Upgrade to most recent `framer-plugin` version and implement BC breaks

### Testing

- [ ] In `mode=code`, it shows the history for the right file


elmarburke:
@niekert We haven’t yet graduated `mode=code` to the stable branch yet, but once we do, we should definitely move to it!

niekert:
ohhh I thought that was part of the code files releases. my bad, I was too eager to remove the still-alive dead code

PR #391, opened 2025-08-20
refactor(global-search): replace useFilter with useAsyncFilter

elmarburke:
### Description

For the global search plugin, replacing the synchronous filter with one that can be paused and reports back progress.
Together with the heavy use of `startTransition`, this will ensure the UI remains fluid and not be slow.

### Testing

- [x] Search on a big project WHILE indexing works fine.
  - Open the plugin on a big project, where indexing takes long (ask me for a remix link)
  - Start searching
  - [x] Rendering performance is okay
  - [x] the filter still works

### DIFF
--- a/plugins/global-search/src/components/SearchScene.tsx
+++ b/plugins/global-search/src/components/SearchScene.tsx
@@ -19,7 +19,7 @@ export function SearchScene() {
     const { searchOptions, optionsMenuItems } = useOptionsMenuItems()
     const deferredQuery = useDeferredValue(query)
 
-    const { results, hasResults, error: filterError, isFiltering } = useAsyncFilter(deferredQuery, searchOptions, index)
+    const { results, hasResults, error: filterError } = useAsyncFilter(deferredQuery, searchOptions, index)
 
     if (filterError) {
         console.error(filterError)
@@ -31,10 +31,8 @@ export function SearchScene() {
     }, [])
 
     useEffect(() => {
-        void framer.showUI(
-            getPluginUiOptions({ query: deferredQuery, hasResults, hasFinalResults: !isIndexing && !isFiltering })
-        )
-    }, [deferredQuery, hasResults, isIndexing, isFiltering])
+        void framer.showUI(getPluginUiOptions({ query: deferredQuery, hasResults, areResultsFinal: !isIndexing }))
+    }, [deferredQuery, hasResults, isIndexing])
 
     return (
         <main className="flex flex-col h-full">
@@ -60,7 +58,7 @@ export function SearchScene() {
             </div>
             <div className="overflow-y-auto px-3 flex flex-col flex-1">
                 {deferredQuery && hasResults && <ResultsList groupedResults={results} />}
-                {deferredQuery && !hasResults && !isIndexing && !isFiltering && <NoResults />}
+                {deferredQuery && !hasResults && !isIndexing && <NoResults />}
             </div>
         </main>
     )
--- a/plugins/global-search/src/utils/filter/AsyncProcessor.ts
+++ b/plugins/global-search/src/utils/filter/AsyncProcessor.ts
@@ -74,11 +74,6 @@ export class TimeBasedAsyncProcessor<TInput, TOutput> {
     async start(items: readonly TInput[], itemProcessor: (item: TInput) => TOutput | false): Promise<void> {
         this.abort()
 
-        if (items.length === 0) {
-            this.emitProgressUpdate([], 0, 0, false)
-            return
-        }
-
         this.events.emit("started")
 
         const abortController = new AbortController()
@@ -109,10 +104,9 @@ export class TimeBasedAsyncProcessor<TInput, TOutput> {
 
                 if (isProcessing) {
                     await waitForIdle()
-                } else {
-                    this.events.emit("completed", results)
                 }
             }
+            this.events.emit("completed", results)
         } catch (error: unknown) {
             if (abortController.signal.aborted) return
 
--- a/plugins/global-search/src/utils/filter/useAsyncFilter.ts
+++ b/plugins/global-search/src/utils/filter/useAsyncFilter.ts
@@ -9,10 +9,6 @@ import { FilterType, MatcherType, type Result } from "./types"
 export interface AsyncFilterState {
     readonly results: readonly EntryResult[]
     readonly hasResults: boolean
-    readonly isFiltering: boolean
-    readonly progress: number
-    readonly processedItems: number
-    readonly totalItems: number
     readonly error: Error | null
 }
 
@@ -26,10 +22,6 @@ export function useAsyncFilter(
     const [state, setState] = useState<AsyncFilterState>({
         results: [],
         hasResults: false,
-        isFiltering: false,
-        progress: 0,
-        processedItems: 0,
-        totalItems: 0,
         error: null,
     })
 
@@ -47,29 +39,23 @@ export function useAsyncFilter(
         processor.on("started", () => {
             // Not resetting the results here to avoid flickering
             startTransition(() => {
-                setState(prev => ({ ...prev, isFiltering: true, error: null }))
+                setState(prev => ({ ...prev, error: null }))
             })
         })
 
-        // use progress could cause item's to get removed when a new index comes in. If this causes issues,
-        // we can move to using the completed event instead, on the tradeoff of having results later
-        processor.on("progress", progress => {
+        processor.on("completed", results => {
             startTransition(() => {
-                const uiResults = groupResults(progress.results)
+                const uiResults = groupResults(results)
                 setState({
                     results: uiResults,
-                    hasResults: progress.results.length > 0,
-                    isFiltering: progress.isProcessing,
-                    progress: progress.progress,
-                    processedItems: progress.processedItems,
-                    totalItems: progress.totalItems,
-                    error: progress.error,
+                    hasResults: results.length > 0,
+                    error: null,
                 })
             })
         })
 
         processor.on("error", error => {
-            setState(prev => ({ ...prev, isFiltering: false, error }))
+            setState(prev => ({ ...prev, error }))
         })
     }
 
--- a/plugins/global-search/src/utils/filter/AsyncProcessor.test.ts
+++ b/plugins/global-search/src/utils/filter/AsyncProcessor.test.ts
@@ -70,18 +70,10 @@ describe("TimeBasedAsyncProcessor", () => {
 
         await processor.start(emptyItems, itemProcessor)
 
-        expect(startedCallback).not.toHaveBeenCalled()
-        expect(completedCallback).not.toHaveBeenCalled()
+        expect(startedCallback).toHaveBeenCalled()
+        expect(completedCallback).toHaveBeenCalled()
         expect(errorCallback).not.toHaveBeenCalled()
-
-        expect(progressCallback).toHaveBeenCalledWith({
-            results: [],
-            isProcessing: false,
-            progress: 0,
-            processedItems: 0,
-            totalItems: 0,
-            error: null,
-        })
+        expect(completedCallback).toHaveBeenCalledWith([])
     })
 
     describe("error handling", () => {


plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -0,0 +1,95 @@
...
+            setState(prev => ({ ...prev, isFiltering: false, error }))
+        })
+    }
+
+    // start processing
+    useEffect(() => {
+        const processor = processorRef.current
+        if (!processor) return
+
+        if (query.trim() === "") {
```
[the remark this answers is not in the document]
elmarburke: Eeeehm, no.

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -17,22 +17,31 @@ export function SearchScene() {
...
     const { searchOptions, optionsMenuItems } = useOptionsMenuItems()
-    const { results, hasResults } = useFilter(query, searchOptions, index)
+    const deferredQuery = useDeferredValue(query)
+
+    const { results, hasResults, error: filterError, isFiltering } = useAsyncFilter(deferredQuery, searchOptions, index)
+
+    if (filterError) {
+        console.error(filterError)
+        throw filterError
+    }
```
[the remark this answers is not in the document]
elmarburke: Yep, this is expected. There is an error boundary for the app, and this should never happen anyway.

plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -0,0 +1,123 @@
...
+                if (isProcessing) {
+                    await waitForIdle()
+                } else {
+                    this.events.emit("completed", results)
+                }
+            }
+        } catch (error: unknown) {
+            if (abortController.signal.aborted) return
+
+            this.events.emit("error", error instanceof Error ? error : new Error("Unknown error", { cause: error }))
```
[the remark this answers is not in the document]
elmarburke: If they don't support the feature, it just is a parameter they don't use. This is fine™

plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -0,0 +1,95 @@
...
+        error: null,
+    })
+
+    const itemProcessor = useMemo((): FilterFunction => {
+        const matchers = [{ type: MatcherType.Text, query, caseSensitive: false }]
+        const filters = [{ type: FilterType.RootNodes, rootNodes }]
+        return createFilterFunction(matchers, filters)
+    }, [query, rootNodes])
+
+    if (!processorRef.current) {
```
alecmev, line 34: Not `useEffect`? And not reactive `asyncProcessorOptions`?

plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        processor.on("error", error => {
+            setState(prev => ({ ...prev, isFiltering: false, error }))
+        })
+    }
+
+    // start processing
+    useEffect(() => {
+        const processor = processorRef.current
+        if (!processor) return
```
alecmev, line 65: Nit: Could do [this](https://react.dev/reference/react/useRef#how-to-avoid-null-checks-when-initializing-use-ref-later) 😛

plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        if (query.trim() === "") {
+            processor.abort()
```
alecmev: Nit: Necessary? Since the cleanup would've been called anyway, for us to get here.

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -17,22 +17,31 @@ export function SearchScene() {
     const { index, isIndexing } = useIndexer()
     const [query, setQuery] = useState("")
     const { searchOptions, optionsMenuItems } = useOptionsMenuItems()
-    const { results, hasResults } = useFilter(query, searchOptions, index)
+    const deferredQuery = useDeferredValue(query)
```
alecmev, line 20: 👌

plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -0,0 +1,123 @@
...
+    progress: Progress<TOutput>
+    completed: readonly TOutput[]
+    error: Error
+}
+
+function waitForIdle(): Promise<void> {
+    return new Promise(resolve =>
+        requestIdleCallback(() => {
+            resolve()
+        })
```
alecmev: https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#browser_compatibility 😢

plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -0,0 +1,123 @@
...
+                    currentIndex++
+                }
+
+                const isProcessing = currentIndex < items.length
+                this.emitProgressUpdate(results, currentIndex, items.length, isProcessing)
+
+                if (isProcessing) {
+                    await waitForIdle()
+                } else {
+                    this.events.emit("completed", results)
```
alecmev: Should this not be after the loop? Then I think you could also drop the special-casing of `items.length === 0` and just have this fired. Also, I see that `"completed"` isn't listened to in `useAsyncFilter`. I don't quite understand the reason for not doing that? Can you elaborate on "the tradeoff of having results later", please?

plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -0,0 +1,123 @@
...
+export class TimeBasedAsyncProcessor<TInput, TOutput> {
+    private abortController: AbortController | null = null
+    private events = new TypedEventEmitter<AsyncProcessorEvents<TOutput>>()
+    private blockLengthMs: number
+
+    public on: typeof this.events.on = (...args) => this.events.on(...args)
+
+    constructor({
+        // the default, 16ms, is the result of 1000ms / 60fps
+        blockLengthMs = 16,
```
alecmev: Nit: 7 maybe? 😛 #ProMotion (with 1ms to spare for actual UI work)

plugins/global-search/src/utils/filter/useAsyncFilter.ts
elmarburke: 1) useEffect: yeah, that's better. Will change
2) Reactivity for the for the parameters is something we don't need now - I might remove the parameter and set it in the hook instead.

plugins/global-search/src/utils/filter/AsyncProcessor.ts
elmarburke: "results later" means when all items are processed. That's a theoretically problem for HUGE lists, that stretch lot's of rendering frames: . It could potentially lead to the list appearing more outdated than it actually is (due to partial rendering).

Giving it a second (and third) thought however, that theoretical problem isn't something to optimise, it additionally simplifies the state handling in the hook.

On `length === 0`: no more an issue when listening on completed. Fixed!

elmarburke:
@alecmev I couldn't found a library that processes lists that simple. Additionally, I am considering moving the index to an IndexedDB and have the AsyncProcessor handle cursors in there.

plugins/global-search/src/utils/filter/useAsyncFilter.ts
elmarburke: There is no reason for it, but I have trust issues with the clean up function. I've remove it

plugins/global-search/src/utils/filter/useAsyncFilter.ts
elmarburke: Yeah, not sure about this one. I don't want to create instances unknown...

plugins/global-search/src/utils/filter/useAsyncFilter.ts
elmarburke: 3. after removing the parameters, no more changes could appear that would justify a useEffect 🎉

plugins/global-search/src/utils/filter/useAsyncFilter.ts
alecmev: Oh, but isn't `processorRef.current` effectively a lock? But, a nit, this is 100% fine!

plugins/global-search/src/utils/filter/useAsyncFilter.ts
alecmev: Tell me what happened, between you and the cleanup function 🥲

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -31,10 +31,8 @@ export function SearchScene() {
     }, [])
 
     useEffect(() => {
-        void framer.showUI(
-            getPluginUiOptions({ query: deferredQuery, hasResults, hasFinalResults: !isIndexing && !isFiltering })
-        )
-    }, [deferredQuery, hasResults, isIndexing, isFiltering])
+        void framer.showUI(getPluginUiOptions({ query: deferredQuery, hasResults, hasFinalResults: !isIndexing }))
```
alecmev: Nit: I trip over `hasFinalResults` a little bit, as if it's talking about different final results, which aren't the same as plain results. Maybe `areResultsFinal`?

plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -41,36 +32,30 @@ export function useAsyncFilter(
...
             startTransition(() => {
-                setState(prev => ({ ...prev, isFiltering: true, error: null }))
+                setState(prev => ({ ...prev, error: null }))
             })
         })
 
-        // use progress could cause item's to get removed when a new index comes in. If this causes issues,
-        // we can move to using the completed event instead, on the tradeoff of having results later
-        processor.on("progress", progress => {
+        processor.on("completed", results => {
```
alecmev: Oh, no more "on progress" at all? Doesn't that mean that it might take minutes until results appear? 🤔

plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -24,11 +24,15 @@ export interface AsyncProcessorEvents<TOutput> extends EventMap {
...
-        })
-    )
+    return new Promise(resolve => {
+        if ("requestIdleCallback" in window) {
+            requestIdleCallback(() => {
+                resolve()
+            })
+        } else {
+            setTimeout(resolve, 1)
+        }
```
alecmev: Nit: Nice! Maybe do this once at the top level? Not necessarily shim, just a `const`.

plugins/global-search/src/utils/filter/useAsyncFilter.ts
elmarburke: it maybe takes a whole 60ms 😨

elmarburke: At least, that's what I saw so far. If it's significantly more in other projects, I'll bring it back!

plugins/global-search/src/components/SearchScene.tsx
elmarburke: Muuuuuuuuch better name! Naming is hard...

plugins/global-search/src/utils/filter/useAsyncFilter.ts
alecmev: Ahh, okay, then all good! 🥳

alecmev (approved):
Noice! 😎 QA'ing!

alecmev:
> Rendering performance is okay

So, I tried on Kits, and when I type "w" and then "o" it takes a couple of seconds for "o" to appear in the input. As the number of results gets smaller the input becomes more responsive. Not sure if it's in scope for this PR though.

elmarburke:
> Not sure if it's in scope for this PR though.

Very good point, I saw this happening and a thing to fix. I had a "search starts a 3 characters" for a while, but that's not ideal.

Something certainly out of scope of this PR, this comes down to rendering the list

alecmev:
QA ✅ 🎉

PR #392, opened 2025-08-21
Init 3rd-party-optimizer plugin

kurtextrem:
### Description

This pull request introduces the 3rd-party-optimizer plugin.

I've introduced a `VITE_CONFIG_PATH` env var which plugins can set if they need to extend the global vite config in this repo. I use it to compile the yieldGTMCalls script with esbuild so the import becomes an inline import. It must be processed so that it becomes minified / JS, so e.g. `?raw` can't be used.

I updated yarn as well, because it's just a minor bump. The rest comes from running `yarn dedupe`. Interestingly, this lead to new TS type & eslint errors, where e.g. `await framer.closePlugin` started erroring.

### Testing

- [x] Open the plugin, click the button: It should add the snippet
- [x] Open the plugin again, the button should show "Remove": Clicking it should remove the snippet.
- [x] Add the snippet, disable the snippet from the settings: The plugin should tell you it's disabled and let you know how to enable it again.

### DIFF
--- a/plugins/3rd-party-optimizer/package.json
+++ b/plugins/3rd-party-optimizer/package.json
@@ -1,7 +1,7 @@
 {
     "name": "3rd-party-optimizer",
     "private": true,
-    "version": "1.0.0",
+    "version": "1.0.1",
     "type": "module",
     "scripts": {
         "dev": "VITE_CONFIG_PATH=$(pwd)/vite.config.ts run g:dev",


plugins/3rd-party-optimizer/package.json
```diff
@@ -0,0 +1,25 @@
...
+        "check-biome": "run g:check-biome",
+        "check-eslint": "run g:check-eslint",
+        "preview": "VITE_CONFIG_PATH=$(pwd)/vite.config.ts run g:preview",
+        "pack": "npx framer-plugin-tools@latest pack",
+        "check-typescript": "run g:check-typescript"
+    },
+    "dependencies": {
+        "react": "^18.2",
+        "react-dom": "^18.2"
+    },
```
alecmev: `framer-plugin`? 😁 As mentioned in the old PR, since the effective version of `framer-plugin` is also updated in the other plugins (resulting in changes), can you, please, put it through `yarn upgrade-interactive`? It'll update the version range in every `package.json` for you.

plugins/csv-import/src/csv.ts
```diff
@@ -207,6 +207,7 @@ function getFieldDataEntryInputForField(
 
         case "divider":
         case "unsupported":
+        default:
```
alecmev: Please, add the missing case explicitly instead, so that future additions highlight this again.

plugins/3rd-party-optimizer/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "devDependencies": {
+        "@types/node": "^22",
+        "@types/react": "^18.2",
+        "@types/react-dom": "^18.2",
+        "esbuild": "^0.25.9"
+    }
```
alecmev: `@framer/vite-config`

plugins/3rd-party-optimizer/vite.config.ts
```diff
@@ -0,0 +1,43 @@
+import { readFileSync } from "node:fs"
+import { dirname, resolve } from "node:path"
+import { fileURLToPath } from "node:url"
+import defaultConfig from "@framer/vite-config"
+import { transformSync } from "esbuild"
+import { defineConfig } from "vite"
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
```
alecmev: Nit: `import.meta.dirname` should be cool, no?

plugins/3rd-party-optimizer/vite.config.ts
```diff
@@ -0,0 +1,43 @@
...
+
+                // Compile and minify with esbuild
+                const result = transformSync(source, {
+                    loader: "ts",
+                    target: "es2022",
+                    minify: true,
+                    format: "iife",
+                })
+
+                return `export default ${JSON.stringify(result.code)}`
```
alecmev: So it's ready-to-go 😛 

```suggestion
                return `export default ${JSON.stringify(`<script>${result.code}</script>)}`
```

plugins/3rd-party-optimizer/src/yieldGTMCalls.ts
alecmev, line 1: Not suggesting you do it right now, but have you considered extracting this into its own non-plugin package, with a standalone build step? Then this plugin wouldn't need a custom Vite config. This could be accommodated relatively trivially with Turbo + it would be a normal dependency and wouldn't need an ambient declaration.

plugins/3rd-party-optimizer/package.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "3rd-party-optimizer",
```
alecmev, line 2: 100% subjective and a huge nit: `third-party-optimizer`? 😁

plugins/3rd-party-optimizer/vite.config.ts
```diff
@@ -0,0 +1,43 @@
...
+        },
+        load(id: string) {
+            if (id === "virtual:yield-gtm-calls") {
+                const filePath = resolve(__dirname, "src/yieldGTMCalls.ts")
+                const source = readFileSync(filePath, "utf-8")
+
+                // Compile and minify with esbuild
+                const result = transformSync(source, {
+                    loader: "ts",
+                    target: "es2022",
```
alecmev: Maybe?

```suggestion
                    target: "es2023",
```

plugins/3rd-party-optimizer/src/vite-env.d.ts
```diff
@@ -0,0 +1,6 @@
+/// <reference types="vite/client" />
+
+declare module "virtual:yield-gtm-calls" {
+    const content: string
+    export default content
+}
```
alecmev: I'd say this belongs in an `index.d.ts` or `globals.d.ts` or something like that 🤔 This is for Vite only, and should be exactly this:

```ts
/// <reference types="vite/client" />

interface ViteTypeOptions {
    strictImportMetaEnv: unknown
}
```

plugins/3rd-party-optimizer/src/App.css
```diff
@@ -0,0 +1,71 @@
...
+    justify-content: center;
+}
+
+.p-image {
+    padding: 4px 0;
+}
+
+.mt-5 {
+    margin-top: 5px;
+}
```
alecmev: You can just use Tailwind, you get it for free in every plugin 😉 See `code-versions`, for example.

plugins/csv-import/src/csv.ts
kurtextrem: There's no missing case as far as I can tell. This fixes a typescript error. Feel free to test this yourself locally - if you remove the default, the return type no longer matches.
To be honest, I don't even understand how this was merged in the first place. This PR doesn't update any typescript or typescript-eslint related dependency. How did it start failing?

plugins/3rd-party-optimizer/vite.config.ts
kurtextrem: Good one!

plugins/3rd-party-optimizer/src/yieldGTMCalls.ts
kurtextrem: Yeah, that's a cool idea. For now I'll pass, because this is just more of a side-project.

plugins/3rd-party-optimizer/package.json
kurtextrem: I like `3rd` more. You'll usually see `3rd-party scripts` or `3rd-party trackers` or even `3rd-party cookies`, not `Third` in any of the terms

plugins/3rd-party-optimizer/vite.config.ts
kurtextrem: Since the code is site-facing, the actually correct target is 2021. Fixed

plugins/3rd-party-optimizer/src/vite-env.d.ts
kurtextrem: I'll go with `globals.d.ts`, but I'd say this is something just the vite-env concerns? (because vite transforms it?)

kurtextrem: Update, I don't, it has to be vite-env. Otherwise:

```
6:26:17 PM [vite] Internal server error: Failed to parse source for import analysis because the content contains invalid JS syntax. If you are using JSX, make sure to name the file with the .jsx or .tsx extension.
```

plugins/3rd-party-optimizer/src/App.css
kurtextrem: Pretty cool!

plugins/3rd-party-optimizer/vite.config.ts
kurtextrem: Oh yeah, good one!

plugins/3rd-party-optimizer/package.json
kurtextrem: Thanks, added! (same for framer-plugin)

plugins/csv-import/src/csv.ts
kurtextrem: image

kurtextrem: actually, found it - added a satisfies never too.

plugins/3rd-party-optimizer/package.json
kurtextrem: Ok the cause seems to be `yarn dedupe`. Apparently no one has run it before me?

alecmev: Nah, the cause is that the ranges are defined as `^3.x` in all plugins, and when you install `3.x+1` in a single plugin, the new version still satisfies `^3.x`, so non-PnP (I think?) Yarn just puts it in `node_modules` as the top-level `framer-plugin` and now all plugins get `3.x+1`.

plugins/3rd-party-optimizer/vite.config.ts
alecmev: Nice!

plugins/cms-export/src/csv.ts
```diff
@@ -149,6 +150,10 @@ export function getDataForCSV(slugFieldName: string | null, fields: Field[], ite
                     continue
                 }
 
+                case "array": {
+                    throw new Error(`Array field ${field.name} is not supported`)
```
alecmev, line 154: Mm, can you `console.warn` and `continue` instead, please? Our approach is to avoid panicking when possible 😛

plugins/csv-import/src/csv.ts
alecmev: Nice, thank you!

plugins/3rd-party-optimizer/package.json
kurtextrem: I didn't have framer-plugin in my manifest before, in this PR. I ran yarn && yarn dedupe, which then removed the outdated 3.3.2 version - as it was a dupe (as, like you say, the dep range satisfies 3.5.2). So whatever plugin introduced 3.5.2, didn't run yarn dedupe after installing (maybe on purpose, maybe not).

kurtextrem: (I realized this after making the fixes, otherwise I'd have reverted the lock file change and someone else would have needed to make those changes) 

It's probably worth introducing a yarn dedupe CI check, or something, so that the next person installing a new framer-plugin version that matches every plugins range needs to do the work and not the person running yarn dedupe by chance.

alecmev: Ooh, I didn't realize it wasn't a check yet! Man, the list of nice-to-haves is so long 😅 Thanks for fixing!

plugins/cms-export/src/csv.ts
alecmev: This PR is good to go, by the way, sans this change!

kurtextrem: mhm jonas said this: https://framer-team.slack.com/archives/C06L5H5ADK2/p1754322628475079?thread_ts=1754321907.514529&cid=C06L5H5ADK2

alecmev: Ohh! Classic 😅 Then I think the error should be `"Logic error"` or something like that, because the problem isn't that we ran into an array field, but that the filtering preceding the loop is broken. But that's a nit.

alecmev (approved):
Nice! Thanks for the patience! 🙏😌

PR #394, opened 2025-08-25
Global Search: indexer state between opens

elmarburke:
### Description

Keep the index between plugin sessions. Refresh the index when opening.

- **Implements persistent indexing with IndexedDB**: Replaces in-memory index with browser-native persistent storage using the idb library.
- **Adds project-scoped databases**: Each Framer project gets its own isolated IndexedDB database (global-search-{projectId})
- **Introduces incremental index versioning**: Tracks index runs to enable incremental updates and cleanup of stale entries
- **Implements resumable async processing**: Uses `requestIdleCallback` to process search queries without blocking the UI, with cursor resumption for large datasets

### Testing

- [x] Filterer works when indexing and "streams" data as it get's added in the index
  - clear the database (if you have one yet)
  - open a large project, search
  - you see the results coming in as they get added
- [x] Old nodes get removed when they are no longer in the index
  - Add a text node with some unique text
  - Open the plugin, let it index
  - Verify, the item is findable
  - Remove it from the doc
  - Open the plugin again
  - Wait for the indexer to be finished
  - search, it should no longer be findable.


plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -80,33 +78,52 @@ export class TimeBasedAsyncProcessor<TInput, TOutput> {
...
         const results: TOutput[] = []
-        let currentIndex = 0
+        let processedItems = 0
+        let lastProcessedKey: string | null = null
 
         try {
-            while (currentIndex < items.length) {
+            // There is an `return` in the loop, that makes sure to exit the loop when the iterator is exhausted
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+            while (true) {
```
elmarburke, line 85: I'm not very happy about the `while(true)` here, but have a hard time thinking on something better. Moving the iterator out of here doesn't work, because we have to keep the reading transactions in tact.

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -148,34 +153,39 @@ export class GlobalSearchIndexer {
     }
 
     async start() {
+        // XXX: The indexer has no "locking mechanism" to prevent multiple instances from running at the same time in multiple tabs.
         try {
```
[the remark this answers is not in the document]
elmarburke: It has a potential for a race condition, but that would require to have two times the same project and the same plugin to be open. The chance is small and certainly out of scope of this PR.

plugins/global-search/src/utils/filter/AsyncProcessor.ts
```diff
@@ -23,55 +17,59 @@ export interface AsyncProcessorEvents<TOutput> extends EventMap {
     error: Error
 }
 
-const setIdleCallback = "requestIdleCallback" in window ? requestIdleCallback : (cb: () => void) => setTimeout(cb, 1)
+/**
+ * The idle time threshold in milliseconds.
+ * This is based on (1000ms / 120fps) - some time to render, etc.
+ */
+const idleTimeThreshold = 7
+const requestIdleCallbackShim =
```
alecmev: Nit: It's not a shim if it doesn't modify `window` 😛 It's a ponyfill! 🦄

plugins/global-search/src/utils/filter/AsyncProcessor.ts
alecmev: IMO, `while (true)` is perfectly valid! Can you, please, adjust `@typescript-eslint/no-unnecessary-condition` and add `allowConstantLoopConditions: true`?

```ts
"@typescript-eslint/no-unnecessary-condition": [
    "error",
    { allowConstantLoopConditions: true },
]
```

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -47,16 +47,16 @@ export class GlobalSearchIndexer {
...
 
-    private upsertEntries(entries: IndexEntry[]) {
-        for (const entry of entries) {
-            this.entries[entry.id] = entry
-            this.eventEmitter.emit("upsert", { entry })
-        }
+    constructor(private db: GlobalSearchDatabase) {}
+
+    private async upsertEntries(entries: IndexEntry[]): Promise<void> {
+        await this.db.upsertEntries(entries)
```
alecmev: Nit: Pretty thin wrapper no? 😛

alecmev (approved):
👌

plugins/global-search/src/utils/filter/AsyncProcessor.ts
elmarburke: Not sure if you're now just horsing around... 

I'll make it `getIdleCallback` to not have naming confusion here.

packages/eslint-config/src/index.ts
```diff
@@ -49,6 +49,7 @@ export default tseslint.config(
                     allowRegExp: false,
                 },
             ],
+            "@typescript-eslint/no-unnecessary-condition": ["error", { allowConstantLoopConditions: true }],
```
alecmev, line 53: Nit:

```suggestion

            // Enabled in strictTypeChecked, but we're cool with while-true
            "@typescript-eslint/no-unnecessary-condition": ["error", { allowConstantLoopConditions: true }],
```

alecmev (approved):

PR #396, opened 2025-08-27
Global search: focus management

elmarburke:
### Description

- auto focus the input when the plugin opens
- adds [react-aria](https://react-spectrum.adobe.com/react-aria)
- introducing "arrow up" and "arrow down" navigation: When typing, pressing ↓ will lead you to the first result (skipping all the headers and stuff)

### Testing

- [x] Search Box is autofocused
  - open the plugin
  - focus is right there
- [x] Arrow Navigation works
  -  start to type
  - press ↓ and (later when down there) ↑ to navigate to an item you like
  - hit space or enter to go to. No mouse needed!


plugins/global-search/src/components/Results.tsx
```diff
@@ -50,20 +52,27 @@ function ResultPerEntry({ entry, results }: { entry: EntryResult["entry"]; resul
...
-            <summary className="flex flex-row gap-2 justify-start items-center h-8 select-none overflow-hidden ps-2 pt-2 sticky top-0 bg-modal-light dark:bg-modal-dark">
-                <div className="flex-shrink-0 flex gap-2 justify-start items-center">
-                    <IconArrowRight
-                        className="text-tertiary-light dark:text-tertiary-dark  transition-transform duration-200 ease-in-out group-open:rotate-90"
-                        aria-hidden="true"
-                    />
+        <details open className="group flex flex-col not-last:pb-2">
+            <summary
+                className="pt-2 sticky top-0 group focus-visible:outline-none bg-modal-light dark:bg-modal-dark"
+                onKeyDown={onKeyDown}
```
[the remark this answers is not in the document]
elmarburke: The `<summary>` element is the receiver of the events when it is active. It's needed here.

plugins/global-search/src/utils/useFocus.ts
```diff
@@ -0,0 +1,36 @@
...
+                    break
+                case "ArrowUp":
+                    event.preventDefault()
+                    next = focusManager?.focusPrevious({
+                        accept: isResultMatch,
+                    })
+                    break
+            }
+
+            next?.scrollIntoView({ behavior: "smooth", block: "center" })
```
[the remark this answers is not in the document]
elmarburke: No.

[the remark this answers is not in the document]
elmarburke: no.

plugins/global-search/src/components/Results.tsx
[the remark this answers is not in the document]
elmarburke: The `<summary>` element is the receiver of the events when it is active. It's needed here.

plugins/global-search/src/components/Results.tsx
```diff
@@ -103,11 +112,18 @@ function Match({
     }, [targetId, collectionFieldId])
 
     const { before, highlight, after } = useHighlightedTextWithContext({ text, range })
+    const { onKeyDown } = useFocusHandlers()
 
     return (
         <button
             onClick={navigateToResult}
-            className="text-secondary-light dark:text-secondary-dark text-xs h-6 text-left select-none cursor-pointer pl-5 hover:bg-option-light dark:hover:bg-option-dark hover:text-primary-light dark:hover:text-primary-dark rounded-lg transition-colors"
+            className={cn(
```
alecmev, line 120: Nit: Could use ``className={`foo\nbar`}`` for formatting, buuut, maybe keep it a single string if possible, so that [automatic sorting](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) (which will hopefully be added in the future) can do its thing.

plugins/global-search/src/utils/useFocus.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useCallback } from "react"
+import { useFocusManager } from "react-aria"
+
+export function useFocusHandlers() {
+    const focusManager = useFocusManager()
+
+    const onKeyDown = useCallback(
+        (event: React.KeyboardEvent<HTMLElement>) => {
+            const isResultMatch = (element: Element) =>
+                element.id === "text-search-input" || element.matches("[data-result-match]")
```
alecmev: There must be a non-stringly-typed way to do this, no? 🤔 Via some sort of ref collection maybe?

elmarburke: I was thinking on that, but didn't really felt it's worth it to add that complexity.

plugins/global-search/src/components/Results.tsx
elmarburke: I was thinking to break up the re-usable bits into "shared strings" that can be combined again. `cn` takes care to filter out "conflicts" already.
I remember that the sorting wasn't working in multi-line strings 😢

alecmev: > "shared strings"

Smells like a component to me 😁 But anyway, just a nit.

plugins/global-search/src/utils/useFocus.ts
elmarburke: Changed it to something less loose – curious what you think

plugins/global-search/src/components/Results.tsx
```diff
@@ -103,11 +112,18 @@ function Match({
...
+                "text-secondary-light dark:text-secondary-dark text-xs h-6 text-left select-none cursor-pointer pl-5  rounded-lg transition-colors",
+                "hover:bg-option-light dark:hover:bg-option-dark hover:text-primary-light dark:hover:text-primary-dark focus-visible:bg-option-light dark:focus-visible:bg-option-dark focus-visible:text-primary-light dark:focus-visible:text-primary-dark",
+                "focus-visible:outline-none focus-visible:ring-0 focus-visible:ring-offset-0"
+            )}
+            data-result-match
```
alecmev: Right?

```suggestion
```

plugins/global-search/src/components/SearchInput.tsx
```diff
@@ -1,19 +1,25 @@
...
 
 export function SearchInput({ className, ...props }: SearchInputProps) {
+    const focusProps = useFocusHandlers("input")
+
     return (
         <label className={cn("flex items-center gap-2 flex-1", className)}>
             <span className="sr-only">Search for anything in your Framer project</span>
             <IconSearch aria-hidden className="text-tertiary-light dark:text-tertiary-dark" />
             <input
+                id="text-search-input"
```
alecmev: And this?

```suggestion
```

plugins/global-search/src/utils/useFocus.ts
```diff
@@ -0,0 +1,46 @@
+import type { FocusableElement } from "@react-types/shared"
+import { useCallback } from "react"
+import { useFocusManager } from "react-aria"
+
+const inputId = "text-search-input"
+const resultDataAttribute = "data-result-match"
+
+const propsByType = {
+    input: { id: inputId },
```
alecmev: Must it be an ID? Could it not also be a data attribute? And does it have to be different from `data-result-match`?

elmarburke: Nah, I changed to to a more general approach just now

plugins/global-search/src/utils/useFocus.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const navigationSelectableAttribute = "data-navigation-selectable"
+
+const isResultMatch = (element: Element) => element.hasAttribute(navigationSelectableAttribute)
+
+export function useFocusHandlers({
+    isSelfSelectable,
+}: {
+    /** If the element itself can be naviagated to (e.g. input or individual match), or if it is just the receiver of the key events (e.g. match group) */
+    isSelfSelectable: boolean
```
alecmev, line 13: Nit: ("self" to me says "can select itself", but that's not the intention, no?)

```suggestion
    isSelectable: boolean
```

elmarburke: I found the `useFocusHandlers` strange to read `isSelectable` without it, and this reminded me somewhat to `align-self` - that's why I've added it.

alecmev: Oki! 😌

PR #398, opened 2025-08-28
Global search: Navigate to items in ManagedCollections

elmarburke:
### Description

Fixes a navigation issue whereby ManagedCollectionItems were not able to navigate to. This fixes it by using the `nodeId` for those all `CollectionItem`s.

### Testing

- [x] Navigation to CollectionItem works from within a Collection
- [x] Navigation to CollectionItem works from within a ManagedCollection


alecmev:
QA ✅ I assume it's expected that it doesn't scroll to managed fields though, since they aren't focusable?

PR #401, opened 2025-08-29
Global search: Skip indexing entries when they are unchanged replicas

elmarburke:
### Description

Skipping indexing for nodes that are replica nodes and their text is unchanged. This makes sure, if you have breakpoints set up, your search results aren't duplicated for each breakpoint in the list.

image

### Testing

- [x] It shows only the "primary" node, not the replicas in breakponts
 - have breakpoints set up
 - you search for text that is on all breakpoints
 - you only see one entry for the primary node, not for the replicas

- [x] it start to show the text when you changed the replica's text
 -  change the copy on one of the breakpoints (keep in mind: there is no change watch yet to update the entries. Re-open the plugin)
 - the item will be indexed now and shows in the result set

### DIFF
--- a/plugins/global-search/src/utils/indexer/indexer.ts
+++ b/plugins/global-search/src/utils/indexer/indexer.ts
@@ -78,7 +78,7 @@ export class GlobalSearchIndexer {
             for await (const node of rootNode.walk()) {
                 if (this.abortRequested) return
 
-                if (!isIndexableNode(node)) continue
+                if (!isIndexableNode(node) || !node.visible) continue
 
                 const text = await this.getNodeText(node)
 
@@ -87,7 +87,7 @@ export class GlobalSearchIndexer {
                 // skipping the entry if it's a duplicate of the original node with the same text
                 if (node.originalId) {
                     const originalNode = await framer.getNode(node.originalId)
-                    if (originalNode && isIndexableNode(originalNode)) {
+                    if (originalNode && isIndexableNode(originalNode) && originalNode.visible) {
                         const originalText = await this.getNodeText(originalNode)
                         if (originalText === text) {
                             continue
--- a/plugins/global-search/src/utils/indexer/types.ts
+++ b/plugins/global-search/src/utils/indexer/types.ts
@@ -3,7 +3,6 @@ import {
     ComponentInstanceNode,
     ComponentNode,
     isComponentInstanceNode,
-    isComponentNode,
     isTextNode,
     TextNode,
     WebPageNode,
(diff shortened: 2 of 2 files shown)


plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -72,17 +78,27 @@ export class GlobalSearchIndexer {
...
+                // skipping the entry if it's a duplicate of the original node with the same text
+                if (node.originalId) {
+                    const originalNode = await framer.getNode(node.originalId)
+                    if (originalNode && isIndexableNode(originalNode)) {
+                        const originalText = await this.getNodeText(originalNode)
+                        if (originalText === text) {
+                            continue
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
elmarburke: I'll look into performance and if parallelisation makes sense, at a different moment. This is't triggering a HTTP request but will do a `postMessage`, which is only half as bad.

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -72,17 +78,27 @@ export class GlobalSearchIndexer {
...
 
-                if (!text && !name) continue
+                // skipping the entry if it's a duplicate of the original node with the same text
+                if (node.originalId) {
+                    const originalNode = await framer.getNode(node.originalId)
+                    if (originalNode && isIndexableNode(originalNode)) {
+                        const originalText = await this.getNodeText(originalNode)
+                        if (originalText === text) {
+                            continue
+                        }
```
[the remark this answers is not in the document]
elmarburke: Error handling of the indexer is out of scope for this PR.

plugins/global-search/src/utils/indexer/types.ts
```diff
@@ -1,20 +1,28 @@
...
 export type IndexNodeRootNode = ComponentNode | WebPageNode
 export type IndexNodeRootNodeType = IndexNodeRootNode["__class"]
 
+export type IndexableNode = Extract<AnyNode, ComponentNode | TextNode | ComponentInstanceNode>
 export type IndexEntryType = AnyNode["__class"]
 
+export function isIndexableNode<T extends AnyNode>(value: T): value is T & IndexableNode {
+    if (isComponentInstanceNode(value)) return true
+    if (isTextNode(value)) return true
+    if (isComponentNode(value)) return true
```
niekert: Is it on purpose that WebPageNode is no longer indexable here? If `ComponentNode` is indexable why not a webpage node here?

elmarburke: I think I can remove this one as well...

niekert: I wonder if this should accept `AnyCanvasNode` to indicate root nodes should be excluded here?

elmarburke:
> Not sure if related to this PR but I'm getting this quite a lot
Yeah, I saw them as well and have to look into why this is happening. I feel it's related to the `id` vs `nodeId` situation

plugins/global-search/src/utils/indexer/types.ts
elmarburke: We don't have `AnyCanvasNode` in `framer-plugin`...

niekert: oh its just `CanvasNode` lol. but its no biggie just a nitpick

elmarburke: leaving it at `AnyNode`, as otherwise, i would have to introduce an other type guard

niekert (approved):
DSMGA 👏

PR #402, opened 2025-09-01
Global Search: Order Results and Search Menu items

elmarburke:
### Description

Orders the Results and the Search Menu items in the following order:

1. Pages
2. Collections
3. Components

image

### Testing

- [x] Results are ordered
  - search for a term in all three categories
  - find the order as above
- [x] Search Menu is in the right order
  - open the Search menu through the falafel icon
  - see the options ordered

### DIFF
--- a/plugins/global-search/src/components/ui/Menu.tsx
+++ b/plugins/global-search/src/components/ui/Menu.tsx
@@ -22,7 +22,6 @@ export const Menu = memo(function Menu({ items, children, className }: MenuProps
                 .showContextMenu(items, {
                     location: { x: buttonBounds.right - 5, y: buttonBounds.bottom },
                     placement: "bottom-left",
-                    width: 200,
                 })
                 .catch((error: unknown) => {
                     framer.notify(


plugins/global-search/src/utils/filter/group-results.ts
```diff
@@ -21,11 +22,19 @@ export function groupResults(items: readonly Result[]): readonly EntryResult[] {
...
+                results,
+            }
+        })
+        .sort((a, b) => comparePreferredOrder(a.entry.rootNodeType, b.entry.rootNodeType))
+}
+
+const orderOfResults = ["WebPageNode", "Collection", "ComponentNode"] satisfies RootNodeType[]
+
+export function comparePreferredOrder<T extends RootNodeType>(a: T, b: T): number {
+    return orderOfResults.indexOf(a) - orderOfResults.indexOf(b)
```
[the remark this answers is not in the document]
elmarburke: three items in the list are fine to look up, really.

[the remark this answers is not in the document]
elmarburke: That's ensured in the types. Non-issue here.

plugins/global-search/src/utils/filter/group-results.ts
```diff
@@ -21,11 +22,19 @@ export function groupResults(items: readonly Result[]): readonly EntryResult[] {
...
+                entry: first.entry,
+                results,
+            }
+        })
+        .sort((a, b) => comparePreferredOrder(a.entry.rootNodeType, b.entry.rootNodeType))
+}
+
+const orderOfResults = ["WebPageNode", "Collection", "ComponentNode"] satisfies RootNodeType[]
+
+export function comparePreferredOrder<T extends RootNodeType>(a: T, b: T): number {
```
alecmev: Does it not work without generics? 😏

I feel like it could use a more descriptive name. Out of context, without a type hint, it's not clear the preferred order of what it's talking about.

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -88,11 +89,13 @@ const optionsMenuLabels = {
     Collection: "Collections",
 } as const satisfies Record<RootNodeType, string>
 
+const sortedRootNodeTypes = entries(optionsEnabled).sort(([a], [b]) => comparePreferredOrder(a, b))
```
alecmev: Nit: Just saw `entries`. Don't have to refactor right now, but I'd advise against using it, see the bottom of [this SO answer](https://stackoverflow.com/a/60142095) for why.

elmarburke: yeah, good point, and in a bigger code base surely something to be concerned about – but here, with the objects being close, the chances are rather small it's going wrong, I feel

alecmev: Yeah, not worth changing right now! But, just to address the rationale, the same "small project" argument could also be used against TypeScript and ESLint and you name it 😁 It's always nice to be able to rely on the soundness of the types, no matter the context, IMO.

elmarburke: Kermit the Frog saying "That's a very good point"

alecmev:
QA is ✅ btw 😂

PR #405, opened 2025-09-01
Global search: async and batched group results

elmarburke:
### Description

Tweaking the performance:
- **No-blocking grouping**: group results now also uses `await waitForIdle()` and processes items as long it's time. And aborts requests. That, will make rendering less blocking, allowing the UI to update in between
- **Debounde short querys**: debouncing queries shorter than 4 characters will make sure we don't process long result lists for nothing and the user is typing.

### Testing

- [x] Short queries (<=3) still works, but with a deply
- [x] Grouping still work
- [x] UI feels snapy™


plugins/global-search/src/utils/filter/group-results.ts
```diff
@@ -1,23 +1,50 @@
...
+        // Wait for idle time before processing
+        const deadline = await waitForIdle()
+
+        // Process items while we have idle time available
+        while (itemIndex < totalItems && deadline.timeRemaining() > 0) {
+            const item = items[itemIndex]
+            if (!item) {
+                itemIndex++
+                continue
+            }
```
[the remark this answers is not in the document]
elmarburke: better safe than sorry.

plugins/global-search/src/utils/idle-utils.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * The idle time threshold in milliseconds.
+ * This is based on (1000ms / 120fps) - some time to render, etc.
+ */
+const idleTimeThreshold = 7
+
+const getIdleCallback =
+    // Unfortunately Safari doesn't support requestIdleCallback, so we need to shim it
```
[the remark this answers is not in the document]
elmarburke: what? `Unfortunatl` sounds very much like a type I would make, but... no?

plugins/global-search/src/utils/filter/group-results.ts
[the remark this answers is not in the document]
alecmev: Should this not be an assert? 🤔

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -114,3 +121,18 @@ function useOptionsMenuItems() {
...
+        const timeout = setTimeout(() => {
+            setDebouncedValue(value)
+        }, delay)
+        return () => {
+            clearTimeout(timeout)
+        }
+    }, [value, delay])
+
+    return debouncedValue
+}
```
alecmev: Nit: Move to a separate file? 😌

plugins/global-search/src/utils/useGroupResults.ts
```diff
@@ -0,0 +1,41 @@
...
+                setError(error instanceof Error ? error : new Error(String(error), { cause: error }))
+            })
+
+        return () => {
+            abortController.abort()
+            startTransition(() => {
+                setGroupedResults([])
+            })
+        }
+    }, [results])
```
alecmev: Is this effective? Looking at `AsyncProcessor` and `SearchScene`, `results` is only ever mutated, so the reference never changes for any given query, no? But I might be missing something.

alecmev: To elaborate, this might still work if `groupResults` can't keep up with `useAsyncFilter`, but if it does, it'll stop updating mid-way, no?

elmarburke: shoot, you are right! This hook might not be smart in that scenario at all...
Instead, I'll re-add it to the event handler, but with an abort controller attached.

plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -46,12 +47,17 @@ export function useAsyncFilter(
...
+            const controller = new AbortController()
+            groupingAbortControllerRef.current = controller
+
+            void groupResults(results, controller.signal).then(results => {
+                startTransition(() => {
+                    setState({
+                        results,
+                        hasResults: results.length > 0,
+                        error: null,
+                    })
```
alecmev: Could `useAsyncFilter` not absorb `groupResults`? That is, put everything in an `entryMap` straight away? Then it doesn't need to start from scratch on every `"progress"` 🤔

elmarburke: But, doing it here would mean we also need to manage the removal of items would require a significant amount of code.
The async with abort controller function now addresses the scenario where users type ahead and lose interest in the result. This is arguably the most critical case.

alecmev (approved):
Noice! 😎

elmarburke:
> In Kits, I first wrote "foo", then erased to "f", and then pressed "o", and it took a ~second for it to appear. Still quite responsive! But just reporting that not perfect yet. That expected, yes?

What this is about is to not freeze the UI and it "feels" fast. With lots of results, there is still the issue on rendering the list etc, so yeah, very much expected :)

PR #406, opened 2025-09-01
Global search: Indexer watches for change

elmarburke:
### Description

Watch for changes in the current root node and index them. That will allow for real time index updates while the plugin is open.

- Adds the project name to the database name for easier identification in the development process
- Runs the indexer for nodes and collections now in parallel
- And as a bonus to this change: the current canvas indexes first
- The indexer spinner is shown 

### Testing

- [x] Edit something on the canvas as you search for something in the same node: it updates automatically
- [x] Do the same as above, but delete the node: it's gone from the index
- [x] The spinner is shown for a sensible time before hiding, signalling "i've updated everything"

### DIFF
--- a/plugins/global-search/src/utils/db.ts
+++ b/plugins/global-search/src/utils/db.ts
@@ -20,8 +20,8 @@ export class GlobalSearchDatabase implements ResumableAsyncIterable<IndexEntry> 
     private db: IDBPDatabase<GlobalSearchDB> | null = null
     private readonly dbName: string
 
-    constructor(projectId: string) {
-        this.dbName = `global-search-${projectId}`
+    constructor(projectId: string, projectName: string) {
+        this.dbName = `global-search-${projectName}-${projectId}`
     }
 
     async open(): Promise<IDBPDatabase<GlobalSearchDB>> {
--- a/plugins/global-search/src/utils/db.ts
+++ b/plugins/global-search/src/utils/db.ts
@@ -11,6 +11,7 @@ interface GlobalSearchDB extends DBSchema {
             type: string
             rootNodeType: string
             addedInIndexRun: number
+            rootNodeIdVersion: [string, number]
         }
     }
 }
@@ -26,12 +27,20 @@ export class GlobalSearchDatabase implements ResumableAsyncIterable<IndexEntry> 
     async open(): Promise<IDBPDatabase<GlobalSearchDB>> {
         if (this.db) return this.db
 
-        this.db = await openDB<GlobalSearchDB>(this.dbName, 1, {
-            upgrade(db) {
-                const entriesStore = db.createObjectStore("entries", { keyPath: "id" })
-                entriesStore.createIndex("rootNodeType", "rootNodeType")
-                entriesStore.createIndex("type", "type")
-                entriesStore.createIndex("addedInIndexRun", "addedInIndexRun")
+        this.db = await openDB<GlobalSearchDB>(this.dbName, 2, {
+            upgrade(db, oldVersion, _newVersion, transaction) {
+                if (oldVersion < 1) {
+                    const entriesStore = db.createObjectStore("entries", { keyPath: "id" })
+                    entriesStore.createIndex("rootNodeType", "rootNodeType")
+                    entriesStore.createIndex("type", "type")
+                    entriesStore.createIndex("addedInIndexRun", "addedInIndexRun")
+                }
+
+                if (oldVersion < 2) {
+                    const entriesStore = transaction.objectStore("entries")
+                    // Add compound index for [rootNodeId, addedInIndexRun] - this is all we need
+                    entriesStore.createIndex("rootNodeIdVersion", ["rootNodeId", "addedInIndexRun"])
+                }
             },
         })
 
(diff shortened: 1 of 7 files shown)


plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -165,35 +169,84 @@ export class GlobalSearchIndexer {
...
+
+        this.eventEmitter.emit("canvasRootChangeStarted")
+
+        try {
+            if (abortController.signal.aborted) return
+
+            const lastIndexRun = await this.db.getLastIndexRun()
+            const currentIndexRun = lastIndexRun + 1
+            await this.processNodes(currentIndexRun, [rootNode as IndexNodeRootNode], abortController.signal)
+            await this.db.clearEntriesForRootNode(rootNode.id, lastIndexRun)
```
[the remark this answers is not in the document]
elmarburke: adjusted the method name to be clear that this is a "direct match" to delete those entries.

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -165,35 +169,84 @@ export class GlobalSearchIndexer {
...
             this.abortRequested = false
             this.eventEmitter.emit("started", { indexRun: currentIndexRun })
 
-            for await (const batch of this.crawlNodes(currentIndexRun, [...pages, ...components])) {
-                // this isn't a unnecassary static expression, as the value could change during the async loop
-                // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-                if (this.abortRequested) break
-                await this.db.upsertEntries(batch)
-            }
+            this.canvasSubscription ??= framer.subscribeToCanvasRoot(rootNode => {
```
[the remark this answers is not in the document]
elmarburke: with the nullish coalescing assignment (what a word to type) there should only be ever one around, when it starts again, the active one is good to be re-used.

plugins/global-search/src/utils/db.ts
```diff
@@ -109,4 +118,25 @@ export class GlobalSearchDatabase implements ResumableAsyncIterable<IndexEntry>
...
+     * This is used for incremental updates when a specific canvas root changes.
+     */
+    async clearEntriesForRootNode(rootNodeId: string, version: number): Promise<void> {
+        const db = await this.open()
+        const tx = db.transaction("entries", "readwrite")
+        const store = tx.objectStore("entries")
+        const index = store.index("rootNodeIdVersion")
+
+        // Use compound index to efficiently find entries with exact [rootNodeId, version] match
+        let cursor = await index.openCursor(IDBKeyRange.only([rootNodeId, version]))
```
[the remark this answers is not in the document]
elmarburke: we are re-assigning the following cursor. This can't be a `const`.

plugins/global-search/src/utils/db.ts
```diff
@@ -19,19 +20,27 @@ export class GlobalSearchDatabase implements ResumableAsyncIterable<IndexEntry>
     private db: IDBPDatabase<GlobalSearchDB> | null = null
     private readonly dbName: string
 
-    constructor(projectId: string) {
-        this.dbName = `global-search-${projectId}`
+    constructor(projectId: string, projectName: string) {
+        this.dbName = `global-search-${projectName}-${projectId}`
```
niekert, line 24: I suppose the action of renaming your project is rare enough that it doesn't matter too much if you lose the indexed state? 

Or should appending this to the db name be a dev only thing?

plugins/global-search/src/utils/db.ts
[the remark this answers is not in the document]
niekert: :joy:

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -165,35 +169,84 @@ export class GlobalSearchIndexer {
...
+        this.currentCanvasRootChangeAbortController = abortController
+
+        this.eventEmitter.emit("canvasRootChangeStarted")
+
+        try {
+            if (abortController.signal.aborted) return
+
+            const lastIndexRun = await this.db.getLastIndexRun()
+            const currentIndexRun = lastIndexRun + 1
+            await this.processNodes(currentIndexRun, [rootNode as IndexNodeRootNode], abortController.signal)
```
niekert: I may be missing something but I think casting `as` here is unsafe, because you always subscribe to the canvas root in `Indexer.start` and it can be other types of nodes too.

wdyt about early returning in the subscription if the root is not indexable node?

plugins/global-search/src/utils/indexer/indexer.ts
```diff
@@ -165,35 +169,84 @@ export class GlobalSearchIndexer {
...
+            this.eventEmitter.emit("canvasRootChangeCompleted")
+        }
+    }
+
+    private async processNodes(
+        currentIndexRun: number,
+        rootNodes: readonly IndexNodeRootNode[],
+        abortSignal?: AbortSignal
+    ) {
+        const validRootNodes = rootNodes.filter(rootNode => isComponentNode(rootNode) || isWebPageNode(rootNode))
```
niekert, line 204: here the `IndexNodeRootNode` is already checked to be a safe type, right?

niekert (approved):
Nice work 👏 

LGTM besides some non-blocking nice to have nits

plugins/global-search/src/utils/db.ts
elmarburke: Yeah, I figured it would just create a fresh index. But I'll remove this again once the main development stream on this plugin finalise.

plugins/global-search/src/utils/indexer/indexer.ts
elmarburke: I meant to have `rootNodes: readonly CanvasRootNode[]`, which is why the filter is here. I'll fixup the change for that!

plugins/global-search/src/utils/indexer/indexer.ts
elmarburke: See other comment below...

PR #408, opened 2025-09-02
pr.co Plugin

clementroche:
Pressroom: 69568
Documentation: https://newsroom-api.pr.co/docs


plugins/prco/public/icon.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 54 53"><path d="M 42.222 36.961 C 51.372 32.599 56.583 21.38 51.085 9.999 C 48.097 3.815 41.671 0 34.767 0 L 19.556 0 C 8.772 0 0 8.718 0 19.433 C 0 36.344 18.185 51.366 18.958 51.998 L 20.188 53 L 53.333 53 Z M 23.16 34.002 C 25.479 39.454 28.061 43.543 28.061 43.543 L 21.778 48.583 C 21.778 48.583 4.444 34.45 4.444 19.433 C 4.444 11.14 11.21 4.417 19.556 4.417 C 27.901 4.417 34.667 11.14 34.667 19.433 C 34.667 26.49 29.762 32.393 23.16 34.002 Z" fill="rgb(233,198,255)"></path></svg>
```
madebyisaacr: The icon has low contrast in light mode. To fix this we can add a background color.

Before:
image

After:
image

New icon made in the mocks project: https://framer.com/projects/Mocks--ZC4wx7CcLvmzhdm4S8XZ-2U5CZ?node=gLanSIqpV

```suggestion
<svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" fill="none" overflow="visible"><g><path d="M 0 0 L 30 0 L 30 30 L 0 30 Z" fill="rgb(12, 41, 33)"></path><path d="M 20.703 18.698 C 24.156 17.052 26.122 12.818 24.047 8.523 C 22.92 6.19 20.495 4.75 17.89 4.75 L 12.15 4.75 C 8.08 4.75 4.77 8.04 4.77 12.083 C 4.77 18.465 11.632 24.133 11.924 24.372 L 12.388 24.75 L 24.896 24.75 Z M 13.51 17.581 C 14.385 19.638 15.359 21.181 15.359 21.181 L 12.988 23.083 C 12.988 23.083 6.447 17.75 6.447 12.083 C 6.447 8.954 9 6.417 12.15 6.417 C 15.299 6.417 17.852 8.954 17.852 12.083 C 17.852 14.746 16.001 16.974 13.51 17.581 Z" fill="rgb(233,198,255)"></path></g></svg>
```

plugins/prco/src/components/SelectDataSource.tsx
```diff
@@ -0,0 +1,95 @@
...
+            {<img src={hero} alt="PrCo Hero" />}
+
+            <form onSubmit={handleSubmit}>
+                <label>
+                    <p>Press Room Id</p>
+                    <input
+                        id="pressRoomId"
+                        type="text"
+                        required
+                        placeholder="Enter Room Id…"
```
madebyisaacr:
```suggestion
                    <p>Press Room ID</p>
                    <input
                        id="pressRoomId"
                        type="text"
                        required
                        placeholder="Enter Room ID…"
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            type: "multiCollectionReference",
+            dataSourceId: ImageDataSource.name,
+            collectionId: "",
+            getValue: value => {
+                const parsed = v.parse(v.array(MediaInfoSchema), value)
+                return parsed.map(item => item.id)
+            },
+        },
+        { id: "summary", name: "Summary", type: "formattedText" },
+        { id: "body_html", name: "Body Html", type: "formattedText" },
```
madebyisaacr: If the content is imported as formatted text, then the HTML formatting won't be visible to users, so you can remove the "html" label.

```suggestion
        { id: "body_html", name: "Body", type: "formattedText" },
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            type: "number",
+            getValue: value => {
+                const shares = v.parse(SocialSchema, value)
+                return shares.facebook
+            },
+        },
+        {
+            id: "linkedin_shares",
+            key: "shares",
+            name: "Linkedin Shares",
```
madebyisaacr:
```suggestion
            name: "LinkedIn Shares",
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            type: "number",
+            getValue: value => {
+                const shares = v.parse(SocialSchema, value)
+                return shares.linkedin
+            },
+        },
+        {
+            id: "twitter_shares",
+            key: "shares",
+            name: "Twitter Shares",
```
madebyisaacr: Maybe this should be something like "X (Twitter) Shares"? It's been called "X" for a few years now, but a lot of people still call it Twitter, so it would be good to have both names here.

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            }
+            return data.output.data
+        },
+    },
+    [
+        { id: "title", name: "Title", type: "string", canBeUsedAsSlug: true },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "description_title", name: "Description Title", type: "string" },
+        { id: "description", name: "Description", type: "string" },
+        { id: "category_id", name: "Category Id", type: "number" },
```
madebyisaacr:
```suggestion
        { id: "category_id", name: "Category ID", type: "number" },
```

plugins/prco/src/App.css
```diff
@@ -0,0 +1,231 @@
...
+    width: 100%;
+    height: 45px;
+    background: linear-gradient(to bottom, transparent, var(--framer-color-bg));
+    pointer-events: none;
+}
+
+input[type="checkbox"]:checked {
+    background-color: #0c2920;
+    border-color: #0c2920;
+}
```
madebyisaacr: Checkboxes have low contrast in dark mode. Removing this section fixes it because the tint color is lighter.

Before:
image

After:
image



```suggestion
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            }
+            return data.output.data
+        },
+    },
+
+    [
+        { id: "title", name: "Title", type: "string", canBeUsedAsSlug: true },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "pressroom_id", name: "Press Room ID", type: "string" },
+        { id: "subtitle", name: "Sub Title", type: "string" },
```
madebyisaacr:
```suggestion
        { id: "subtitle", name: "Subtitle", type: "string" },
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+                    id: "image_image",
+                    name: "Image",
+                    type: "image",
+                },
+            ],
+            getValue: value => {
+                const parsed = v.parse(ImageSizesSchema, value)
+
+                return [parsed.large?.url, parsed.medium?.url, parsed.small?.url, parsed.original?.url].filter(v => v)
+            },
```
madebyisaacr, line 90: Is it useful to import multiple sizes of each image when Framer automatically handles optimizing image sizes?

I think it makes more sense to only import the largest image size as a single image field, then Framer will automatically generate multiple sizes and load the smallest usable size in each place that the images are used.

This will also reduce the time it takes to sync content, as it will have to import fewer images.

Another note: the images are not actually usable in this format, as images in gallery fields can't be connected to image properties without creating a gallery list.

image

plugins/prco/src/dataSources.ts
madebyisaacr: There are a lot of empty lines in the imported content for this field.

image

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+                throw new Error("Error parsing PrCo data")
+            }
+            return data.output.data
+        },
+    },
+    [
+        { id: "title", name: "Title", type: "string" },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "type", name: "Type", type: "string" },
```
madebyisaacr: Is `type` always "document", or are there other possible values for this field? In the testing press room ID, all documents have "document" as the type. If type is always "document" this field can be removed, if not it can stay.

image

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+        { id: "total_size", name: "Total Size", type: "number" },
+        { id: "is_locked", name: "Is Locked", type: "boolean" },
+        { id: "is_password_protected", name: "Is Password Protected", type: "boolean" },
+        // { // multiCollectionReference doesn't support multiple collections (images, movies)
+        //     id: "medias",
+        //     name: MediaDataSource.name,
+        //     type: "multiCollectionReference",
+        //     dataSourceId: MediaDataSource.name,
+        //     collectionId: "",
+        // },
```
madebyisaacr: A potential solution is to add two multi reference fields, one for images and one for movies. The original order wouldn't be preserved if the images and movies are mixed in with each other, but at least you would be able to import the media for each media kit item.

madebyisaacr (changes requested):
Great work! I haven't done a full code review yet - this review mostly focuses on fields and UX.

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            return data.output.data
+        },
+    },
+    [
+        { id: "title", name: "Title", type: "string" },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "type", name: "Type", type: "string" },
+        { id: "content_type", name: "Content Type", type: "string" },
+        { id: "has_thumb", name: "Has Thumb", type: "boolean" },
```
madebyisaacr:
```suggestion
        { id: "has_thumb", name: "Has Thumbnail", type: "boolean" },
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+
+            if (!data.success) {
+                console.log("Error parsing PrCo data:", data.issues)
+                throw new Error("Error parsing PrCo data")
+            }
+            return data.output.data
+        },
+    },
+    [
+        { id: "title", name: "Title", type: "string" },
```
madebyisaacr, line 311: Allow using the document title as slug.

```suggestion
        { id: "title", name: "Title", type: "string", canBeUsedAsSlug: true },
```

madebyisaacr (changes requested):

plugins/prco/src/components/SelectDataSource.tsx
```diff
@@ -0,0 +1,95 @@
...
+                })
+        },
+        [pressRoomId, selectedDataSourceId, onSelectPressRoomId, onSelectDataSource]
+    )
+
+    const isButtonDisabled = !pressRoomId || !selectedDataSourceId || isLoading || !isAllowedToManage
+
+    return (
+        <main className="framer-hide-scrollbar setup">
+            {<img src={hero} alt="PrCo Hero" />}
```
madebyisaacr:
```suggestion
            {<img src={hero} alt="PR.co Hero" />}
```

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,364 @@
...
+                        } catch (error) {
+                            console.error(error)
+                        }
+
+                        fieldData[field.id] = {
+                            type: field.type,
+                            value: validUrl,
+                        }
+                        break
+                    }
```
madebyisaacr: Use `framer.uploadFile()` for file types.

This is a great solution for preventing file upload errors. I'm going to experiment with using this method in the other CMS plugins.

```suggestion
                    case "image":
                    case "file": {
                        const url = v.is(StringifiableSchema, value) ? String(value) : ""

                        let validUrl: string | null = null

                        // Prevent error if image or file is not valid
                        try {
                            if (field.type === "image") {
                                const uploadedImage = await framer.uploadImage({ image: url })
                                validUrl = uploadedImage.url
                            } else {
                                const uploadedFile = await framer.uploadFile({ file: url })
                                validUrl = uploadedFile.url
                            }
                        } catch (error) {
                            console.error(error)
                        }

                        fieldData[field.id] = {
                            type: field.type,
                            value: validUrl,
                        }
                        break
                    }
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+            }))
+
+            return dataWithId
+        },
+    },
+    [
+        { id: "name", name: "Name", type: "string", canBeUsedAsSlug: true },
+        { id: "pressroom_id", name: "Press Room ID", type: "string" },
+        { id: "description", name: "Description", type: "string" },
+        { id: "layout", name: "Layout", type: "string" }, // ideally be a "enum" but no API documentation
```
madebyisaacr: You can generate the enum cases based on the values returned from the API.

```js
const layoutValues = new Set(items.map(item => item.fieldData.layout.value))

const enumCases = []

for (const value of layoutValues) {
  enumCases.push({
    id: value,
    name: value.charAt(0).toUpperCase() + value.slice(1), // Capitalize first letter
  })
}

const layoutField = {
  id: "layout",
  name: "Layout",
  type: "enum",
  cases: enumCases,
}
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,512 @@
...
+                console.log("Error parsing PrCo data:", data.issues)
+                throw new Error("Error parsing PrCo data")
+            }
+            return data.output.data
+        },
+    },
+    [
+        { id: "title", name: "Title", type: "string" },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "permalink", name: "Permalink", type: "string" },
```
madebyisaacr, line 313:
```suggestion
        { id: "permalink", name: "Permalink", type: "link" },
```

plugins/prco/package.json
```diff
@@ -0,0 +1,25 @@
...
+        "dev": "run g:dev",
+        "build": "run g:build",
+        "check-biome": "run g:check-biome",
+        "check-eslint": "run g:check-eslint",
+        "preview": "run g:preview",
+        "pack": "npx framer-plugin-tools@latest pack",
+        "check-typescript": "run g:check-typescript"
+    },
+    "dependencies": {
+        "framer-plugin": "^3.5.2",
```
madebyisaacr: I don't think there are any CMS-related changes in 3.6, but it would still be better to use the latest `framer-plugin`  version here.

```suggestion
        "framer-plugin": "^3.6.0",
```

plugins/prco/src/dataSources.ts
clementroche: smart!

plugins/prco/src/dataSources.ts
clementroche: those are relative url: (eg: `/documents/47642/`) which framer won't accept

plugins/prco/src/dataSources.ts
clementroche: It's not documented anywhere

clementroche: it's not documented but it looks safe to be removed

plugins/prco/src/dataSources.ts
clementroche: Where would you put this code tho? Currently fields are added in FieldMapping.tsx, where we don't have pulled the items from API yet.

madebyisaacr: Add a post-processing step on fields to calculate cases based on items. May not be worth for such a small improvement though, as it would also require these changes:
- Loading items before fields (so the items can be accessed by `generateCases`)
- Setting fields on every sync instead of only in the field mapping UI, as enum cases can change on each sync.

If you want to do it, FieldMapping.tsx line 201-225 is the area to look at.

```typescript
{
	id: "layout",
	name: "Layout",
	type: "string",
	generateCases: (items) => {
		const layoutValues = new Set(items.map((item) => item.fieldData.layout.value));

		const enumCases = [];

		for (const value of layoutValues) {
			enumCases.push({
				id: value,
				name: value.charAt(0).toUpperCase() + value.slice(1), // Capitalize first letter
			});
		}

		return enumCases;
	},
};

```

plugins/prco/src/dataSources.ts
clementroche: indeed!

plugins/prco/src/dataSources.ts
clementroche: Probably a bit overkill, i've made it enum with only one case @madebyisaacr. That's probably enough for now. wdyt?

madebyisaacr: @clementroche does it always have only one case?

If so then the field can be removed since there's only one possible value, right?

madebyisaacr:
@clementroche yes, I'll do another QA and code review today.

plugins/prco/src/dataSources.ts
clementroche: For now yes, it's not documented.

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,386 @@
...
+
+                        // prevent breaking if image is not valid
+                        try {
+                            if (field.type === "image") {
+                                const uploadedImage = await framer.uploadImage({ image: url })
+                                validUrl = uploadedImage.url
+                            } else {
+                                const uploadedFile = await framer.uploadFile({ file: url })
+                                validUrl = uploadedFile.url
+                            }
```
madebyisaacr: Syncing a large collection of images or files takes a very long time because it fetches and uploads each file one after the other, waiting for the previous file to finish before starting the next one. It would be much faster if it processed multiple items at a time.

The Notion plugin solved this problem with fetching pages by using a library called "p-limit" that runs multiple JS promises with a limit of 5 at a time. I'd recommend doing the same here.

Search "limit" in this file:
https://github.com/framer/plugins/blob/main/plugins/notion/src/data.ts

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,386 @@
...
+        slugFieldId: slugField.id,
+    })
+    const itemIds = new Set(items.map(item => item.id))
+    const unsyncedItemsIds = existingItemsIds.filter(existingItemId => !itemIds.has(existingItemId))
+
+    await collection.removeItems(unsyncedItemsIds)
+    await collection.addItems(items)
+    await collection.setPluginData(pressRoomIdPluginKey, pressRoomId)
+    await collection.setPluginData(dataSourceIdPluginKey, dataSource.id)
+    await collection.setPluginData(slugFieldIdPluginKey, slugField.id)
```
madebyisaacr: Use `Promise.all` to increase speed

```suggestion
    await Promise.all([
        collection.removeItems(unsyncedItemsIds),
        collection.addItems(items),
        collection.setPluginData(pressRoomIdPluginKey, pressRoomId),
        collection.setPluginData(dataSourceIdPluginKey, dataSource.id),
        collection.setPluginData(slugFieldIdPluginKey, slugField.id)
    ]);
```

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,386 @@
...
+    collection: ManagedCollection,
+    dataSource: PrCoDataSource,
+    fields: readonly PrCoField[],
+    slugField: ManagedCollectionFieldInput
+): Promise<void> {
+    const existingItemsIds = await collection.getItemIds()
+    const items = await getItems(dataSource, fields, {
+        pressRoomId: pressRoomId,
+        slugFieldId: slugField.id,
+    })
```
madebyisaacr:
```suggestion
    const [existingItemsIds, items] = await Promise.all([
        collection.getItemIds(),
        getItems(dataSource, fields, {
            pressRoomId: pressRoomId,
            slugFieldId: slugField.id,
        })
    ]);
```

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,386 @@
...
+    if (!framer.isAllowedTo(...syncMethods)) {
+        void framer.closePlugin("You are not allowed to sync this collection.", {
+            variant: "error",
+        })
+        return { didSync: false }
+    }
+
+    try {
+        const dataSource = await getDataSource(previousPressRoomId, previousDataSourceId)
+        const existingFields = await collection.getFields()
```
madebyisaacr:
```suggestion
        const [dataSource, existingFields] = await Promise.all([
            getDataSource(previousPressRoomId, previousDataSourceId),
            collection.getFields()
        ]);
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,509 @@
...
+            id: "clipping_image",
+            name: "Clipping Image",
+            type: "image",
+            getValue: value => {
+                return v.parse(ClippingImageAsImageSchema, value)
+            },
+        },
+        {
+            id: "sizes",
+            name: "Sizes",
```
madebyisaacr:
```suggestion
            id: "sizes",
            name: "Image",
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,509 @@
...
+        { id: "title", name: "Title", type: "string", canBeUsedAsSlug: true },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "pressroom_id", name: "Press Room ID", type: "string" },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "content_type", name: "Content Type", type: "string" },
+        { id: "transparent", name: "Transparent", type: "boolean" },
+        { id: "file_size", name: "File Size", type: "number" },
+        {
+            id: "sizes",
+            name: "Sizes",
```
madebyisaacr:
```suggestion
            id: "sizes",
            name: "Image",
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,509 @@
...
+        { id: "title", name: "Title", type: "string" },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "content_type", name: "Content Type", type: "string" },
+        { id: "has_thumb", name: "Has Thumbnail", type: "boolean" },
+        { id: "file_size", name: "File Size", type: "number" },
+        { id: "url", name: "URL", type: "link" },
+        {
+            id: "sizes",
+            name: "Sizes",
```
madebyisaacr:
```suggestion
            id: "sizes",
            name: "Image",
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,509 @@
...
+    },
+    [
+        { id: "title", name: "Title", type: "string", canBeUsedAsSlug: true },
+        { id: "id", name: "ID", type: "string", canBeUsedAsSlug: true },
+        { id: "pressroom_id", name: "Press Room ID", type: "string" },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "content_type", name: "Content Type", type: "string" },
+        { id: "file_size", name: "File Size", type: "number" },
+        { id: "url", name: "URL", type: "link" },
+        { id: "webm_url", name: "Webm URL", type: "link" },
```
madebyisaacr:
```suggestion
        { id: "webm_url", name: "WebM URL", type: "link" },
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,509 @@
...
+        { id: "mp4_url", name: "MP4 URL", type: "link" },
+        { id: "mp4_1080p_url", name: "MP4 1080P URL", type: "link" },
+        { id: "thumbnail_url", name: "Thumbnail", type: "image" },
+        { id: "thumbnail_1080p_url", name: "Thumbnail 1080P URL", type: "image" },
```
madebyisaacr: No need to import two sizes of the same image, for the same reason as removing the arrays of multiple image sizes from the previous review.

```suggestion
        { id: "thumbnail_url", name: "Thumbnail", type: "image" },
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,509 @@
...
+                return parsed.replaceAll("<p></p>", "")
+            },
+        },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "full_url", name: "Full URL", type: "link" },
+        { id: "state", name: "State", type: "enum", cases: [{ id: "published", name: "Published" }] }, // ideally be a "enum" but API is not documented
+        { id: "pdf", name: "PDF", type: "link" },
+        { id: "show_in_timeline", name: "Show In Timeline", type: "boolean" },
+        { id: "show_boilerplate_text", name: "Show Boilerplate Text", type: "boolean" },
+        { id: "freeform_two", name: "Freeform Two", type: "boolean" },
```
madebyisaacr: Why is it called "Freeform Two"? "Freeform" would be a better name unless the "two" is something specific in the platform.

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,386 @@
...
+        const mappedFields = dataSource.fields.filter(field =>
+            existingFields.some(existingField => existingField.id === field.id)
+        )
+
+        const slugField = dataSource.fields.find(field => field.id === previousSlugFieldId)
+        if (!slugField) {
+            framer.notify(`No field matches the slug field id “${previousSlugFieldId}”. Sync will not be performed.`, {
```
madebyisaacr:
```suggestion
            framer.notify(`No field matches the slug field ID “${previousSlugFieldId}”. Sync will not be performed.`, {
```

plugins/prco/README.md
```diff
@@ -0,0 +1,7 @@
+# pr.co Plugin
+
+Plugin for syncing pr.co data into the Framer CMS.
```
madebyisaacr:
```suggestion
# PR.co Plugin

Plugin for syncing PR.co data into the Framer CMS.
```

plugins/prco/src/data.ts
clementroche: Nice trick, it improves the import speed, I kept to 5 to be sure that it doesn't get flagged by the API.

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,404 @@
...
+} from "framer-plugin"
+import pLimit from "p-limit"
+
+import * as v from "valibot"
+import { hasOwnProperty } from "./api-types"
+import { dataSources, type PrCoDataSource, type PrCoField } from "./dataSources"
+import { assertNever, isCollectionReference } from "./utils"
+
+// This is to prevent rate limiting.
+const CONCURRENCY_LIMIT = 5
```
madebyisaacr:
```suggestion
// This is to process multiple items at a time.
const CONCURRENCY_LIMIT = 5
```

plugins/prco/framer.json
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "pr0co0",
```
alecmev: I'd use a random string, as to avoid giving IDs any meaning and then later relying on it.

```suggestion
    "id": "iphIHc",
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,508 @@
...
+    MediaKitsSchema,
+    type PrCoItem,
+    PressReleasesSchema,
+    SocialSchema,
+    TagsSchema,
+} from "./api-types"
+
+const API_URL = "https://api.pr.co/v1"
+
+export interface PrCoDataSource {
```
alecmev: Nit, not sure how others feel, but I'd use `Prco` instead of `PrCo` (in every name). `PrCo` trips me up when I read the types, and it makes the `Field` or `DataSource` part stand out less.

```suggestion
export interface PrcoDataSource {
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,508 @@
...
+    [
+        { id: "name", name: "Name", type: "string", canBeUsedAsSlug: true },
+        { id: "pressroom_id", name: "Press Room ID", type: "string" },
+        { id: "description", name: "Description", type: "string" },
+        // {
+        //     id: "layout",
+        //     name: "Layout",
+        //     type: "enum",
+        //     cases: [{ id: "classic", name: "Classic" }],
+        // },
```
alecmev: What's the plan with this?

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,508 @@
...
+
+function createDataSource(
+    {
+        name,
+        fetch,
+    }: {
+        name: string
+        fetch: (pressRoomId: string) => Promise<PrCoItem[]>
+    },
+    [idField, slugField, ...fields]: [PrCoField, PrCoField, ...PrCoField[]]
```
alecmev: If `idField` and `slugField` must always be provided then I'd make then named options in the first object-parameter, so that there's no confusion about which is which. In the data source above, for example, I see at least one example where `title` and `id` are swapped (and there's no slug to speak of).

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,508 @@
...
+    try {
+        const response = await fetch(url)
+        return await response.json()
+    } catch (error) {
+        console.error("Error fetching PrCo data:", error)
+        throw error
+    }
+}
+
+export type PrCoField = ManagedCollectionFieldInput & {
```
alecmev: Must this be an intersection? If the `ManagedCollectionFieldInput` structure were just a standalone property in `PrcoField` then `removePrCoKeys` wouldn't be needed at all. You can make exceptions for collections with `Exclude<ManagedCollectionFieldInput, { type: ... }>` and such.

Also, I think you should try removing `removePrCoKeys` regardless, since excess properties passed to the API should be okay.

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,508 @@
...
+            name: "Body",
+            type: "formattedText",
+            getValue: value => {
+                const parsed = v.parse(v.string(), value)
+                return parsed.replaceAll("<p></p>", "")
+            },
+        },
+        { id: "permalink", name: "Permalink", type: "string" },
+        { id: "full_url", name: "Full URL", type: "link" },
+        // { id: "state", name: "State", type: "enum", cases: [{ id: "published", name: "Published" }] }, // ideally be a "enum" but API is not documented
```
alecmev: Remove?

plugins/prco/src/data.ts
```diff
@@ -0,0 +1,406 @@
...
+                                console.error(`Invalid enum value for: ${field.name}: ${parsedValue}`)
+                            }
+
+                            break
+                        }
+                        default:
+                            assertNever(
+                                field,
+                                new Error(`Unsupported field type: ${(field as unknown as { type: string }).type}`)
+                            )
```
alecmev: Just `assertNever(field)` should be okay, I think, because the source of the "unexpected" field type isn't Framer but this plugin's own data source definitions, which are already verified to have only supported fields, so anything that's unsupported is a logic error.

plugins/prco/src/App.tsx
```diff
@@ -0,0 +1,84 @@
...
+            .catch((error: unknown) => {
+                console.error(`Error loading previously configured data source “${previousDataSourceId}”.`, error)
+                framer.notify(`Error loading previously configured data source “${previousDataSourceId}”.`, {
+                    variant: "error",
+                })
+            })
+            .finally(() => {
+                setIsLoading(false)
+            })
+    }, [previousDataSourceId, previousPressRoomId])
```
alecmev: As per warning?

```suggestion
    }, [previousDataSourceId, previousPressRoomId, pressRoomId])
```

plugins/prco/src/dataSources.ts
madebyisaacr: Can be deleted I think. It's an undocumented value returned from the API that only has one possible value as far as I can tell.

plugins/prco/src/dataSources.ts
clementroche: I think it's safe to just remove those fields in such way, the order is meaningless. It's an artifact from previous plugins.

```
function createDataSource(
    {
        name,
        fetch,
    }: {
        name: string
        fetch: (pressRoomId: string) => Promise<PrCoItem[]>
    },
    [...fields]: PrCoField[]
): PrcoDataSource {
    return {
        id: name,
        name,
        fields,
        fetch,
    }
}
```

plugins/prco/src/dataSources.ts
clementroche: I'm not sure how to implement this one without completly change the whole codebase.

alecmev: Okay, that's alright, I guess it's more of a tip for the future! I think `removePrCoKeys` could still be dropped though, can you try it and see if everything continues to work?

plugins/prco/src/api-types.ts
```diff
@@ -0,0 +1,150 @@
...
+    category_id: v.nullable(v.number()),
+    media_count: v.nullable(v.number()),
+    total_size: v.nullable(v.number()),
+    is_locked: v.nullable(v.boolean()),
+    is_password_protected: v.nullable(v.boolean()),
+    media: v.array(MediaInfoSchema),
+})
+export type MediaKits = v.InferOutput<typeof MediaKitsSchema>
+
+export type PrCoItem = PressReleases | Clippings | MediaKits | MediaInfo | Tags
```
alecmev:
```suggestion
export type PrcoItem = PressReleases | Clippings | MediaKits | MediaInfo | Tags
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,501 @@
...
+    try {
+        const response = await fetch(url)
+        return await response.json()
+    } catch (error) {
+        console.error("Error fetching PrCo data:", error)
+        throw error
+    }
+}
+
+export type PrCoField = ManagedCollectionFieldInput & {
```
alecmev:
```suggestion
export type PrcoField = ManagedCollectionFieldInput & {
```

plugins/prco/src/dataSources.ts
```diff
@@ -0,0 +1,501 @@
...
+     * The fields of the data source.
+     *
+     * The first field is the ID field.
+     * The rest of the fields are the fields of the data source.
+     */
+    fields: readonly PrCoField[]
+    fetch: (pressRoomId: string) => Promise<PrCoItem[]>
+}
+
+async function fetchPrCoData(url: string): Promise<unknown> {
```
alecmev:
```suggestion
async function fetchPrcoData(url: string): Promise<unknown> {
```

plugins/prco/src/dataSources.ts
clementroche: `removePrcoKeys` is still nedeed it seems. Moslty because of `getValue`

image

triozer:
@clementroche can you rebase on `framer:main`?

PR #410, opened 2025-09-03
Global Search: Remove DevToolsScene and related functionality

elmarburke:
### Description

Removes the Dev Tool for the index, now we're preparing for the release and the the entries can also be viewed in the IndexedDB

### Testing

- [x] Dev Tools are no longer visible


niekert (approved):
LFG 🚀

PR #411, opened 2025-09-03
Global Search Add keyboard shortcuts to close the plugin

elmarburke:
### Description

Add a keyboard shortcut to close the plugin:
- `esc`: that's natural
- `cmd+shift+f`: that will open the plugin, so it should close it as well

### Testing

- [x] `esc` closes the plugin
- [x] `cmd+shift+f` closes the plugin

### DIFF
--- a/plugins/global-search/src/App.tsx
+++ b/plugins/global-search/src/App.tsx
@@ -7,6 +7,19 @@ import { SearchScene } from "./components/SearchScene"
 import { IndexerProvider } from "./utils/indexer/IndexerProvider"
 import { getPluginUiOptions } from "./utils/plugin-ui"
 
+// Close the plugin when "open combination" or "escape" is pressed
+document.addEventListener("keydown", event => {
+    const isModifierPressed = event.metaKey || event.ctrlKey
+    const isOwnOpenCombination = isModifierPressed && event.shiftKey && event.code === "KeyF"
+    const isEscape = event.key === "Escape"
+
+    if (!isEscape && !isOwnOpenCombination) return
+
+    // This will close the plugin, but also show a message that the plugin was closed.
+    // We might add a "silent" option later.
+    void framer.closePlugin()
+})
+
 void framer.showUI(getPluginUiOptions({ query: undefined, hasResults: false }))
 
 const projectInfo = await framer.getProjectInfo()


plugins/global-search/src/App.tsx
```diff
@@ -7,6 +7,19 @@ import { SearchScene } from "./components/SearchScene"
...
+    const isModifierPressed = event.metaKey || event.ctrlKey
+    const isOwnOpenCombination = isModifierPressed && event.shiftKey && event.key === "f"
+    const isEscape = event.key === "Escape"
+
+    if (!isEscape && !isOwnOpenCombination) return
+
+    // This will close the plugin, but also show a message that the plugin was closed.
+    // We might add a "silent" option later.
+    void framer.closePlugin()
+})
```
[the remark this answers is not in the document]
elmarburke: that plugin will get closed (and with it the iframe it's on)

plugins/global-search/src/App.tsx
```diff
@@ -7,6 +7,19 @@ import { SearchScene } from "./components/SearchScene"
...
+// Close the plugin when "open combination" or "escape" is pressed
+document.addEventListener("keydown", event => {
+    const isModifierPressed = event.metaKey || event.ctrlKey
+    const isOwnOpenCombination = isModifierPressed && event.shiftKey && event.code === "KeyF"
+    const isEscape = event.key === "Escape"
+
+    if (!isEscape && !isOwnOpenCombination) return
+
+    // This will close the plugin, but also show a message that the plugin was closed.
+    // We might add a "silent" option later.
```
niekert, line 19: huh, the message feels weird here. I feel like we added this for CMS sync specifically but they all seem to have a custom close message.

I guess technically changing that to not display a message if no text is provided would be a breaking change, but it seems more natural? I know @alecmev wanted to separate the message from the `.close()` action but that would def be a breaking change.

Are you planning to add the silent option before release or tbd?

niekert (approved):
LGTM but the message is quite jarring imo

plugins/global-search/src/App.tsx
elmarburke: Yeah, before stable release but not before the beta.

elmarburke: I was thinking we could have a "variant" that is silent, which makes it a change without a BC break.

PR #416, opened 2025-09-04
Notion: optimize syncing files and images

madebyisaacr:
### Description

This pull request optimizes re-syncing collections in the Notion plugin by only including field values if the item has been changed since the last sync or the field type has changed. This eliminates one API call per image or file on unchanged items.

When re-syncing a database with 50 images and no changes in Notion, it now takes about 1.5-2 seconds to sync compared to 5-8 seconds before.

https://github.com/user-attachments/assets/91218d79-e2b2-43e9-8b27-e7a5105e2691

### Testing

Here's a database with 50 images:
https://www.notion.so/framer/264adf6e8c968083994bc1de22c55953?v=264adf6e8c96808aa97d000c2f134e07&source=copy_link

- [x] Sync the database
- [x] Add, remove, and modify items and make sure the changes are synced correctly
- [x] Change field types

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -201,12 +201,14 @@ export async function syncCollection(
                 return null
             }
 
-            if (fieldsById.has(pageContentProperty.id) && item.id && !isUnchanged) {
+            const skipContent = isUnchanged && !updatedFieldIds.has(pageContentProperty.id)
+            if (fieldsById.has(pageContentProperty.id) && item.id && !skipContent) {
                 const contentHTML = await getPageBlocksAsRichText(item.id)
                 fieldData[pageContentProperty.id] = { type: "formattedText", value: contentHTML }
             }
 
-            if (fieldsById.has(pageCoverProperty.id) && !isUnchanged) {
+            const skipCover = isUnchanged && !updatedFieldIds.has(pageCoverProperty.id)
+            if (fieldsById.has(pageCoverProperty.id) && !skipCover) {
                 let coverValue: string | null = null
 
                 if (item.cover) {


plugins/notion/src/data.ts
```diff
@@ -88,10 +88,20 @@ export async function syncCollection(
...
     slugField: ManagedCollectionFieldInput,
     ignoredFieldIds: Set<string>,
-    lastSynced: string | null
+    lastSynced: string | null,
+    existingFields?: readonly ManagedCollectionFieldInput[]
 ) {
     const fieldsById = new Map(fields.map(field => [field.id, field]))
 
+    // Track which fields have had their type changed
+    const fieldTypeChanged = new Map<string, boolean>()
```
kaloyanvi: we dont need to fill this with data for both changed and unchanged right? It could be just the changed ids -> `const fieldTypeChanged = new Set<string>()`

madebyisaacr: Good point. Switched to a set of changed field IDs.

plugins/notion/src/data.ts
```diff
@@ -88,10 +88,22 @@ export async function syncCollection(
...
+    // Track which fields have had their type changed
+    const updatedFieldIds = new Set<string>()
+    if (existingFields) {
+        for (const field of fields) {
+            const existingField = existingFields.find(f => f.id === field.id)
+            if (!existingField || existingField.type !== field.type) {
+                updatedFieldIds.add(field.id)
+            }
+        }
+    }
```
kaloyanvi: We are doing the `.find()` logic on every iteration which leads to On^2, what do you think of something like this?

```ts
const updatedFieldIds = new Set<string>()
const existingFieldsMap = new Map(existingFields?.map(f => [f.id, f]) ?? [])
for (const field of fields) {
    const existingField = existingFieldsMap.get(field.id)
    if (!existingField || existingField.type !== field.type) {
        updatedFieldIds.add(field.id)
    }
}
```

madebyisaacr: Done

plugins/notion/src/data.ts
```diff
@@ -184,12 +201,12 @@ export async function syncCollection(
...
             }
 
-            if (fieldsById.has(pageContentProperty.id) && item.id && !skipContent) {
+            if (fieldsById.has(pageContentProperty.id) && item.id && !isUnchanged) {
                 const contentHTML = await getPageBlocksAsRichText(item.id)
                 fieldData[pageContentProperty.id] = { type: "formattedText", value: contentHTML }
             }
 
-            if (fieldsById.has(pageCoverProperty.id)) {
+            if (fieldsById.has(pageCoverProperty.id) && !isUnchanged) {
```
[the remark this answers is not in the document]
madebyisaacr: Great catch here. The field type of content and cover can't be changed, but was broken when you would sync with content or cover disabled then enable them and sync again. Fixed that.

madebyisaacr:
Fixed another bug and added the optimization you suggested @kaloyanvi. Ready for QA now.

kaloyanvi:
Btw, for things like this we should start adding tests. We should be able to simulate all cases to update / not update based on item / field with some mock data.

PR #419, opened 2025-09-05
Global Search: Code File Support

elmarburke:
### Description

image

- adds code files to the indexer
- added icon for them

### Testing

- [ ] Search for content in code file and navigate to it
  - search for content in your code file
  - click on the result
  - you are in the code file


niekert (approved):
Code LGTM 👏 Will we add navigating to a specific line in the future?

niekert:
Nice one. I do feel like searching for text in code is quite different from searching for layers or other text that ends up on the site. Makes me wonder if searching code file contents should be off by default?

Issue #420, opened 2025-09-05
Support opening CSV Import from the canvas or marketplace


madebyisaacr:
When you open the CSV Import plugin from the canvas or from the marketplace, it shows this message and doesn't open:

<img width="394" height="78" alt="Image" src="https://github.com/user-attachments/assets/6f6aca63-a14b-42e8-b6e3-c45a03e3b963" />

It should show a popup where you can select a collection from the list of collections instead, similar to CMS Export which works fine on the canvas.

<img width="368" height="443" alt="Image" src="https://github.com/user-attachments/assets/93bfbc5e-1001-4e78-bbbb-9fff459e8274" />

PR #421, opened 2025-09-05
CSV Import: support opening from canvas and marketplace

madebyisaacr:
### Description

This pull request adds support for opening the CSV Import plugin from the canvas and from the marketplace.

Closes https://github.com/framer/plugins/issues/420

Also updated to `framer-plugin` 3.6.0

**Before this PR:** 
If you open the plugin from the canvas, it shows this notification and doesn't open:
Image

**After:**
If you open the plugin from the canvas, it shows a dropdown of all writable collections so you can pick one. The file upload button, drag and drop, and pasting are disabled when no collection is selected.
image

The dropdown isn't shown when you open the plugin from the CMS, because we can assume if you open it from a collection then you want to import into that collection.
image

### Testing

- [x] Add a sample blog to your project
- [x] Open the plugin from the canvas
- [x] Select a collection and import into it
- [x] Make sure drag and drop and pasting is not enabled when no collection is selected
- [x] Open the plugin with a collection open

### DIFF
--- a/plugins/csv-import/src/csv.ts
+++ b/plugins/csv-import/src/csv.ts
@@ -208,7 +208,6 @@ function getFieldDataEntryInputForField(
         case "array":
         case "divider":
         case "unsupported":
-        case "array":
             return new ConversionError(`Unsupported field type “${field.type}”`)
 
         default:


plugins/csv-import/src/main.tsx
```diff
@@ -10,13 +10,8 @@ if (!root) throw new Error("Root element not found")
 
 const collection = await framer.getActiveCollection()
 
-if (!collection) {
-    await framer.closePlugin("Please select a Collection to import into")
-    throw new Error("Unreachable")
-}
-
-if (collection.readonly) {
-    await framer.closePlugin("CSV Import can only be used on writable Collections")
```
kaloyanvi: why are we dropping this?

plugins/csv-import/src/App.tsx
```diff
@@ -144,6 +147,24 @@ export function App({ collection }: { collection: Collection }) {
...
+                const writableCollections = collections.filter(collection => collection.managedBy === "user")
+                setCollections(writableCollections)
+            } catch (error) {
+                console.error(error)
+                framer.notify("Failed to load collections", { variant: "error" })
+            }
+        }
+
+        void task()
+    }, [collection])
```
kaloyanvi: The async `getCollections()` call in the useEffect might complete after the component unmounts, leading to state updates on unmounted components. We should probably add a cleanup, maybe something like

```typescript
useEffect(() => {
    // Only load collections if opened without a collection already selected
    if (collection) return

    const abortController = new AbortController()

    const task = async () => {
        try {
            const collections = await framer.getCollections()
            
            // Check if component was unmounted before setting state
            if (abortController.signal.aborted) return
            
            const writableCollections = collections.filter(collection => collection.managedBy === "user")
            setCollections(writableCollections)
        } catch (error) {
            // Only handle error if component is still mounted
            if (abortController.signal.aborted) return
            
            console.error(error)
            framer.notify("Failed to load collections", { variant: "error" })
        }
    }

    void task()

    return () => {
        abortController.abort()
    }
}, [collection])
```

plugins/csv-import/src/main.tsx
madebyisaacr: `Collection.readonly` is deprecated now. The `collection.managedBy !== "user"` check right below this replaces it.

plugins/csv-import/src/App.tsx
madebyisaacr: Done

kaloyanvi (approved):

Issue #84, opened 2024-10-12
Severe Lag When Loading Large Collections in CMS Export Plugin (e.g. 2800+ Entries)


KlotzJesse:
**Describe the bug**  
When using the Framer CMS Export plugin with a large collection (e.g., 2800 entries), the entire Framer plugin becomes unresponsive and lags for about 20-30 seconds after selecting the collection from the dropdown. The UI freezes, with no loading indicator or spinner. Additionally, all 2800 entries are added to the DOM, even though the preview only shows 7 rows and 3 columns, which heavily impacts the user experience.

---

**To Reproduce**  
1. Open the Framer CMS Export plugin.
2. Select a large collection (e.g., 2800+ entries) from the dropdown.
3. Notice the lag and unresponsiveness that follows.
4. Inspect the DOM to see that all 2800 entries are rendered, despite only 7 being shown in the preview.

---

**Expected behavior**  
I expect the plugin to render a maximum of 7 entries (as the preview only shows this many) and provide a loading state or spinner to indicate progress while the data is being loaded. Only the necessary entries should be added to the DOM.

---

**Additional context**  
This issue makes it very difficult to work with large collections, and it seems that optimizing the DOM rendering for previews could greatly improve the user experience.

PR #426, opened 2025-09-08
Improve performance of CMS Export plugin with large collections

madebyisaacr:
### Description

This pull request fixes the severe lag when opening the CMS Export plugin in a large collections.

Closes https://github.com/framer/plugins/issues/84

The problem was that it generates a preview table row for all items, so if there's 2,000 items and 10 columns, that's over 20K HTML elements being generated at once.

**Changes:**
* Limit preview to 25 rows.
* Limit cell content and title length in preview.
* Added loading spinners to copy and export buttons.

https://github.com/user-attachments/assets/83a49e41-4971-4396-8cca-9e171fd0c068

### Testing

Here's a test project with about 1000 items, each with long content.
https://framer.com/remix/6ZXOm01Lt6vCpZgTM8qV

- [ ] Open the [CMS Export plugin](https://www.framer.com/marketplace/plugins/cms-export/)
- [ ] Run this PR locally and open the development plugin
- [ ] Compare speed of opening and rendering content between versions
- [ ] Check that only 25 rows are rendered in the preview
- [ ] Test copying and exporting CSV

### DIFF
--- a/plugins/cms-export/src/PreviewTable.tsx
+++ b/plugins/cms-export/src/PreviewTable.tsx
@@ -6,6 +6,8 @@ import { getDataForCSV } from "./csv"
 
 import "./PreviewTable.css"
 
+const PREVIEW_ROW_LIMIT = 25
+
 const cellTitleOverrides: Record<string, string> = {
     ":draft": "Is Draft?",
 }
@@ -32,7 +34,7 @@ export function PreviewTable({ collection }: Props) {
             const fields = await collection.getFields()
             const items = await collection.getItems()
 
-            setPreviewCSV(getDataForCSV(collection.slugFieldName, fields, items))
+            setPreviewCSV(getDataForCSV(collection.slugFieldName, fields, items, PREVIEW_ROW_LIMIT))
         }
 
         const resize = () => {
@@ -77,7 +79,7 @@ export function PreviewTable({ collection }: Props) {
                                 <tr key={`${rowIndex}`}>
                                     {row.map((cell, columnIndex) => (
                                         <td key={`${rowIndex}-${columnIndex}`} title={cell}>
-                                            {cell}
+                                            {getFirstLine(cell)}
                                         </td>
                                     ))}
                                 </tr>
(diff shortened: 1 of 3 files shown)


plugins/cms-export/src/PreviewTable.tsx
```diff
@@ -22,17 +24,11 @@ export function PreviewTable({ collection }: Props) {
     const [showGradient, setShowGradient] = useState(false)
 
     useEffect(() => {
-        void framer.showUI({
-            width: 340,
-            height: 370,
-            resizable: false,
-        })
-
```
madebyisaacr, line 30: This is a duplicate of App.tsx line 16-20

plugins/cms-export/src/PreviewTable.tsx
```diff
@@ -93,3 +88,15 @@ export function PreviewTable({ collection }: Props) {
...
+    if (text.length <= 100) return text
+
+    return `${text.substring(0, 100)}…`
+}
+
+function limitCellTitle(text: string) {
+    if (text.length <= 500) return text
+
+    return `${text.substring(0, 500)}…`
+}
```
alecmev: Is this right? The title is allowed to be 5 times longer than the content?

alecmev: Also, I'd extract the lengths into constants, just in case.

plugins/cms-export/src/csv.ts
```diff
@@ -54,9 +54,15 @@ function isFieldSupported(field: Field): field is SupportedField {
...
-export function getDataForCSV(slugFieldName: string | null, fields: Field[], items: CollectionItem[]): Rows {
+export function getDataForCSV(
+    slugFieldName: string | null,
+    fields: Field[],
+    items: CollectionItem[],
+    limit?: number
+): Rows {
     const rows: Rows = []
     const supportedFields = fields.filter(isFieldSupported)
+    const limitedItems = typeof limit === "number" ? items.slice(0, limit) : items
```
alecmev, line 65: Nit: This is fine, but could also do an early return in the `for-of` below, where it compares `rows.length - 1` (to account for the header) to `limit`.

plugins/cms-export/src/PreviewTable.tsx
madebyisaacr: Yes, the title is the hover tooltip so it can be much longer. Cells are always 103px wide so the content can be shortened to just the first part.

Issue #424, opened 2025-09-08
Support image alt text in Notion plugin


madebyisaacr:
The Notion plugin supports importing images but there's no way to add alt text to them.

We can reuse the format from CSV Import and Export plugins, which is to add another column with the `:alt` suffix.

For example, you add two properties: a files & media property and a text property with the same name plus an `:alt` suffix. The plugin would automatically link them and add it as alt text on the image.

PR #434, opened 2025-09-10
Add ability to choose alt text fields to notion

tom-james-watson:
### Description

You can now set the field type for a notion field to being the alt text for any other image field.

https://github.com/user-attachments/assets/060f3a3d-aaa2-42a8-9bf2-be3f141679a9

Closes https://github.com/framer/plugins/issues/424

### Testing

- [ ] First sync works as expected
  - Run this plugin
  - Authenticate and give access to this page https://www.notion.so/framer/Images-with-alt-text-269adf6e8c96801da179d4939c4d44c5?source=copy_link
  - Sync "Cats"
  - Check type selection works as expected
    - Ignore "Cover" field
    - Change type for "Some Alt Text" and confirm there's nothing in there about alt text, because "An Image" is set to "File" by default
    - Set "An Image" field to "Image" type
    - Change type of "Some Alt Text" and now you should be able to select "Alt Text..."->"An Image"
    - Change "An Image" type back to "File" and confirm that "Some Alt Text" reverts to plain text
    - Revert to image and alt text.
  - Imports work consistently
    - Import the collection
    - Collection should be set up as expected:
      - Cover should not be a field
      - An Image should be a field
      - Some Alt Text should not be a field
      - Values for An Image should include the relevant alt text
    - Subsequent imports don't break things
      - Press "Manage" and import the collection again and confirm nothing broke
      - Press "Sync" and check the same
    - Updating alt text works
      - Update the alt text of a row on notion
      - Sync and confirm the alt text updated

### DIFF
--- a/plugins/notion/src/FieldMapping.tsx
+++ b/plugins/notion/src/FieldMapping.tsx
@@ -1,5 +1,11 @@
 import classNames from "classnames"
-import { framer, type ManagedCollection, type ManagedCollectionField, useIsAllowedTo } from "framer-plugin"
+import {
+    framer,
+    type ManagedCollection,
+    type ManagedCollectionField,
+    type MenuItem,
+    useIsAllowedTo,
+} from "framer-plugin"
 import { useEffect, useMemo, useState } from "react"
 import {
     type FieldId,
@@ -13,10 +19,11 @@ import {
     type DataSource,
     fieldsInfoToCollectionFields,
     mergeFieldsInfoWithExistingFields,
+    parseAltTextMappings,
     parseIgnoredFieldIds,
     syncCollection,
 } from "./data"
-import { assert, syncMethods } from "./utils"
+import { syncMethods } from "./utils"
 
 const labelByFieldTypeOption: Record<ManagedCollectionField["type"], string> = {
     boolean: "Toggle",
@@ -40,9 +47,12 @@ interface FieldMappingRowProps {
     isAllowedToManage: boolean
     unsupported: boolean
     missingCollection: boolean
+    fieldsInfo: FieldInfo[]
+    ignoredFieldIds: Set<string>
     onToggleIgnored: (fieldId: string) => void
     onNameChange: (fieldId: string, name: string) => void
     onFieldTypeChange: (fieldId: string, type: ManagedCollectionField["type"]) => void
+    onAltTextFieldChange?: (fieldId: string, altTextForImageFieldId: string | null) => void
 }
 
 function FieldMappingRow({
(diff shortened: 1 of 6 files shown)


plugins/notion/src/data.ts
```diff
@@ -71,27 +71,45 @@ export async function getDataSource(databaseId: string, abortSignal?: AbortSigna
...
+    fields: ManagedCollectionFieldInput[],
     slugField: ManagedCollectionFieldInput,
     ignoredFieldIds: Set<string>,
-    lastSynced: string | null
+    lastSynced: string | null,
+    fieldsInfo: FieldInfo[]
 ) {
     const fieldsById = new Map(fields.map(field => [field.id, field]))
 
+    const altTextMappings: Record<string, string> = {}
```
alecmev, line 106: Nit:

```suggestion
    const altTextMappings = new Map<string, string>()
```

plugins/notion/src/data.ts
```diff
@@ -208,6 +226,21 @@ export async function syncCollection(
...
+
+                if (imageFieldData?.type === "image" && altTextFieldData?.type === "string") {
+                    fieldData[imageFieldId] = {
+                        ...imageFieldData,
+                        alt: altTextFieldData.value,
+                    }
+                }
+
+                // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+                delete fieldData[altTextForImageFieldId]
```
alecmev, line 241: Nit: It's fine, but could also just be ` = undefined`, no?

plugins/notion/src/FieldMapping.tsx
```diff
@@ -250,14 +347,14 @@ export function FieldMapping({
                     return
                 }
 
-                await collection.setFields(fieldsToSync)
```
alecmev, line 253: I understand why you had to do this, but there's a reason why it's outside of `syncCollection`, which is that syncing shouldn't depend on the `"ManagedCollection.setFields"` method being allowed (as it requires more permissions than the rest of the methods).

plugins/notion/src/FieldMapping.tsx
```diff
@@ -51,14 +61,71 @@ function FieldMappingRow({
...
     const { id, name, originalName, type, allowedTypes } = fieldInfo
     const isFieldUnavailable = unsupported || missingCollection
     const disabled = isFieldUnavailable || ignored || !isAllowedToManage
 
+    const fieldTypeButtonText = useMemo(() => {
+        if (!type) return "Select type..."
+
+        if (fieldInfo.altTextForImageFieldId) {
+            const targetField = fieldsInfo.find(f => f.id === fieldInfo.altTextForImageFieldId)
+            return `${targetField?.name ?? "Unknown"} Alt Text`
```
alecmev, line 80: Is it not an issue that `targetField` is allowed to be `undefined`?

plugins/notion/src/FieldMapping.tsx
```diff
@@ -40,9 +47,12 @@ interface FieldMappingRowProps {
     isAllowedToManage: boolean
     unsupported: boolean
     missingCollection: boolean
+    fieldsInfo: FieldInfo[]
+    ignoredFieldIds: Set<string>
     onToggleIgnored: (fieldId: string) => void
     onNameChange: (fieldId: string, name: string) => void
     onFieldTypeChange: (fieldId: string, type: ManagedCollectionField["type"]) => void
+    onAltTextFieldChange?: (fieldId: string, altTextForImageFieldId: string | null) => void
```
alecmev, line 55: `FieldMappingRow` is used only once, and `onAltTextFieldChange` is always provided there, how come this is optional?

plugins/notion/src/FieldMapping.tsx
```diff
@@ -51,14 +61,71 @@ function FieldMappingRow({
...
+                },
+            }))
+
+            menuItems.push({
+                label: "Alt Text...",
+                submenu: altTextSubmenu,
+            })
+        }
+
+        await framer.showContextMenu(menuItems, {
```
alecmev, line 120: Nit: I guess `handleFieldTypeClick` itself doesn't need to be `async` and this could be just `void`. Then it could be passed to `onClick` directly.

PR #435, opened 2025-09-10
feat(global-search): highlight code file match

elmarburke:
### Description

This will make a match in a CodeFile result highlight when clicked.

Sep-10-2025 14-00-45

### Testing

- [x] Navigation to a Code File Match works
  - Search for something in a code file
  - Click the result
  - It highlights the searched string
- [x] Highlighting a Collection item field remains working
  - Search for something in a collection item (not mananged)
  - click the result
  - see it highlighted in the sidebar


plugins/global-search/src/components/Results.tsx
```diff
@@ -81,86 +78,25 @@ function ResultPerEntry({ entry, results }: { entry: EntryResult["entry"]; resul
             </summary>
             <ul className="flex flex-col">
                 {resultsWithRanges.map(result => {
-                    const targetId =
+                    const props: MatchProps =
                         result.type === ResultType.CollectionItemField
-                            ? result.entry.collectionItemId
-                            : result.entry.nodeId
+                            ? {
```
niekert: nitpick so feel free to ignore but passing the result to the match component to me feels cleaner than creating this object and spreading it inline in nested JSX here

plugins/global-search/src/utils/filter/ranges.ts
```diff
@@ -30,3 +30,26 @@ export function findRanges(text: string, query: string, isCaseSensitive: boolean
 export function rangeLength([start, end]: Range): number {
     return end - start
 }
+
+/**
+ * Converts a 0-based Range (string indices) to a 1-based CodeFilePosition (line/column numbers).
+ */
+export function rangeToCodeFileLocation([startColumn, endColumn]: Range, text: string) {
```
niekert, line 37: nit: could benefit from a smoll unit test

PR #436, opened 2025-09-10
perf(global-search): use `rootNode.getNodesWithType("TextNode")` instead of .walk()

elmarburke:
### Description

Instead of "walking the tree" in the plugin and sending over unnecessary nodes, walk on the Vekter/Framer side and only receive what we need.

This shaved off two-thirds of the index time on the Framer project. But it comes with the (small) downside that it could delay indexing on HUGE root nodes.

### Testing

- [x] Indexing of text nodes still works fine™ 
  - [x] Open the plugin on a page like Framer
  - [x] Search for things, find what you expect to find on the canvas (this change doesn't change the indexer for collections or code files)

### DIFF
--- a/plugins/global-search/src/utils/indexer/types.ts
+++ b/plugins/global-search/src/utils/indexer/types.ts
@@ -1,21 +1,12 @@
-import {
-    type AnyNode,
-    ComponentInstanceNode,
-    ComponentNode,
-    isComponentInstanceNode,
-    isTextNode,
-    TextNode,
-    WebPageNode,
-} from "framer-plugin"
+import { type AnyNode, ComponentNode, isTextNode, TextNode, WebPageNode } from "framer-plugin"
 
 export type IndexNodeRootNode = ComponentNode | WebPageNode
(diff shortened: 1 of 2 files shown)


plugins/global-search/src/utils/indexer/types.ts
```diff
@@ -1,21 +1,12 @@
...
-} from "framer-plugin"
+import { type AnyNode, ComponentNode, isTextNode, TextNode, WebPageNode } from "framer-plugin"
 
 export type IndexNodeRootNode = ComponentNode | WebPageNode
 export type IndexNodeRootNodeType = IndexNodeRootNode["__class"]
 
-export type IndexableNode = Extract<AnyNode, TextNode | ComponentInstanceNode>
-export type IndexEntryType = AnyNode["__class"]
+export type IndexableNode = TextNode
+export type IndexEntryType = IndexableNode["__class"]
```
alecmev, line 7: Nit: Is there a (near) future where these will become unions again?

elmarburke: Unclear right now, but I think so.

Issue #432, opened 2025-09-09
Notion no longer refreshes database list on window focus


niekert:
## Overview

Previously the Notion plugin would automatically refresh the list of databases on window focus. This was useful because users may gave the database access to the Framer connection and then come back to the plugin window.

This is a regression introduced in https://github.com/framer/plugins/pull/221 because this behavior came out of the box with react-query previously

## Reproduction

1. Open Notion plugin to the databases screen
2. Add a new database in notion and give acess to the plugin
3. Go back to framer, notice it doesn't refresh the database list
4. You can see in network tab it no longer refreshes

PR #437, opened 2025-09-11
Refresh Notion databases on window focus

madebyisaacr:
### Description

This pull request updates the Notion plugin to refresh the databases list when the window is refocused.

Closes https://github.com/framer/plugins/issues/432

Slack: https://framer-team.slack.com/archives/C06L5H5ADK2/p1757427154222629

### Testing

- [x] Connect Notion account
- [x] Switch to another window and refocus the Notion plugin window to verify that the databases list refreshes

### DIFF
--- a/plugins/notion/src/SelectDataSource.tsx
+++ b/plugins/notion/src/SelectDataSource.tsx
@@ -10,6 +10,7 @@ enum Status {
     Loading = "loading",
     Ready = "ready",
     Error = "error",
+    Refreshing = "refreshing",
 }
 
 export function SelectDataSource({ onSelectDataSource }: SelectDataSourceProps) {
@@ -18,9 +19,9 @@ export function SelectDataSource({ onSelectDataSource }: SelectDataSourceProps) 
     const [dataSources, setDataSources] = useState<DataSource[]>([])
     const isFirstFocusRef = useRef(true)
 
-    const fetchDataSources = useCallback(async () => {
+    const fetchDataSources = useCallback(async (status: Status) => {
         try {
-            setStatus(Status.Loading)
+            setStatus(status)
             const dataSources = await getDataSources()
             setDataSources(dataSources)
             setStatus(Status.Ready)
@@ -49,11 +50,11 @@ export function SelectDataSource({ onSelectDataSource }: SelectDataSourceProps) 
                 isFirstFocusRef.current = false
                 return
             }
-            void fetchDataSources()
+            void fetchDataSources(Status.Refreshing)
         }
 
         window.addEventListener("focus", handleWindowFocus)
-        void fetchDataSources()
+        void fetchDataSources(Status.Loading)
 
         return () => {
             window.removeEventListener("focus", handleWindowFocus)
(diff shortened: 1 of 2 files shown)


plugins/notion/src/SelectDataSource.tsx
```diff
@@ -16,24 +16,49 @@ export function SelectDataSource({ onSelectDataSource }: SelectDataSourceProps)
...
+            console.error(error)
+            setStatus(Status.Error)
+        }
+    }, [])
+
+    useEffect(() => {
+        const handleWindowFocus = () => {
+            if (isFirstFocusRef.current) {
+                isFirstFocusRef.current = false
+                return
```
madebyisaacr, line 51: I made it ignore the first time the window is focused because the plugin window isn't focused until you click inside it. So without this, it loads once when the plugin opens, then again the first time you click inside the plugin.

niekert: I think it would be nice to use the visibilitychange API for this instead. I think it works in plugins as well and doesn't need focus of the iframe. And makes it easier to detect the switch from invisible to visible. 


This is also what react-query uses. See for inspo:


https://github.com/TanStack/query/blob/main/packages/query-core/src/focusManager.ts

madebyisaacr: The `visibilitychange` event doesn't work in the Framer desktop app. In the browser when you switch tabs it fires, but in the desktop app it doesn't.

niekert: oh totally mb, this probably never worked on Desktop app then :D

plugins/notion/src/SelectDataSource.tsx
```diff
@@ -16,24 +16,49 @@ export function SelectDataSource({ onSelectDataSource }: SelectDataSourceProps)
...
-                setDataSources(dataSources)
-                setStatus(Status.Ready)
-                if (dataSources.length > 0) {
-                    setSelectedDatabaseId(dataSources[0]?.id ?? null)
+    const fetchDataSources = useCallback(async () => {
+        try {
+            setStatus(Status.Loading)
+            const dataSources = await getDataSources()
+            setDataSources(dataSources)
+            setStatus(Status.Ready)
```
niekert, line 27: Maybe a nit but it does feel odd to me that refetching makes the button and select box disabled. I think when refetching its fine to do it in the background, or have a tiny indicator on the select only while still allowing interacting with it

madebyisaacr: Added a loading spinner on the select while refreshing. It now only disables the buttons the first time it loads.

image

plugins/notion/src/SelectDataSource.tsx
```diff
@@ -10,30 +10,56 @@ enum Status {
...
+                    return dataSources[0]?.id ?? null
                 }
-            } catch (error) {
-                console.error(error)
-                setStatus(Status.Error)
+
+                return prev
+            })
+        } catch (error) {
+            console.error(error)
```
niekert, line 42: nit: leftover loggo that was already there

plugins/notion/src/SelectDataSource.tsx
niekert: tyvm 🙏

plugins/notion/src/SelectDataSource.tsx
madebyisaacr: Do you think the log should be removed?

niekert: We typically avoid logging to console in production. But I realized we have proper error tracking for these plugins 😅 its fine to keep as well

PR #444, opened 2025-09-18
Global Search: Introducing Virtualised List

elmarburke:
### Description

To enhance the rendering performance, the PR is bringing a virtualised list to the results.
It brings [TanStack Virtual](https://tanstack.com/virtual/latest) for handling virtual lists.

- Removed the `<details/>`/`<summary/>` structure and went with a flat list
- introduced virtual list
- added a little fade-out effect now when you start scrolling (as per design)
- replace `startTransition` with `useTransition` to allow for skipping renders

### Testing

- [x] Performance on large results is performing well
  - Open the plugin on a large project
  - search for the company name
  - it performs well (search might take a second)
- [x] Root Node Groups still can be collapsed


plugins/global-search/src/utils/filter/group-results.ts
```diff
@@ -1,61 +1,84 @@
...
-            return {
-                entry: first.entry,
-                results,
+    const groups: PreparedGroup[] = Array.from(entryMap.values()).map(results => {
+        const [first] = results
+
+        const matches: PreparedResult[] = []
+        for (const result of results) {
+            for (const range of result.ranges) {
+                const id = `${result.id}-${range.join("-")}`
```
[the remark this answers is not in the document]
elmarburke: The ranges are unique and lightweight enough. This is fine™

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+
+                    const rowStyle = {
+                        transform: `translateY(${virtualRow.start}px)`,
+                    } satisfies CSSProperties
+
+                    const isSticky = isActiveSticky(virtualRow.index)
+                    const rowClassName = cn(
+                        "w-full left-0",
+                        isSticky ? "sticky" : "absolute",
+                        item.type === "group-header" ? "z-2" : "z-0",
```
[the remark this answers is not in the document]
elmarburke: nope, tailwind has us covered

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+        count: virtualItems.length,
+        getScrollElement: () => scrollElementRef.current,
+        estimateSize: index => {
+            const item = virtualItems[index]
+            if (!item) return 0
+            if (item.type === "group-header") return 35
+
+            const nextItem = virtualItems[index + 1]
+            const isLastMatchInGroup = nextItem?.groupId !== item.groupId
+            return isLastMatchInGroup ? 35 : 30
```
alecmev: Nit: For a follow-up, any chance of sharing these numbers with the styling of the rows? So that they don't drift apart if/when the CSS changes and rows become taller/shorter.

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+        },
+        rangeExtractor: range => {
+            activeStickyIndexRef.current = stickyIndexes.findLast(index => range.startIndex >= index) ?? 0
```
alecmev, line 35: Is this not potentially expensive? If it's called 60+ times a second when scrolling? Or does the library somehow cache and/or pre-calculate these?

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+            const next = new Set([activeStickyIndexRef.current, ...defaultRangeExtractor(range)])
+            return [...next].sort((a, b) => a - b)
```
alecmev, line 37: This also feels potentially expensive. `defaultRangeExtractor(range)` already returns a sorted list of numbers. Inserting one (if not present) should be achievable by ~a single pass~ reading the first item and calculating the offset + splicing.

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+export function ResultsList({ groups }: ResultsProps) {
+    const scrollElementRef = useRef<HTMLDivElement>(null)
+    const activeStickyIndexRef = useRef(0)
+    const { collapsedGroups, toggleGroup } = useCollapsedGroups()
+    const { virtualItems, stickyIndexes } = useProcessedResults(groups, collapsedGroups)
+
+    const isActiveSticky = useCallback((index: number) => activeStickyIndexRef.current === index, [])
+
+    const rowVirtualizer = useVirtualizer({
+        overscan: 10,
```
alecmev, line 20: Did you arrive at this through experiments? Asking since the default is `1`.

plugins/global-search/src/components/Results.tsx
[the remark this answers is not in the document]
alecmev: I don't know why it suggested `z-10` 😄 I guess it could be `z-header`? But eh.

plugins/global-search/src/components/Results.tsx
elmarburke: I found it "feeling" better when scrolling a bit faster – no science or real numbers were harmed in this experiment.

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+                                isSticky={isSticky}
+                                showFadeOut={isSticky && hasScrolled}
+                                hasTopBorder={hasScrolled}
+                                style={!isSticky ? rowStyle : undefined}
+                                className={rowClassName}
+                            />
+                        )
+                    }
+
+                    if (item.result.type === ResultType.CollectionItemField) {
```
alecmev, line 87: Nit: Could `Match` not be made to receive an `item`? Then it could do this `if` internally, and you'd only have to have a single `<Match />` here.

plugins/global-search/src/components/Results.tsx
elmarburke: It's called when TanStack Virtual changes the range, which could be that often, but normally isn't. Given this is also in the official "sticky" [example](https://tanstack.com/virtual/latest/docs/framework/react/examples/sticky), this is probably fine

plugins/global-search/src/components/Results.tsx
elmarburke: same as above. Additionally here: this contains (depending on the screen size) maybe 50 items. Sorting them like is is most likely fine

plugins/global-search/src/components/Results.tsx
```diff
@@ -1,108 +1,192 @@
...
+                    const item = virtualItems[virtualRow.index]
+                    if (!item) return null
+
+                    const isLastMatchInGroup = virtualItems[virtualRow.index + 1]?.groupId !== item.groupId
+
+                    const rowStyle = {
+                        transform: `translateY(${virtualRow.start}px)`,
+                    } satisfies CSSProperties
+
+                    const isSticky = isActiveSticky(virtualRow.index)
```
alecmev: Could this not just be?

```suggestion
                    const isSticky = activeStickyIndexRef.current === virtualRow.index
```

plugins/global-search/src/components/Results.tsx
alecmev: Ok! Sorting - sure, especially if it's pre-mostly-sorted, but it also instantiates a `Set` 🥲

plugins/global-search/src/components/Results.tsx
alecmev: I wonder if that example is optimized for ease of comprehension rather than performance 🤔 But, okay, let's worry about it if/when it actually causes issues!

plugins/global-search/src/components/Results.tsx
elmarburke: Will do that in a refactoring round later :)

alecmev (approved):
👌

plugins/global-search/src/components/Results.tsx
elmarburke: Will cover this when also looking into the thing above. Adding a TODO comment here.

plugins/global-search/src/components/Results.tsx
elmarburke: Added to my list

PR #445, opened 2025-09-19
feat(global-search): show spinner only on initial indexing

elmarburke:
### Description

This update modifies the spinner behavior to display only during the initial index run.
Now, metadata is stored in IndexedDB only after a successful run.
Next to the spinner, the behavior for "No Results" has changed as well: it will show "No Results" once the indexing is complete, but will display "Searching…" while the indexing is still in progress.

### Testing

- [ ] Opening the plugin in a fresh project shows the spinner
- [ ] Opening it again, doesn't show it anymore


plugins/global-search/src/utils/db.ts
```diff
@@ -131,4 +148,34 @@ export class GlobalSearchDatabase implements ResumableAsyncIterable<IndexEntry>
...
+
+    async hasCompletedInitialIndex(): Promise<boolean> {
+        const metadata = await this.getIndexMetadata()
+        const fullyIndexedOn = metadata?.fullyIndexedOn
+        if (fullyIndexedOn === undefined) return false
+
+        // Consider the index valid if it was completed within the last month
+        const currentTime = Date.now()
+        const oneMonthAgo = currentTime - 1000 * 60 * 60 * 24 * 30
+        return fullyIndexedOn >= oneMonthAgo
```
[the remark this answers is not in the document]
elmarburke: That's why there is a comment AND a named variable

plugins/global-search/src/utils/indexer/IndexerProvider.tsx
```diff
@@ -42,9 +51,11 @@ export function IndexerProvider({
         }
 
         const onCompleted = () => {
+            void db.setInitialIndexCompleted(Date.now())
             startTransition(() => {
                 setIsIndexing(false)
                 setDataVersion(prev => prev + 1)
+                setHasCompletedInitialIndex(true)
```
kaloyanvi, line 67: I am trying to understand how this effect works so I might be missing something but wouldnt React re-run the effect because `hasCompletedInitialIndex` is in the dependency array? 

This will then re-register all event handlers and also create a new indexing cycle? 🤔

elmarburke: That one should've been gone now – `hasCompletedInitialIndex` should never have been in there...

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -74,7 +80,8 @@ export function SearchScene() {
                 </div>
                 <div className="overflow-y-auto px-3 flex flex-col flex-1 scrollbar-hidden">
                     {queryToUse && hasResults && <ResultsList groups={results} />}
-                    {queryToUse && !hasResults && !isIndexing && !isFilterRunning && <NoResults />}
+                    {noResultsState === "searching" && <ResultMessage>Searching…</ResultMessage>}
+                    {noResultsState === "no-results" && <ResultMessage>No Results</ResultMessage>}
```
kaloyanvi, line 84: nit: Think it'd be clearer whats going on here if we have a single component which has props `queryToUse, hasResults, isFilterRunning, isIndexingWithMinimumDuration` and then we dont need a hook.

plugins/global-search/src/utils/db.ts
```diff
@@ -131,4 +148,34 @@ export class GlobalSearchDatabase implements ResumableAsyncIterable<IndexEntry>
...
+    }
+
+    async hasCompletedInitialIndex(): Promise<boolean> {
+        const metadata = await this.getIndexMetadata()
+        const fullyIndexedOn = metadata?.fullyIndexedOn
+        if (fullyIndexedOn === undefined) return false
+
+        // Consider the index valid if it was completed within the last month
+        const currentTime = Date.now()
+        const oneMonthAgo = currentTime - 1000 * 60 * 60 * 24 * 30
```
kaloyanvi: nit: maybe split `1000 * 60 * 60 * 24 * 30` into separate const somewhere top level?

.yarn/cache/fake-indexeddb-npm-6.2.2-e6c391670a-b2dd94419e.zip
kaloyanvi, line 1: do we still need this?

plugins/global-search/src/components/SearchScene.tsx
elmarburke: I thought of that, but wanted to have most of the decisions happen in this scene.

.yarn/cache/fake-indexeddb-npm-6.2.2-e6c391670a-b2dd94419e.zip
elmarburke: nope, not anymore. About to remove, now the issue is resolved... ("resolved" by removing the smoke test)

plugins/global-search/src/components/SearchScene.tsx
kaloyanvi: Sure but we are already splitting logic up in 
`const noResultsState = useNoResultsState(queryToUse, hasResults, isFilterRunning, isIndexingWithMinimumDuration)` 

so I suggested given we do that and instead of the hook we can just have the component `ResultMessage` in the same file just as you are doing with the hook.

kaloyanvi: Its not a strong opinion though so you can ignore me if you think it wont be nicer.

elmarburke: I will reconsider the structure when doing another small refactor.

PR #448, opened 2025-09-22
Global search: Design Pages Support

elmarburke:
### Description

Introducing: indexing for text on Design Pages 

image

- Updates `framer-plugin`
- Extends root nodes to also search for design pages
- Introducing the option in the search option
- Render the root node's icon in the result list

### Testing

- [x] Indexing, finding, and navigating to a design page works
  - Open the plugin on a project with a text node on a design page
  - Search for the text in the plugin
  - [x] You see the node's text in the result
  - [x] you can navigate to it with a click

### DIFF
--- a/plugins/global-search/src/components/GroupHeader.tsx
+++ b/plugins/global-search/src/components/GroupHeader.tsx
@@ -73,6 +73,7 @@ const defaultIconClassName = "text-tertiary-light dark:text-tertiary-dark"
 export function ResultIcon({ rootNodeType }: { rootNodeType: RootNodeType }) {
     switch (rootNodeType) {
         case "WebPageNode":
+        case "DesignPageNode":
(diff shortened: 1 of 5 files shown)


plugins/global-search/src/components/GroupHeader.tsx
```diff
@@ -73,6 +73,7 @@ const defaultIconClassName = "text-tertiary-light dark:text-tertiary-dark"
 export function ResultIcon({ rootNodeType }: { rootNodeType: RootNodeType }) {
     switch (rootNodeType) {
         case "WebPageNode":
+        case "DesignPageNode":
```
elmarburke, line 76: FYI: We use the same icon in Framer for both.

triozer (approved):
LGTM!!!

PR #449, opened 2025-09-23
CSV import: fix removing time value

madebyisaacr:
### Description

This pull request fixes the CSV Import plugin removing time from date values.

This doesn't fix the issue though, because time is also removed in Vekter. This PR makes it so that the full date/time value is passed to `Collection.addItems()`, so once the bug is fixed in Vekter, time values will import correctly. I tracked down the bug and how to fix it in https://github.com/framer/company/issues/34464.

### Testing

- [x] Create a CMS collection with a Date field with the Time option enabled.
- [x] Copy and paste this CSV data into the plugin:
```
"Slug","Title","Date"
"getting-started","Getting Started","2025-09-18T09:02:00.000Z"
"whats-new","What’s New","2025-09-11T04:00:00.000Z"
"styling-elements","Styling Elements","2025-08-14T15:10:00.000Z"
"importing-content","Importing Content","2025-08-04T05:00:00.000Z"
"best-practices","Best Practices","2025-07-23T23:05:00.000Z"
```
- [x] Observe that all items have time set to 12:00 AM despite each row in the CSV having a different time value, however the correct time values are passed to `collectionItems2`.

### DIFF
--- a/plugins/csv-import/src/App.tsx
+++ b/plugins/csv-import/src/App.tsx
@@ -220,7 +220,7 @@ export function App({ collection }: { collection: Collection | null }) {
 
                 await importItems(result)
             } catch (error) {
-                if (error instanceof FramerPluginClosedError) return
+                if (error instanceof FramerPluginClosedError) throw error
 
                 console.error(error)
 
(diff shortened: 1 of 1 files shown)


plugins/csv-import/src/App.tsx
```diff
@@ -220,13 +220,14 @@ export function App({ collection }: { collection: Collection | null }) {
 
                 await importItems(result)
             } catch (error) {
+                if (error instanceof FramerPluginClosedError) return
```
alecmev: Ahh, no, I'd re-throw, just let it bubble up all the way to the top. `closePlugin` throws it to make sure that all further processing stops.

madebyisaacr: Ok. Is this better?

alecmev (approved):
Perfect, cheers! 👌

PR #450, opened 2025-09-23
refactor(global-search): keep browser focus but allow for keyboard nav

elmarburke:
### Description

Changing how items are focused during keyboard navigation. Previously, the native focus was on the results. That blocked users from typing ahead. Now, it uses a "virtual focus" so typing remains possible. 

- Removed useFocusHandlers and added integrated useSelection for "virtual" focus handling in GroupHeader, Match, and SearchInput components.
- Updated ResultsList to manage virtual item IDs with useSelection.
- Removed useFocus and `react-aria`

### Testing

- [x] Type, navigate, type
  - search for something
  - use the arrow key to navigate to an item
  - type more
  - it works
- [x] Headers are not selected by up/down navigation
- [x] Clicking does set virtual focus
  - search for something, click the result with the mouse
  - the active state is looking correct
- [x] Same as above, but with the tab key

### DIFF
--- a/plugins/global-search/src/components/GroupHeader.tsx
+++ b/plugins/global-search/src/components/GroupHeader.tsx
@@ -30,6 +30,7 @@ export const GroupHeader = memo(
 
         return (
             <button
+                tabIndex={-1}
                 ref={ref}
                 onClick={onToggle}
                 className={cn(
--- a/plugins/global-search/src/components/Results.tsx
+++ b/plugins/global-search/src/components/Results.tsx
@@ -53,7 +53,7 @@ export function ResultsList({ groups }: ResultsProps) {
     return (
         <div
             ref={scrollElementRef}
-            className="flex-1 min-h-0 overflow-auto scrollbar-hidden contain-strict"
+            className="flex-1 min-h-0 overflow-auto scrollbar-hidden contain-strict focus-visible:outline-focus-ring-light focus-visible:dark:outline-focus-ring-dark focus-visible:outline-2 rounded-lg"
             role="listbox"
             aria-label="Search results"
         >
--- a/plugins/global-search/src/components/Match.tsx
+++ b/plugins/global-search/src/components/Match.tsx
@@ -49,6 +49,7 @@ export const Match = forwardRef<HTMLButtonElement, MatchProps>(function Match(pr
     return (
         <button
             ref={ref}
+            tabIndex={-1}
             onClick={navigateToResult}
             className={cn(
                 "text-secondary-light dark:text-secondary-dark text-xs w-full text-left select-none cursor-pointer pl-5 rounded-lg transition-colors h-6 left-0",
--- a/plugins/global-search/src/components/SearchScene.tsx
+++ b/plugins/global-search/src/components/SearchScene.tsx
@@ -78,7 +78,7 @@ export function SearchScene() {
                         <IconEllipsis className="text-framer-text-tertiary-light dark:text-framer-text-tertiary-dark" />
                     </Menu>
                 </div>
-                <div className="overflow-y-auto px-3 flex flex-col flex-1 scrollbar-hidden">
+                <div className="px-3 flex flex-col flex-1 scrollbar-hidden">
                     {queryToUse && hasResults && <ResultsList groups={results} />}
                     {noResultsState === "searching" && <ResultMessage>Searching…</ResultMessage>}
                     {noResultsState === "no-results" && <ResultMessage>No Results</ResultMessage>}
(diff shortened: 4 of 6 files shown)


plugins/global-search/src/components/Results.tsx
```diff
@@ -157,16 +171,18 @@ function useCollapsedGroups(): { collapsedGroups: ReadonlySet<string>; toggleGro
...
         const virtualItems: VirtualItem[] = []
+        const virtualItemIds: string[] = []
         const stickyIndexes: number[] = []
 
         for (const group of groupedResults) {
             const isExpanded = !collapsedGroups.has(group.entry.id)
 
             stickyIndexes.push(virtualItems.length)
 
+            virtualItemIds.push(`${HDR_PREFIX}${group.entry.id}`)
```
kaloyanvi: nit: Since we are reusing `${HDR_PREFIX}${group.entry.id}`, can we create a tiny helper which accepts as argument `group.entry.id`?

plugins/global-search/src/utils/selection/context.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface SelectionContextValue {
+    readonly activeId: string | null
+    /** Update item ids for navigation. Include headers with HDR_PREFIX. */
+    readonly setItems: (ids: readonly string[]) => void
+    /** Focus props for an element with given id. */
+    readonly getFocusProps: (id: string) => {
+        id: string
+        onFocus: () => void
+        onKeyDown: (e: React.KeyboardEvent<HTMLElement>) => void
```
kaloyanvi:
```suggestion
        onFocus: React.FocusEventHandler<HTMLElement>
        onKeyDown: React.KeyboardEventHandler<HTMLElement>
```

plugins/global-search/src/utils/selection/context.ts
```diff
@@ -0,0 +1,17 @@
...
+    }
+    /** Focus props for the input. */
+    readonly inputFocusProps: { onKeyDown: (e: React.KeyboardEvent<HTMLElement>) => void }
```
kaloyanvi:
```suggestion
    readonly inputFocusProps: { onKeyDown: React.KeyboardEventHandler<HTMLElement> }
```

plugins/global-search/src/utils/selection/SelectionProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    const handleKeyDown = useCallback(
+        (event: React.KeyboardEvent<HTMLElement>) => {
+            if (event.key === "ArrowDown" || event.key === "ArrowUp") {
+                event.preventDefault()
+                const currentIndex = activeId ? idsRef.current.indexOf(activeId) : null
+                const direction = event.key === "ArrowDown" ? NavigationDirection.Down : NavigationDirection.Up
+                const nextId = findNextSelectable(currentIndex, direction)
+                setActiveId(nextId)
```
kaloyanvi, line 60: Can we check if `event.repeat` is true and not navigate to the first item if we are on the last item and we are going down and event is repeating? Same for when we are on the first item and direction is up and event is repeating.

plugins/global-search/src/utils/selection/SelectionProvider.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react"
+import { isHeader, NavigationDirection } from "./constants"
+import { SelectionContext } from "./context"
+
+export function SelectionProvider({ children }: { children: React.ReactNode }) {
+    const [activeId, setActiveId] = useState<string | null>(null)
+    const idsRef = useRef<readonly string[]>([])
+
+    // scroll to active id
+    useEffect(() => {
```
kaloyanvi: wdyt of `useLayoutEffect` for this?

plugins/global-search/src/components/GroupHeader.tsx
```diff
@@ -39,8 +40,8 @@ export const GroupHeader = memo(
                     className
                 )}
                 aria-expanded={isExpanded}
-                {...focusProps}
                 {...props}
+                {...getFocusProps(headerId(entry.id))}
```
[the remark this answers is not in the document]
elmarburke: That's very much intended here!

kaloyanvi:
QA ✅

elmarburke:
As discussed offline: Split the tab and the arrow key navigation into two different things. Added an outline with focus-visbile to visually seperate the virtual and the real focus state.

PR #454, opened 2025-09-24
Support design pages in Tidy Up and Renamer

madebyisaacr:
### Description

This pull request makes the Tidy Up and Renamer plugins work in design pages.

### Testing

- [ ] Create a design page with some randomly placed frames
- [ ] Run the Tidy Up plugin in the design page
- [ ] Run the Renamer plugin and rename the frames


plugins/renamer/src/search/indexer.ts
```diff
@@ -104,15 +112,16 @@ export class Indexer {
...
                 framer.getNodesWithType("ComponentNode"),
+                framer.getNodesWithType("DesignPageNode"),
             ])
-            return [...webPages, ...componentNodes]
+            return [...webPages, ...componentNodes, ...designPages]
         }
 
         const root = await framer.getCanvasRoot()
-        if (!isWebPageNode(root) && !isComponentNode(root)) return []
+        if (!isWebPageNode(root) && !isComponentNode(root) && !isDesignPageNode(root)) return []
```
jonastreub, line 124: In `App.tsx` we check for `isVectorSetNode`, but here not, is that on purpose?

madebyisaacr: Yes. Most nodes in vector sets like paths, groups, shapes etc. are not supported in the API yet. The only thing you could rename are the variants (VectorSetItemNode)

PR #455, opened 2025-09-24
[Urgent] Fix text search not opening on safari

kaloyanvi:
### Description

**Problem**: Plugin failed to load in Safari with infinite spinner and no error messages.

**Root Cause**: Safari doesn't support requestIdleCallback in older versions, and the existing fallback detection wasn't robust enough.

**Solution**: Improved Safari detection for requestIdleCallback by using explicit typeof check instead of relying on nullish coalescing alone.

### Testing

- [x] Plugin works on Safari
- [x] Plugin still works on other browser


elmarburke (approved):
🙏

PR #457, opened 2025-09-25
Global search: limit filterer rerun

elmarburke:
### Description

Limit re-running the filterer based on the indexer state:

- When it's the **initial run**:
  - Restart the filterer on data change. Lots of new items might come in
- When it's not, and the data updates because of an edit:
  - Only the query triggers the filterer to update

### Testing

- [x] Fresh project updates the query (hit: use a large project)
  - Clear the indexedDB, open the plugin again
  - Very quickly, search for something
  - You should see the result and don't end up on "no results" (this might be hard to see, it's a success if you see 
- [x] On a fully indexed project
  - Open the plugin, search for something on the page
  - Remove that something from the TextNode
  - The plugin will still show the now "outdated" search result -> no more flickering when lots of changes happen

### DIFF
--- a/plugins/global-search/src/utils/filter/useAsyncFilter.ts
+++ b/plugins/global-search/src/utils/filter/useAsyncFilter.ts
@@ -18,7 +18,9 @@ export function useAsyncFilter(
     query: string,
     rootNodes: readonly RootNodeType[],
     database: GlobalSearchDatabase,
-    dataVersion: number
+    {
+        dataVersion,
+    }: { restartOnVersionChange: true; dataVersion: number } | { restartOnVersionChange: false; dataVersion?: never }
 ): AsyncFilterState {
     const processorRef = useRef<IdleCallbackAsyncProcessor<Result> | null>(null)
     /** If this matches the current query and root nodes, we don't update the results with each progress update but instead wait for the completed event */
@@ -92,7 +94,6 @@ export function useAsyncFilter(
         })
     }
 
-    // start processing
     useEffect(() => {
         const processor = processorRef.current
         if (!processor) return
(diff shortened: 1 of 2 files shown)


plugins/global-search/src/utils/filter/useAsyncFilter.ts
```diff
@@ -18,7 +18,9 @@ export function useAsyncFilter(
     query: string,
     rootNodes: readonly RootNodeType[],
     database: GlobalSearchDatabase,
-    dataVersion: number
+    {
+        dataVersion,
+    }: { restartOnVersionChange: true; dataVersion: number } | { restartOnVersionChange: false; dataVersion?: never }
```
kaloyanvi, line 23: isnt that just unnecessary complexity?

kaloyanvi: Can we simplify to something like `dataVersion: number | null`?

elmarburke: We can, but then the version number would implicitly be bound to its updating status. I prefer the clarity it brings.

kaloyanvi: Personally, not a big fan of the discriminated union here for something so simple. If the reason is readability I think we should ad TS doc on the `dataVersion` type and/or add an extra argument to the hook with `restartOnVersionChange`.

elmarburke: Adding a TSDoc is a good way to hide critical information. The alternative is to have two `useEffect`, one with and one without `dataVersion` as a dependency, and a mutual early return for one or the other.
If that sounds like a good solution to you, i'll add that.

kaloyanvi: Nah, not two `useEffects`. Ultimately its a nit, lets merge and move on to the next thing.

PR #458, opened 2025-09-25
Global search: spinner for queries

elmarburke:
### Description

This includes two adjustments for the spinner.

- Show the spinner for the initial indexation only when a query is entered
- Show the spinner also for long-running queries (above 0.5s)

Note: This PR may display the spinner due to filter reruns on data changes. That's covered in https://github.com/framer/plugins/pull/457.

### Testing

- [ ] no spinner visible when no query is filled in
- [ ] Initial index spinner shows when a query is filled in
- [ ] long queries show the spinner (consider enabling slowdown in devtools for this)


plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -73,9 +73,12 @@ export function SearchScene() {
...
                     <div
-                        title="Indexing..."
-                        className="aria-hidden:opacity-0 transition flex items-center justify-center"
-                        aria-hidden={!isIndexingWithMinimumDuration}
+                        title="Searching…"
+                        className={cn(
+                            "aria-hidden:opacity-0 transition-opacity flex items-center justify-center",
+                            !isInitialIndexing && "delay-500"
+                        )}
+                        aria-hidden={!queryToUse || (!isInitialIndexing && !isFilterRunning)}
```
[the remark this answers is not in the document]
elmarburke: 1) They will fade in/out, so not "flickery”.
2) When it's done, it's good to hide. This indicator has the purpose of telling the users that the final results are not in yet.

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -73,9 +73,12 @@ export function SearchScene() {
                     <SearchInput value={query} onChange={handleQueryChange} />
 
                     <div
-                        title="Indexing..."
-                        className="aria-hidden:opacity-0 transition flex items-center justify-center"
-                        aria-hidden={!isIndexingWithMinimumDuration}
+                        title="Searching…"
```
kaloyanvi, line 78: ❤️

elmarburke: `…`4life

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -59,28 +53,33 @@ export function SearchScene() {
...
+                        className={cn(
+                            "aria-hidden:opacity-0 transition-opacity flex items-center justify-center",
+                            !isInitialIndexing && "delay-500"
```
elmarburke: I'm undecided if this delay (based on "how long is the filterer running" vs "how long since the last character".

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -59,28 +55,28 @@ export function SearchScene() {
...
                 >
                     <SearchInput value={query} onChange={handleQueryChange} />
 
                     <div
-                        title="Indexing..."
-                        className="aria-hidden:opacity-0 transition flex items-center justify-center"
-                        aria-hidden={!isIndexingWithMinimumDuration}
+                        title="Searching…"
+                        className="aria-hidden:opacity-0 transition-opacity flex items-center justify-center"
+                        aria-hidden={!deferredQuery || (!isInitialIndexing && !showFiltererIndicator)}
```
kaloyanvi: nit: would be nice to have this in const

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -143,20 +138,45 @@ function useOptionsMenuItems() {
     return { searchOptions, optionsMenuItems }
 }
 
-function useNoResultsState(queryToUse: string, hasResults: boolean, isFilterRunning: boolean, isIndexing: boolean) {
-    const [noResultsState, setNoResultsState] = useState<"searching" | "no-results" | false>(false)
+/**
+ * Debounces the "no results" visibility based on the previous state
+ */
+function useDeferredNoResults({
+    query: query,
```
kaloyanvi:
```suggestion
    query,
```

plugins/global-search/src/components/SearchScene.tsx
```diff
@@ -143,20 +138,45 @@ function useOptionsMenuItems() {
     return { searchOptions, optionsMenuItems }
 }
 
-function useNoResultsState(queryToUse: string, hasResults: boolean, isFilterRunning: boolean, isIndexing: boolean) {
-    const [noResultsState, setNoResultsState] = useState<"searching" | "no-results" | false>(false)
+/**
+ * Debounces the "no results" visibility based on the previous state
+ */
+function useDeferredNoResults({
```
kaloyanvi: Where is the deferring in this hook?

elmarburke: The status causes it to "defer" the value. Not in React's `useDefferedValue`, but in that this is not a directly derived value.

kaloyanvi: Generally values that can be derived directly dont really need a hook, so then should we call most react hooks deferring?

elmarburke: no, but `useWhenIsNoResultsVisible` is a bit wordy

kaloyanvi: `useNoResults` is still fitting imo, if you have strong opinion for the name go for it.

elmarburke: updated.

PR #460, opened 2025-09-26
Global search: various visual fixes in the UI

elmarburke:
### Description

To review, enable "ignore whitespace".

- enable antialiased font rendering
- align font weights (now it's antialiased)
- redo height calculation and spacing between items to avoid "cut off" items, add index so the lib checks what is in
- align group header colours

image

### Testing

- [x] Fonts are rendered :chefkiss: (same as in Framer Studio)
- [x] The first result after a header isn't cut off by 1px
- [x] The colour of the group header icons matches those in the sidebar (as the sidebar where the pages and designs are)

### DIFF
--- a/plugins/global-search/src/components/GroupHeader.tsx
+++ b/plugins/global-search/src/components/GroupHeader.tsx
@@ -34,12 +34,13 @@ export const GroupHeader = memo(
                 ref={ref}
                 onClick={onToggle}
                 className={cn(
-                    "group w-full focus:outline-none cursor-pointer text-left",
-                    isSticky ? "sticky -top-px z-1" : "",
+                    "group w-full focus:outline-none cursor-pointer text-left data-[header=normal]:absolute data-[header=normal]:left-0 data-[header=sticky]:sticky data-[header=sticky]:-top-px data-[header=sticky]:z-10 z-20",
+                    index === 0 && "data-[header=sticky]:top-0",
                     showFadeOut &&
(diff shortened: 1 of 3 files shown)


plugins/global-search/index.html
```diff
@@ -1,14 +1,13 @@
...
-  
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/icon.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Global Search</title>
+  </head>
+  <body class="flex flex-col h-screen antialiased font-medium">
```
tom-james-watson, line 9: I thought we discussed adding antialiased to the default plugin styles?

tom-james-watson (commented):
Not sure what this one means or how I test it

> The colour of the group header icons matches those in the sidebar (as the sidebar where the pages and designs are)

plugins/global-search/index.html
elmarburke: It's in there, but in this project, I didn't import it – mainly for some default styling also applied to the elements that derived from the designed styles

elmarburke:
Thanks @tom-james-watson for taking a look!

> Not sure what this one means or how I test it

This is the icon on the collapsible row; it should have the same colour as in the pages panel

tom-james-watson:
If you mean these, I'm not aware of them ever being anything other than grey
image

tom-james-watson:
Ahhh ok, I understood it as meaning like that the red ones should be red or something 😅

tom-james-watson (approved):

PR #462, opened 2025-09-30
Global search: Group Header Height

elmarburke:
### Description

- Adjust the height of the headers to the height of the sidebar.

image

### Testing

- [ ] Collapsed headers have the same height as in the sidebar


plugins/global-search/src/components/GroupHeader.tsx
```diff
@@ -46,12 +46,12 @@ export const GroupHeader = memo(
                 {...props}
                 {...getFocusProps(headerId(entry.id))}
             >
-                <div className="bg-modal-light dark:bg-modal-dark mt-1">
+                <div className="h-[calc(100%-5px)] mt-1 bg-modal-light dark:bg-modal-dark">
```
triozer, line 49: Why are we doing this? 👀

triozer (approved):
LGTM

plugins/global-search/src/components/GroupHeader.tsx
elmarburke: I needed to have absolute heights in here to make sure it's never overflowing. And instead of use the exact pixel value, I made this dynamic on the surrounding container (which has a height themself)

Issue #443, opened 2025-09-17
Duplicate slug in google sheets silently errors


tom-james-watson:
https://framer-team.slack.com/archives/C01B14R6E22/p1758096153132709

<img width="3938" height="1528" alt="Image" src="https://github.com/user-attachments/assets/1d8e1041-00b4-4c55-9a6e-20e1f3f5084c" />

This ends up just looking like it's syncing forever.

jurrehoutkamp:
happened again

<img width="623" height="66" alt="Image" src="https://github.com/user-attachments/assets/a407c266-607a-432a-9555-7b0c865a4c0d" />

Issue #440, opened 2025-09-15
Google sheets column without a name leads to an unclear error


tom-james-watson:
<img width="1180" height="1454" alt="Image" src="https://github.com/user-attachments/assets/dbf8ef4e-54d2-48ee-9190-34a2ba143e42" />

https://framer-team.slack.com/archives/C01B14R6E22/p1757767118763659

Apparently was caused by a column not having a name.

Issue #302, opened 2025-07-14
Sheets: when importing Date field, we should try to parse them as JS dates


niekert:
**Describe the bug**
Currently we expect the Date to be a number, and we ignore if it's anything else. We should try to parse a date if it's a string.

**To Reproduce**
1. Create a new sheet
2. Add a column name in the first row
3. Add a date `30-04-2025` in the second row
4. Open Sheets Plugin
5. Set the field type to date
6. Synchronize

**Expected behavior**
The date is correct despite EU locale formatting

**Actual behavior**
The date is incorrect. The workaround is to format all cells as dates

PR #468, opened 2025-10-01
Google Sheets bug fixes

madebyisaacr:
### Description

This pull request fixes multiple bugs in the Google Sheets plugin.

* Fixed empty header cells causing an error. Now columns with an empty header cell are silently ignored.
Closes https://github.com/framer/plugins/issues/440
* If items have duplicate slugs, instead of silently ignoring it, we now show an error toast with a list of duplicate slugs.
Closes https://github.com/framer/plugins/issues/443
* Added support for importing dates in ISO format.
Closes https://github.com/framer/plugins/issues/302
* Deleted items are now removed before new items are added, eliminating some errors involving duplicate slugs or duplicate IDs.
Closes https://github.com/framer/company/issues/33639
* Disabled columns can now be used as the slug field. Before this change, if a column was disabled, it would no longer be available for use as the slug.
* For cells with an invalid value/no value, it now uses the default value for that field type instead of skipping the field.

I know we normally try to keep each change in a separate PR, but some of the changes would have merge conflicts or rely on each other, and I didn't want to worry about rebasing and managing multiple reviews.

### Testing

- [x] Sync a sheet with columns with empty header cells, at the beginning, middle, and end of the sheet columns
- [x] Sync a sheet with duplicate slugs
- [x] Sync a sheet with dates in ISO format
- [x] Disable a column, select it as the slug field, and sync your sheet
- [x] Sync a sheet with invalid values, such as text in a number field and observe that it gets set to the default value (`0` for number fields)
- [x] Sync a sheet, then select a different slug field, and re-sync.

---

> [!NOTE]
> Ignores empty-header columns, supports ISO date parsing, ensures unique slugs with numeric suffixes, allows slug from disabled columns, applies default values for invalid cells, and deletes removed items before adding new ones.
> 
> - **Sheets processing/sync**:
>   - Ignore columns with empty headers; `colFieldTypes` and processing accept `null` for such columns.
>   - Support ISO date strings in addition to serial numbers; validate via regex.
>   - Ensure unique slugs by appending numeric suffixes and updating item IDs.
>   - Apply default values for invalid/missing cells per field type instead of skipping.
>   - Allow slug extraction even if the column is ignored; process up to the longest row (`maxRowLength`).
>   - Compute and remove deleted items before adding new ones; set item order only when items exist.
>   - Expand `CellValue` to include `null` and refine boolean handling.
> - **UI/Field mapping**:
>   - Preserve column positions with `null` entries for empty headers; filter with `isDefined` for rendering.
>   - Permit selecting slug from disabled columns; simplify slug field derivation.
>   - Send `colFieldTypes` as `(type|null)` and handle name/type overrides safely.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a37207fd706399793d09cb0027e87a1b9d352b96. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/plugins/google-sheets/src/sheets.ts
+++ b/plugins/google-sheets/src/sheets.ts
@@ -254,7 +254,7 @@ interface ProcessSheetRowParams {
     fieldTypes: CollectionFieldType[]
     row: Row
     rowIndex: number
-    rowLength: number
+    columnCount: number
     uniqueHeaderRowNames: string[]
     slugFieldColumnIndex: number
     ignoredFieldColumnIndexes: number[]
@@ -326,7 +326,7 @@ function getFieldDataEntryInput(type: CollectionFieldType, cellValue: CellValue)
             return isDefined(cellValue) ? { type, value: CELL_BOOLEAN_VALUES.includes(cellValue) } : null
         }
         case "date": {
-            // Google Sheets numeric date values
+            // Google Sheets numeric date values (Lotus 1-2-3 format)
             if (typeof cellValue === "number") {
                 try {
                     const date = extractDateFromSerialNumber(cellValue)
@@ -337,11 +337,8 @@ function getFieldDataEntryInput(type: CollectionFieldType, cellValue: CellValue)
             }
 
             // ISO date format
-            if (typeof cellValue === "string") {
-                if (isValidISODate(cellValue)) {
-                    return { type, value: cellValue }
-                }
-                return null
+            if (typeof cellValue === "string" && isValidISODate(cellValue)) {
+                return { type, value: cellValue }
             }
 
             return null
@@ -363,7 +360,7 @@ function getFieldDataEntryInput(type: CollectionFieldType, cellValue: CellValue)
 function processSheetRow({
     row,
     rowIndex,
-    rowLength,
+    columnCount,
     uniqueHeaderRowNames,
     ignoredFieldColumnIndexes,
     slugFieldColumnIndex,
@@ -374,7 +371,7 @@ function processSheetRow({
     let slugValue: string | null = null
     let itemId: string | null = null
 
-    for (let i = 0; i < rowLength; i++) {
+    for (let i = 0; i < columnCount; i++) {
         const cell = row[i] ?? null
 
         const fieldType = fieldTypes[i]
@@ -484,19 +481,17 @@ function processSheet(rows: Row[], processRowParams: Omit<ProcessSheetRowParams,
     )
     const collectionItems = result.filter(isDefined)
 
-    // Detect duplicate slugs and report error if any are found
-    const slugCounts = new Map<string, number>()
+    // Find duplicate slugs and report error if any are found
+    const seenSlugs = new Set<string>()
     const duplicateSlugs = new Set<string>()
 
-    // Count occurrences of each slug
-    collectionItems.forEach(item => {
-        const count = slugCounts.get(item.slug) ?? 0
-        slugCounts.set(item.slug, count + 1)
-
-        if (count > 0) {
+    for (const item of collectionItems) {
+        if (seenSlugs.has(item.slug)) {
             duplicateSlugs.add(item.slug)
+        } else {
+            seenSlugs.add(item.slug)
         }
-    })
+    }
 
     if (duplicateSlugs.size > 0) {
         const slugList = formatListWithAnd(Array.from(duplicateSlugs))
@@ -504,10 +499,8 @@ function processSheet(rows: Row[], processRowParams: Omit<ProcessSheetRowParams,
         throw new Error(`Duplicate slug${pluralSuffix} found: ${slugList}. Each item must have a unique slug.`)
     }
 
-    const processedItems = collectionItems
-
     return {
-        collectionItems: processedItems,
+        collectionItems,
         status,
     }
 }
@@ -569,7 +562,7 @@ export async function syncSheet({
         fieldTypes: colFieldTypes,
         ignoredFieldColumnIndexes: ignoredColumns.map(col => uniqueHeaderRowNames.indexOf(col)),
         slugFieldColumnIndex: slugColumn ? uniqueHeaderRowNames.indexOf(slugColumn) : -1,
-        rowLength: headerRow.length,
+        columnCount: headerRow.length,
     })
 
     // Calculate items to delete based on what's in the collection vs what we processed


plugins/google-sheets/src/sheets.ts
```diff
@@ -409,8 +477,36 @@ function processSheet(
...
+
+        usedSlugs.add(uniqueSlug)
+
+        // If slug changed, also update the item ID because it's based on the slug
+        if (uniqueSlug !== item.slug) {
+            return {
+                ...item,
+                slug: uniqueSlug,
+                id: generateHashId(uniqueSlug),
+            }
```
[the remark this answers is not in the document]
madebyisaacr: The ID can't be based on the original field value, because if there is a duplicate slug, it means multiple items have the same slug and therefore the same ID because ID is a hash of the slug value.

It won't break ID stability between each sync unless the slug changes, which is the same as how it works for items without duplicate slugs.

plugins/google-sheets/src/sheets.ts
```diff
@@ -25,6 +25,7 @@ const PLUGIN_SLUG_COLUMN_KEY = "sheetsPluginSlugColumn"
 
 const CELL_BOOLEAN_VALUES = ["Y", "yes", "true", "TRUE", "Yes", 1, true]
 const HEADER_ROW_DELIMITER = "OIhpKTpp"
+const ISO_DATE_REGEX = /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(\.\d{3})?Z?)?$/
```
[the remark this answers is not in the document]
madebyisaacr: Fixed

plugins/google-sheets/src/sheets.ts
```diff
@@ -298,24 +323,36 @@ function getFieldDataEntryInput(type: CollectionFieldType, cellValue: CellValue)
...
-            return { type, value: CELL_BOOLEAN_VALUES.includes(cellValue) }
+            return isDefined(cellValue) ? { type, value: CELL_BOOLEAN_VALUES.includes(cellValue) } : null
         }
         case "date": {
-            if (typeof cellValue !== "number") return null
-            try {
-                const date = extractDateFromSerialNumber(cellValue)
-                return { type, value: date.toISOString() }
-            } catch {
+            // Google Sheets numeric date values
```
elmarburke: nit: according to the extraction function, this is "Lotus 1-2-3 format" (wow, really wow, some software has really a lot of legacy to carry around)

plugins/google-sheets/src/sheets.ts
```diff
@@ -298,24 +323,36 @@ function getFieldDataEntryInput(type: CollectionFieldType, cellValue: CellValue)
...
+                    const date = extractDateFromSerialNumber(cellValue)
+                    return { type, value: date.toISOString() }
+                } catch {
+                    return null
+                }
+            }
+
+            // ISO date format
+            if (typeof cellValue === "string") {
+                if (isValidISODate(cellValue)) {
```
elmarburke:
```suggestion
            if (typeof cellValue === "string" && isValidISODate(cellValue)) {
```

plugins/google-sheets/src/sheets.ts
```diff
@@ -326,7 +363,7 @@ function getFieldDataEntryInput(type: CollectionFieldType, cellValue: CellValue)
 function processSheetRow({
     row,
     rowIndex,
-    unsyncedRowIds,
+    rowLength,
```
elmarburke: nit: I think the naming could be clearer.
```suggestion
    rowCount,
```

plugins/google-sheets/src/sheets.ts
```diff
@@ -409,8 +484,30 @@ function processSheet(
...
+        if (count > 0) {
+            duplicateSlugs.add(item.slug)
+        }
+    })
+
+    if (duplicateSlugs.size > 0) {
+        const slugList = formatListWithAnd(Array.from(duplicateSlugs))
+        const pluralSuffix = duplicateSlugs.size > 1 ? "s" : ""
+        throw new Error(`Duplicate slug${pluralSuffix} found: ${slugList}. Each item must have a unique slug.`)
+    }
```
elmarburke: Consider (no must) refactoring this to have an array of all slugs you can filter them. I think the performance win you get from the lookup isn't worth the code.

```ts
const duplicateSlugs = collectionItems.map(item => item.slug).filter((item, index) => array.indexOf(item) !== index)
```

plugins/google-sheets/src/sheets.ts
```diff
@@ -409,8 +484,30 @@ function processSheet(
...
+
+    const processedItems = collectionItems
+
     return {
-        collectionItems,
+        collectionItems: processedItems,
```
elmarburke:
```suggestion
    return {
        collectionItems,
```

elmarburke (approved):

elmarburke (approved):
Code LGTM

QA: The points described passed, but I've noticed:

- Duplicate Slug
  -  first run, works, shows toast ✅ 
  - removed the line in the sheet, kept the dialog open: still the same error
- Default Values
  - Works as described
  - Maybe we can add a warning in a toast? I can imagine getting other values that you expected could be bad w/o knowing.

plugins/google-sheets/src/sheets.ts
madebyisaacr: `rowCount` sounds like the total number of rows. Changed it to `columnCount` instead.

plugins/google-sheets/src/sheets.ts
madebyisaacr: Not sure an array filter is the best way to do it, but I have updated the code with slightly simpler logic that still uses maps.

plugins/google-sheets/src/sheets.ts
madebyisaacr: Done

plugins/google-sheets/src/sheets.ts
```diff
@@ -484,30 +481,26 @@ function processSheet(rows: Row[], processRowParams: Omit<ProcessSheetRowParams,
...
-
-        if (count > 0) {
+    for (const item of collectionItems) {
+        if (seenSlugs.has(item.slug)) {
             duplicateSlugs.add(item.slug)
+        } else {
+            seenSlugs.add(item.slug)
         }
-    })
+    }
```
elmarburke, line 494: Muuuuuuch nicer! Thanks 👍

Issue #465, opened 2025-09-30
Show spinner on Log In button while waiting for log in result in Airtable plugin


madebyisaacr:
In the Airtable plugin, after clicking the Log In button, it should show a loading spinner on the button. Right now it keeps the log in text. 

<img width="333" height="402" alt="Image" src="https://github.com/user-attachments/assets/e28fc5eb-0105-49aa-ab45-9246ccc732dd" />

Other plugins such as Notion and Sheets already do this.

<img width="330" height="401" alt="Image" src="https://github.com/user-attachments/assets/851f75ea-a92b-4859-a9b7-228b5b6612e8" />

PR #470, opened 2025-10-01
Add loading spinner to Airtable log in button

madebyisaacr:
### Description

This pull request adds a loading state to the log in button in Airtable to provide instant feedback to users and for consistency with other plugins.

Closes https://github.com/framer/plugins/issues/465

image

### Testing

- [ ] Click the log in button in Airtable

### DIFF
--- a/plugins/airtable/src/Login.tsx
+++ b/plugins/airtable/src/Login.tsx
@@ -1,7 +1,7 @@
 import "./App.css"
 
 import { framer } from "framer-plugin"
-import { useLayoutEffect, useRef } from "react"
+import { useLayoutEffect, useRef, useState } from "react"
 import auth from "./auth"
 import { showLoginUI } from "./ui"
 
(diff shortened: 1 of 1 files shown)


plugins/airtable/src/Login.tsx
```diff
@@ -68,7 +73,7 @@ export function Authenticate({ onAuthenticated }: AuthenticationProps) {
             </ol>
 
             <button className="action-button" type="submit">
```
elmarburke: Consider adding a `disabled={isLoading}` to also prevent multiple actions at once, also, consider adding
```suggestion
            <button className="action-button" type="submit"
                aria-busy={isLoading}
                aria-disabled={isLoading}
                aria-label={isLoading ? "Logging in" : "Log in"}
            >
```

madebyisaacr: It shouldn't be disabled because if you close the Airtable log in tab before finishing, it will still show as loading in the plugin but you need to click it again to try logging in again. If it's disabled, you'd have to close the plugin and reopen it to get back to the Airtable log in screen.

Added `aria-label` to the button.

PR #472, opened 2025-10-03
Crowdin Plugin

clementroche:
Crowdin Plugin: Import and Export Localization from Crowdin.


plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+    })
+
+    // ------------------ Import from Crowdin ------------------
+    async function importFromCrowdIn() {
+        if (!accessToken || !projectId || !activeLocale) {
+            framer.notify("Access Token, Project ID, or active locale missing", {
+                variant: "error",
+            })
+            return
+        }
```
madebyisaacr: The error message can be more specific.

```suggestion
        if (!accessToken) {
            framer.notify("Access Token is missing", {
                variant: "error",
            })
            return
        } else if (!projectId) {
            framer.notify("Project ID is missing", {
                variant: "error",
            })
            return
        } else if (!activeLocale) {
            framer.notify("Active locale is missing", {
                variant: "error",
            })
            return
        }
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+            const { xliff, targetLocale } = parseXliff(fileContent, locales)
+            const valuesBySource = createValuesBySourceFromXliff(xliff, targetLocale)
+
+            const result = await framer.setLocalizationData({ valuesBySource })
+
+            if (result.valuesBySource.errors.length > 0) {
+                throw new Error(`Import errors: ${result.valuesBySource.errors.map(error => error.error).join(", ")}`)
+            }
+
+            framer.notify(`Successfully imported localizations for ${targetLocale.name}`)
```
madebyisaacr:
```suggestion
            framer.notify(`Successfully imported localizations for ${targetLocale.name}`, { variant: "success" })
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+            setIsLoading(false)
+        }
+    }
+    async function exportToCrowdIn() {
+        if (!accessToken || !projectId || !activeLocale) {
+            framer.notify("Access Token, Project ID, or active locale missing", {
+                variant: "error",
+            })
+            return
+        }
```
madebyisaacr: The error message can be more specific.

```suggestion
        if (!accessToken) {
            framer.notify("Access Token is missing", {
                variant: "error",
            })
            return
        } else if (!projectId) {
            framer.notify("Project ID is missing", {
                variant: "error",
            })
            return
        } else if (!activeLocale) {
            framer.notify("Active locale is missing", {
                variant: "error",
            })
            return
        }
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+                throw new Error(`Import errors: ${result.valuesBySource.errors.map(error => error.error).join(", ")}`)
+            }
+
+            framer.notify(`Successfully imported localizations for ${targetLocale.name}`)
+        } catch (err) {
+            console.error("Error importing from Crowdin:", err)
+            framer.notify(
+                "Error importing from Crowdin Could be because translation missing in CrowdIn. Please check the source",
+                { variant: "error" }
+            )
```
madebyisaacr, line 166: Other plugins use this pattern to provide a more specific error message if possible.

```suggestion
            framer.notify(
                `Error importing from Crowdin: ${error instanceof Error ? error.message : "An unknown error occurred"}`,
                { variant: "error" }
            )
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+            if (uploadRes && !uploadRes.ok) {
+                const errMsg = await uploadRes.text()
+                framer.notify(`Crowdin upload failed: ${errMsg}`, { variant: "error" })
+                return
+            }
+
+            framer.notify("Export to Crowdin complete", { variant: "success" })
+        } catch (err) {
+            console.error("Error exporting to Crowdin:", err)
+            framer.notify("Error exporting to Crowdin", { variant: "error" })
```
madebyisaacr:
```suggestion
            framer.notify(`Error exporting to Crowdin: ${error instanceof Error ? error.message : "An unknown error occurred"}`, { variant: "error" })
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+                setAccessToken(storedToken)
+                validateAccessToken(storedToken)
+            }
+        }
+        void loadStoredToken()
+    }, [validateAccessToken])
+
+    // const crowdinClient = new crowdin({
+    //     token: accessToken,
+    // })
```
madebyisaacr:
```suggestion
```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+}
+.crowdin-button-primary {
+    color: var(--framer-color-text-reversed);
+    background-color: #263238;
+}
+.crowdin-button-primary:hover,
+.crowdin-button-primary:focus {
+    background-color: var(--framer-color-text);
+    color: #222222;
+}
```
madebyisaacr: Make the hover and focus states not apply if the button is disabled.

https://github.com/user-attachments/assets/6b229737-6b4b-481b-9b92-60d821be6dbb

```suggestion
.crowdin-button-primary:not(:disabled):hover,
.crowdin-button-primary:not(:disabled):focus {
    background-color: var(--framer-color-text);
    color: #222222;
}
```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+
+.setup select {
+    width: 100%;
+}
+
+.setup input,
+.setup select {
+    width: 134px;
+}
```
madebyisaacr: These styles conflict:

```suggestion
.setup input,
.setup select {
    width: 134px;
}
```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+    gap: 8px;
+    justify-content: space-between;
+}
+.button-stack button {
+    flex: 1;
+}
+.crowdin-button-primary {
+    color: var(--framer-color-text-reversed);
+    background-color: #263238;
+}
```
madebyisaacr: Combine duplicate .setup rules and add an empty line between each rule.

```suggestion
.setup {
    width: 100%;
    display: flex;
    flex-direction: column;
    align-items: start;
    gap: 15px;
    width: 100%;
    padding-bottom: 15px;
}

a {
    color: currentColor;
}

.button-stack {
    display: flex;
    gap: 8px;
    justify-content: space-between;
}

.button-stack button {
    flex: 1;
}

.crowdin-button-primary {
    color: var(--framer-color-text-reversed);
    background-color: #263238;
}

```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+
+.setup button {
+    margin-top: 5px;
+}
+
+.setup img {
+    user-select: none;
+    -webkit-user-select: none;
+    pointer-events: none;
+}
```
madebyisaacr, line 71: Combine img styles:

```suggestion
.setup img {
    height: 180px;
    width: 100%;
    border-radius: 10px;
    background-color: #263238;
    object-fit: contain;
    user-select: none;
    -webkit-user-select: none;
    pointer-events: none;
}

.setup button {
    margin-top: 5px;
}
```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+    flex-direction: column;
+    align-items: start;
+    padding: 0 15px 0;
+    height: 100%;
+    gap: 15px;
+
+    user-select: none;
+    -webkit-user-select: none;
+}
+.form-field {
```
madebyisaacr, line 27: Add empty line.

```suggestion
}

.form-field {
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+                {isLoading && (
+                    <div className="loader">
+                        <Loading />
+                    </div>
+                )}
+                <label className="show">
+                    <p>Access Token</p>
+                    <input
+                        type="text"
+                        placeholder="Enter Access Token…"
```
madebyisaacr: Fix placeholder text getting cut off.

Before:
image

After:
image

```suggestion
                        placeholder="Enter Token…"
```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+    border-radius: 8px;
+}
+
+.setup {
+    align-items: center;
+}
+
+.setup select {
+    font-weight: 600;
+}
```
madebyisaacr: Looks weird with the select having a bolder weight than the input right above it. The default weight is fine.

```suggestion
```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,127 @@
...
+
+.logo {
+    color: var(--framer-color-text-tertiary);
+    background-color: var(--framer-color-bg-tertiary);
+    border-radius: 8px;
+}
+
+.setup {
+    align-items: center;
+}
```
madebyisaacr: Conflicts with the .setup style later down in this file.

```suggestion
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+            framer.notify("Error exporting to Crowdin", { variant: "error" })
+        } finally {
+            setIsLoading(false)
+        }
+    }
+
+    return (
+        <main className="framer-hide-scrollbar setup">
+            <img src={hero} alt="Crowdin Hero" />
+            <div className="form-field" style={{ position: "relative" }}>
```
madebyisaacr: The CSS for .form-field already has position: relative

```suggestion
            <div className="form-field">
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+                <label className="show">
+                    <p>Project</p>
+                    <select
+                        value={projectId || ""}
+                        onChange={e => {
+                            setProjectId(Number(e.target.value))
+                        }}
+                        disabled={!accessToken}
+                    >
+                        <option value="">Choose Project…</option>
```
madebyisaacr: Make it a disabled option. It will still be selected by default.

```suggestion
                        <option value="" disabled>Choose Project…</option>
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+                    textDirection: "ltr",
+                    pluralCategoryNames: ["one", "other"],
+                    threeLettersCode: language.name.substring(0, 3),
+                    twoLettersCode: language.slug,
+                    dialectOf: language.slug,
+                }),
+            })
+
+            framer.notify(
+                `Language ${language.code} is not present in CrowdIn. Please check region and language code in Framer`,
```
madebyisaacr:
```suggestion
                `Language ${language.code} is not present in Crowdin. Please check your locale's region and language code in Framer`,
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+        const storageList = await fetch(`${API_URL}/storages`, {
+            headers: { Authorization: `Bearer ${accessToken}` },
+        })
+
+        const storageData = (await storageList.json()) as unknown
+
+        // Validate response with valibot
+        const parsed = v.safeParse(v.object({ data: v.array(StoragesSchema) }), storageData)
+        if (!parsed.success) {
+            console.error("Error parsing CrowdIn storages:", parsed.issues)
```
madebyisaacr:
```suggestion
            console.error("Error parsing Crowdin storages:", parsed.issues)
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+// -------------------- Get or Create File --------------------
+export async function getFileId(projectId: number, fileName: string, accessToken: string): Promise<number | Response> {
+    try {
+        const filesRes = await fetch(`${API_URL}/projects/${projectId}/files`, {
+            headers: { Authorization: `Bearer ${accessToken}` },
+        })
+        const filesData = (await filesRes.json()) as unknown
+        const parsed = v.safeParse(v.object({ data: v.array(ProjectsSchema) }), filesData)
+        if (!parsed.success) {
+            console.error("Error parsing CrowdIn files:", parsed.issues)
```
madebyisaacr:
```suggestion
            console.error("Error parsing Crowdin files:", parsed.issues)
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+//             return "new"
+//         case "translated":
+//             return "needsReview"
+//         case "reviewed":
+//         case "final":
+//             return "done"
+//         default:
+//             state satisfies never
+//     }
+// }
```
madebyisaacr: If this function isn't going to be used, it can be removed. Unless it may be useful in the future, then you can ignore this.

```suggestion
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+
+    const targetLanguage = xliffElement.getAttribute("target-language")
+    if (!targetLanguage) throw new Error("No target language found in XLIFF")
+
+    const targetLocale = locales.find(locale => locale.code === targetLanguage)
+    if (!targetLocale) {
+        throw new Error(`No locale found for language code: ${targetLanguage}`)
+    }
+
+    return { xliff: xliff, targetLocale }
```
madebyisaacr:
```suggestion
    return { xliff, targetLocale }
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+
+    const units = xliffDocument.querySelectorAll("trans-unit")
+    for (const unit of units) {
+        const id = unit.getAttribute("resname")
+        const target = unit.querySelector("target")
+        if (!id || !target) continue
+        const targetValue = target.textContent
+        // const state = target.getAttribute("state") as XliffState | null
+        // const status = xliffStateToStatus(state ?? "final")
+        // const needsReview = status === "needsReview"
```
madebyisaacr: Same as previous comment, if this may be useful later then it can be kept, otherwise remove it.

```suggestion
```

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,519 @@
...
+
+        // Ignore missing or empty values
+        if (!targetValue) continue
+
+        valuesBySource[id] = {
+            [targetLocale.id]: {
+                action: "set",
+                value: targetValue,
+                // needsReview: needsReview ? needsReview : undefined,
```
madebyisaacr:
```suggestion
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,248 @@
...
+            // persist token
+            if (framer.isAllowedTo("setPluginData")) {
+                void framer.setPluginData("accessToken", token)
+            }
+            const projectsGroupsApi = new ProjectsGroups({ token })
+            projectsGroupsApi
+                .withFetchAll()
+                .listProjects()
+                .then(response => {
+                    console.log(response.data)
```
madebyisaacr: Only log data while testing.

```suggestion
                    if (window.location.hostname === "localhost") {
                        console.log(response.data);
                    }
```

madebyisaacr (changes requested):
Looks pretty good. Mostly commented on error messages, styles and code details.

---

The import and export buttons are disabled after I entered an API key (with all scopes enabled) and I have no idea why.

image

There's no notification or messages in the console aside from this log of the data:
image

If something is wrong with my access token or Crowdin account, it should have an error notification or some way of communicating why the buttons are still disabled.

---

Add a link to the settings page where you can make an API access token: https://crowdin.com/settings#api-key

---

There should be something that tells you which scopes are necessary to use the plugin. Someone could make a doc in [localization support](https://www.framer.com/help/localization/) with this info and add a link to it inside the plugin. But you guys probably don't need to do anything here - the people who manage documentation can do this.

image

---

Does it work with both file based and string based projects? It asks you to pick one when you sign up and create a new project:

image

clementroche:
Thank you @madebyisaacr!

The project should work with both file-based and string based actually. It seems it only changes the interface.

plugins/crowdin/src/main.tsx
```diff
@@ -0,0 +1,28 @@
...
+const locales = await framer.getLocales()
+
+const isAllowedToSetLocalizationData = framer.isAllowedTo("setLocalizationData")
+
+if (!isAllowedToSetLocalizationData) {
+    framer.notify("You are not allowed to set localization data", { variant: "error" })
+    framer.closePlugin()
+} else if (!activeLocale) {
+    framer.notify("No active locale found", { variant: "error" })
+    framer.closePlugin()
```
madebyisaacr: Combine to avoid two notifications.

Image

```suggestion
    framer.closePlugin("No active locale found. Please select a locale.", { variant: "error" })
```

plugins/crowdin/src/main.tsx
```diff
@@ -0,0 +1,28 @@
...
+if (!root) throw new Error("Root element not found")
+
+const activeLocale = await framer.getActiveLocale()
+const locales = await framer.getLocales()
+
+const isAllowedToSetLocalizationData = framer.isAllowedTo("setLocalizationData")
+
+if (!isAllowedToSetLocalizationData) {
+    framer.notify("You are not allowed to set localization data", { variant: "error" })
+    framer.closePlugin()
```
madebyisaacr: The plugin should open if the user doesn't have permission to set localization data, because you can also export data which is allowed when you only have view permissions.

For example, see how the locale sync and CSV import plugins work, where they can still be opened if you don't have permissions but the import button is disabled.

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,259 @@
...
+                </label>
+                <div className="button-stack">
+                    <button
+                        type="button"
+                        onClick={() => {
+                            void importFromCrowdIn()
+                        }}
+                    >
+                        Import
+                    </button>
```
madebyisaacr, line 284: Disable this button if `setLocalizationData` permissions are not allowed.

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,108 @@
+/* Your Plugin CSS */
+
```
madebyisaacr, line 2:
```suggestion

[data-framer-theme="light"] #root {
    --framer-color-tint: #263238;
    --framer-color-tint-dark: #33424a;
    --framer-color-tint-extra-dark: #3d4e57;
}

[data-framer-theme="dark"] #root {
    --framer-color-tint: #fff;
    --framer-color-tint-dark: #e3e3e3;
    --framer-color-tint-extra-dark: #c3c3c3;
    --framer-color-text-reversed: #000;
}

```

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,108 @@
...
+}
+.crowdin-button-primary {
+    color: var(--framer-color-text-reversed);
+    background-color: #263238;
+}
+.crowdin-button-primary:not(:disabled):hover,
+.crowdin-button-primary:not(:disabled):focus {
+    background-color: var(--framer-color-text);
+    color: #222222;
+}
```
madebyisaacr:
```suggestion
```

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+                    <button
+                        type="button"
+                        className="crowdin-button-primary"
```
madebyisaacr:
```suggestion
                        className="framer-button-primary"
```

madebyisaacr (changes requested):
When there's an export error, it still shows the success notification. The success notification should not be triggered if there's an error.

Image

The error seems to be wrong though, because my Crowdin project has an "es" locale and it actually did export the localization data to Crowdin. It created a "framer-source-en-US.xliff" file in the es locale that matches the Framer localization data.

Image

Image

It makes two API calls to the same endpoint "https://api.crowdin.com/api/v2/languages" and the second one fails (maybe because it doesn't have the limit param?) And before that it has successful requests to get files and create a file.

Image

madebyisaacr:
Why does it send a PATCH request to "https://api.crowdin.com/api/v2/projects/835172" that attempts to replace the list of locales in Crowdin with the same locales that are already there? This request also fails (404 error).

The locales list here is from Crowdin, not Framer, because my Framer project only has 1 locale.

image

plugins/crowdin/src/main.tsx
clementroche: indeed!

clementroche:
@madebyisaacr, i've improved the error management. 
The issue you noticed comes from the fact that Framer uses ISO 639-1 while Crowdin uses their own Language system. For instance the Framer Locale "es" (without region) isn't a valid Crowdin language. There is a way to create custom language in Crowdin but it requires informations that Framer doesn't provide such as `textDirection`, `threeLettersCode` etc. (cf: [documentation](https://support.crowdin.com/developer/api/v2/#tag/Languages/operation/api.languages.post)). The best way remains to notify the user that this Locale can't be exported since there is a misconfiguration. That way the user will have to manually create the custom language or make it match with Crowdin.

plugins/crowdin/src/xliff.ts
```diff
@@ -0,0 +1,476 @@
...
+            [targetLocale.id]: {
+                action: "set",
+                value: targetValue,
+                needsReview: false,
+            },
+        }
+    }
+
+    return valuesBySource
+}
```
kaloyanvi, line 70: Seems like the XLIFF parser pulls ids from `resname`, but the exporter (and Crowdin) use the standard `id` attribute, so `valuesBySource` is always empty and nothing is imported.

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,263 @@
...
+// ----- App component -----
+export function App({ activeLocale, locales }: { activeLocale: Locale | null; locales: readonly Locale[] }) {
+    const isAllowedToSetLocalizationData = useIsAllowedTo("setLocalizationData")
+
+    const [accessToken, setAccessToken] = useState<string>("")
+    const [projectList, setProjectList] = useState<readonly Project[]>([])
+    const [projectId, setProjectId] = useState<number>(0)
+    const [isLoading, setIsLoading] = useState(false)
+
+    const validateAccessToken = useCallback((token: string) => {
```
kaloyanvi, line 57: We are validating on every character currently, so if the user is writing the token instead of pasting, user will see the spinner like 20 times.

kaloyanvi: With the current UX it probably makes most sense to hook up the input (token) to a `useDeferredValue` and then use that value for the validation in the `useEffect`. Alternative would be to do it on input blur which might be nicer and more predictable. What do you think?

plugins/crowdin/src/components/Loading.tsx
```diff
@@ -0,0 +1,7 @@
+export function Loading() {
+    return (
+        <main className="loading">
```
kaloyanvi: We shouldn't be using two `<main>` per HTML document.

plugins/crowdin/src/App.css
```diff
@@ -0,0 +1,111 @@
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: var(--framer-color-bg);
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    z-Index: 10;
```
kaloyanvi, line 45: whats the reason for showing the loader as an overlay? This means the user can still type while its loading?

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,263 @@
...
+void framer.showUI({ width: 325, height: 325 })
+
+interface Project {
+    readonly id: number
+    readonly name: string
+}
+
+interface CrowdinStorageResponse {
+    data: {
+        id: string
```
kaloyanvi: In the response this is a number and not a string

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,263 @@
...
+    }, [validateAccessToken])
+
+    const createCrowdinClient = (token: string) => ({
+        projects: new ProjectsGroups({ token }),
+        translations: new Translations({ token }),
+    })
+
+    // ------------------ Import from Crowdin ------------------
+    async function importFromCrowdIn() {
+        if (!accessToken || !projectId || !activeLocale) {
```
kaloyanvi: We are disabling the button based on locale permissions but we should also check for `isAllowedToSetLocalizationData` as well here just like we are doing with `accessToken` and `projectId` in both places.

plugins/crowdin/src/App.tsx
```diff
@@ -0,0 +1,263 @@
...
+                    />
+                </label>
+                <label className="show">
+                    <p>Project</p>
+                    <select
+                        value={projectId || ""}
+                        onChange={e => {
+                            setProjectId(Number(e.target.value))
+                        }}
+                        disabled={!accessToken}
```
kaloyanvi: Should this be disabled if the token is not valid as well?

kaloyanvi: Even better if we dont show at all the project field until the token has been validated. @andreugordillovazquez wdyt?

plugins/crowdin/src/xliff.ts
kaloyanvi, line 1: Can we add return types to all functions which are exported?

kaloyanvi (changes requested):
Left some comments.

plugins/crowdin/src/App.tsx
clementroche: I've opted for debounced validation

plugins/crowdin/src/xliff.ts
clementroche: It's correctly imported on my side. Or maybe i'm missing something?

plugins/crowdin/src/App.tsx
clementroche: i've disabled the dropdown if no projects are available

plugins/crowdin/src/App.css
clementroche: The input is blurred when loader is visible.

PR #473, opened 2025-10-05
Improve duplicate slug errors in Notion and Airtable

madebyisaacr:
### Description

This pull request adds descriptive error messages for duplicate slugs in Notion and Airtable, and improves other error messages.

Closes https://github.com/framer/plugins/issues/467
Closes https://github.com/framer/plugins/issues/466

Before:
Pasted Graphic

After:
image

The code in this PR was copied from https://github.com/framer/plugins/pull/468

### Testing

- [ ] Sync a Notion database with duplicate slugs. [Testing database](https://www.notion.so/framer/1efadf6e8c9680888be6fa350bc0f369?v=1efadf6e8c968018b2b6000cffaa93ca)
- [ ] Sync an Airtable base with duplicate slugs

PR #479, opened 2025-10-19
Fixed duplicate slug issue on import

bengold:
### Description

Fixes issue where import would fail due to duplicate slugs being created

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -82,6 +82,58 @@ export function mergeFieldsInfoWithExistingFields(
     })
 }
 
+/**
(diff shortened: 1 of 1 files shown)


triozer:
The Notion plugin is read-only, so we should never update slugs on import. Instead, we should surface a clearer toast indicating that two items share the same slug. We don’t want to create confusion about why updating an item in Notion didn’t update the correct item in the CMS.

@madebyisaacr recently fixed this for ([Google Sheets](https://github.com/framer/plugins/pull/468#top)). We should mirror that behavior in Notion by detecting duplicates and showing an explicit error toast, rather than mutating slugs.

madebyisaacr:
@triozer Notion has the same error that lists the duplicate slugs as Google Sheets - added in https://github.com/framer/plugins/pull/473

PR #482, opened 2025-10-30
Hide unnecessary warnings and errors in Notion

madebyisaacr:
### Description

This pull request hides the "skip item" warnings for each item with no changes except on localhost, and the plugin close error when you sync in the field mapping UI. These can be confusing to users as they are not actual errors.

### Testing

- [x] Sync a collection
- [x] Check the console to see if there are any warnings or errors from the plugin

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -117,7 +117,7 @@ export async function syncCollection(
 
             let isUnchanged = false
             if (isUnchangedSinceLastSync(item.last_edited_time, lastSynced)) {
-                if (import.meta.env.MODE  === "development") {
+                if (import.meta.env.MODE === "development") {
                     console.warn({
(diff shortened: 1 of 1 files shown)


plugins/notion/src/data.ts
```diff
@@ -117,10 +117,12 @@ export async function syncCollection(
 
             let isUnchanged = false
             if (isUnchangedSinceLastSync(item.last_edited_time, lastSynced)) {
-                console.warn({
-                    message: `Skipping content update. last updated: ${formatDate(item.last_edited_time)}, last synced: ${lastSynced ? formatDate(lastSynced) : "never"}`,
-                    url: item.url,
-                })
+                if (location.hostname.includes("localhost")) {
```
elmarburke: nit: instead of checking in the runtime, this could rely on the `import.meta.env.MODE  === "development"` (https://vite.dev/guide/env-and-mode.html#modes)

```suggestion
                if (import.meta.env.MODE  === "development") {
```

elmarburke (approved):

Issue #494, opened 2025-11-25
Support Notion formula fields in slug field dropdown


andreugordillovazquez:
A customer has requested an enhancement to the Framer Notion plugin. The current Slug Field dropdown only supports Notion text fields, which prevents users from selecting Formula fields that compute slugs dynamically.

### Use Case

Some users maintain slugs as derived fields in Notion using formulas. Because the plugin only accepts text fields, they cannot select these formula-based fields as their slug source when syncing to Framer.

### Proposed Solution

Extend the Slug Field dropdown to also support Formula fields from Notion databases. This would allow users to leverage computed values directly, ensuring parity between Notion logic and what Framer uses as slugs.

<img width="1084" height="197" alt="Image" src="https://github.com/user-attachments/assets/f17d153a-cc24-45ba-8a21-1ca5474142d6" />

PR #495, opened 2025-11-26
Notion: allow using formula fields as slug

madebyisaacr:
### Description

This pull request updates the Notion plugin to allow using formula fields as the slug field.

Closes https://github.com/framer/plugins/issues/494

### Testing

- [x] Connect a database with a formula ([this testing database](https://www.notion.so/framer/224adf6e8c9680eca17fda0e00267005?v=224adf6e8c9680899fb5000c4fa462a1) has one)
- [x] In the slug field dropdown, select the formula field
- [x] Sync the database and verify that the item slugs are correct

### DIFF
--- a/plugins/notion/src/api.ts
+++ b/plugins/notion/src/api.ts
@@ -62,7 +62,7 @@ export const pageCoverProperty: FieldInfo = {
 }
 
 // The valid field types that can be used as a slug, in order of preference
-const slugFieldTypes: NotionProperty["type"][] = ["title", "rich_text", "unique_id"]
+const slugFieldTypes: NotionProperty["type"][] = ["title", "rich_text", "unique_id", "formula"]
 
 export const supportedCMSTypeByNotionPropertyType = {
     checkbox: ["boolean"],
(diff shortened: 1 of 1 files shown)


plugins/notion/src/api.ts
```diff
@@ -262,6 +262,25 @@ export function getSlugValue(property: PageObjectResponse["properties"][string])
...
                 ? `${property.unique_id.prefix}-${String(property.unique_id.number)}`
                 : String(property.unique_id.number)
+        case "formula":
+            switch (property.formula.type) {
+                case "string": {
+                    return property.formula.string ?? null
+                }
+                case "number": {
+                    const number = property.formula.number
+                    return typeof number === "number" && !isNaN(number) ? String(number) : null
```
triozer, line 272:
```suggestion
                    return Number.isFinite(number) ? String(number) : null
```

madebyisaacr: Opened a PR for this change: https://github.com/framer/plugins/pull/497

PR #502, opened 2025-12-09
Fix markdown in Airtable

madebyisaacr:
### Description

This pull request reverts most of the changes from https://github.com/framer/plugins/pull/500. Markdown parsing is different in Framer studio. It removes HTML tags from markdown parsed content, making it impossible to import `` tags, HTML tables, etc. so I moved markdown parsing back to the plugin with `marked` to allow importing both markdown and HTML content mixed together.

While testing, I noticed that it was sending a lot of unnecessary Airtable-specific data in `collection.setFields()`, so I added a `processFields` function which converts the fields list to only include Framer API compatible data.

### Testing

- [x] Import an Airtable base with a rich text field with both markdown and HTML content
- Here's a base to test with: https://airtable.com/appTZWyotYJMt26Xt/shrJ0RUFU1stomNi5. There's a "Copy base" button at the top.

### DIFF
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -310,7 +310,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
                             field.collectionId !== ""
                     )
 
-                await collection.setFields(extractFields(fieldsToSync))
+                await collection.setFields(processFields(fieldsToSync))
                 await syncCollection(collection, dataSource, fieldsToSync, selectedSlugField.id)
                 framer.closePlugin("Synchronization successful", {
                     variant: "success",
@@ -431,7 +431,7 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
 }
 
 // Remove Airtable-specific keys from the fields.
-function extractFields(fields: PossibleField[]): ManagedCollectionField[] {
+function processFields(fields: PossibleField[]): ManagedCollectionField[] {
     return fields
         .map(field => {
             if (field.type === "unsupported") return null


plugins/airtable/src/FieldMapping.tsx
```diff
@@ -429,3 +429,45 @@ export function FieldMapping({ collection, dataSource, initialSlugFieldId }: Fie
         </form>
     )
 }
+
+// Remove Airtable-specific keys from the fields.
+function processFields(fields: PossibleField[]): ManagedCollectionField[] {
+    return fields
+        .map(field => {
```
kaloyanvi, line 436: nit: filter + map

madebyisaacr: Is it ok to use `as`? When I move filter before map, it has a typescript error unless I add `as ManagedCollectionField[]` at the end.

```typescript
// Remove Airtable-specific keys from the fields.
function processFields(fields: PossibleField[]): ManagedCollectionField[] {
    return fields
        .filter((field): field is PossibleField => field.type !== "unsupported")
        .map(field => {
            const value = {
                id: field.id,
                userEditable: field.userEditable,
                name: field.name,
                type: field.type,
            }

            switch (field.type) {
                case "array":
                    return {
                        ...value,
                        fields: field.fields,
                    }
                case "collectionReference":
                case "multiCollectionReference":
                    return {
                        ...value,
                        collectionId: field.collectionId,
                    }
                case "file":
                    return {
                        ...value,
                        allowedFileTypes: field.allowedFileTypes,
                    }
                case "enum":
                    return {
                        ...value,
                        cases: field.cases,
                    }
            }

            return value
        }) as ManagedCollectionField[]
}
```

PR #503, opened 2025-12-11
[CSV Plugin] select and create collections from plugin

Nick-Lucas:
https://github.com/framer/company/issues/35502

### Description

Supports selection and creation of Collections in the CMS directly from the plugin. Increases clarity of where data will be imported to and an escape hatch to create a new collection

### Testing

- [x] Open the plugin and import a CSV
  - [ ] Be on a collection
  - [ ] Open plugin with CMD+Opt+L or the plugins menu
  - [ ] Current collection should match
  - [ ] Import a CSV should work
- [x] Open a plugin on a managed collection and receive info warning
  - [ ] Be on a managed collection
  - [ ] Open plugin with CMD+Opt+L or the plugins menu
  - [ ] Warning toast appears about compatibility
- [x] Open a plugin via CMS->Add->Import Collection and create a collection
  - [ ] Be on a collection
  - [ ] CMS->Add->Import Collection
  - [ ] Plugins current Collection will be the original collection
  - [ ] Create a new collection
  - [ ] Current collection should switch to new collection
  - [ ] Import a CSV should work and import to the new collection


plugins/csv-import/src/CollectionSelector.tsx
```diff
@@ -0,0 +1,149 @@
...
+            onCollectionChange(newCollection)
+            setShowCreateCollectionLine(false)
+            setNewCollectionName("")
+        }
+    }
+
+    useEffect(() => {
+        if (showCreateCollectionLine) {
+            createCollectionInputRef.current?.focus()
+        }
```
elmarburke, line 46: can't you autofocus with the attribure on the input element?

elmarburke (approved):

plugins/csv-import/src/CollectionSelector.tsx
Nick-Lucas: It never actually gets removed from the DOM because nice animations, so I don't think that would work unfortunately

PR #513, opened 2026-01-12
CSV Import: Support datetime import

Nick-Lucas:
### Description

Adds support for the new displayTime flag on date fields, adds inference for times in JS/ISO format.

Rather than trying to support arbitrary flags across the system, we now map the field.type to a "virtual" type which splits date fields by displayTime into a datetime type. This simplifies things a lot

image
image

### Testing

[people.csv](https://github.com/user-attachments/files/24567028/people.csv)

- [x] Upload people.csv on a new collection and ensure the date and datetime fields are detected correctly
- [x] Upload people.csv on a existing collection and ensure the data maps and imports correctly

### DIFF
--- a/plugins/csv-import/src/routes/FieldMapper.tsx
+++ b/plugins/csv-import/src/routes/FieldMapper.tsx
@@ -181,9 +181,10 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
                     // Map to existing field
                     const targetField = existingFields.find(f => f.id === targetFieldId)
                     const targetVirtualType = targetField ? sdkTypeToVirtual(targetField) : null
-                    const hasTypeMismatch = targetField && targetVirtualType
-                        ? !isTypeCompatible(item.inferredField.inferredType, targetVirtualType)
-                        : false
+                    const hasTypeMismatch =
+                        targetField && targetVirtualType
+                            ? !isTypeCompatible(item.inferredField.inferredType, targetVirtualType)
+                            : false
 
                     return {
                         ...item,


plugins/csv-import/src/utils/typeInference.ts
```diff
@@ -13,10 +13,27 @@ const URL_PATTERN = /^https?:\/\/.+/i
...
+
+        // Check hours, minutes, seconds, milliseconds using UTC to avoid timezone issues
+        return (
+            date.getUTCHours() !== 0 ||
+            date.getUTCMinutes() !== 0 ||
+            date.getUTCSeconds() !== 0 ||
+            date.getUTCMilliseconds() !== 0
+        )
+    })
+}
```
triozer, line 31: The time detection treats any non‑zero UTC time components as “datetime”. At midnight, all components are zero, so this will infer “date”, even though the source clearly includes a time. That’s surprising and lossy.

Probably easier to split on the delimiter `T`?

Nick-Lucas: Yeah I wasn't sure how deep is worth going here. Reality is supporting more than perfect ISO strings is a big task to get right, particularly where the most common editor for CSVs is Excel and that doesn't even support/output ISO in the first place. So I was essentially thinking that for now it's not worth it and to wait for feedback on formats we need to support

Claude actually did draft out some extra logic here but I deleted it because of that

triozer: Well, Framer only supports ISO dates: https://www.framer.com/developers/property-controls#date-controltype.date

Nick-Lucas: So in that case I think this simplified code is fine, if it mis-detects the value is going to be invalid anyway.

In the future we could definitely add support for converting various formats

PR #516, opened 2026-01-13
CSV Import: fix top divider width

madebyisaacr:
### Description

This pull request fixes the top divider width in CSV import.

Before: divider has 15px padding on left and none on right.
Pasted Graphic 7

After: divider has 15px padding on both sides.
image

### Testing

- [ ] Open CSV Import, upload a CSV, and look at the top divider


Nick-Lucas:
This one was a design choice with @johannes-ger so unless you can convince him we probably won't merge this :)

PR #518, opened 2026-01-13
CSV Import: createCollection permission

madebyisaacr:
### Description

This pull request removes the createCollection requirement from importing CSV files, so users with permission to edit collections but not create new ones can use CSV import. createCollection is still checked for creating new collections in the plugin, but not for importing into existing collections.

### Testing

- [x] Import a CSV into a project where you have access to edit collections but not create new collections.

### DIFF
--- a/plugins/csv-import/src/App.tsx
+++ b/plugins/csv-import/src/App.tsx
@@ -17,12 +17,7 @@ import { ImportError, type ImportItem, prepareImportPayload } from "./utils/prep
 
 export function App({ initialCollection }: { initialCollection: Collection | null }) {
     const [collection, setCollection] = useState<Collection | null>(initialCollection)
-    const hasAllPermissions = useIsAllowedTo(
-        "Collection.addItems",
-        "Collection.addFields",
-        "Collection.removeFields",
-        "createCollection"
-    )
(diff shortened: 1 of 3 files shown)


plugins/csv-import/src/App.tsx
```diff
@@ -17,12 +17,7 @@ import { ImportError, type ImportItem, prepareImportPayload } from "./utils/prep
 
 export function App({ initialCollection }: { initialCollection: Collection | null }) {
     const [collection, setCollection] = useState<Collection | null>(initialCollection)
-    const hasAllPermissions = useIsAllowedTo(
-        "Collection.addItems",
-        "Collection.addFields",
-        "Collection.removeFields",
-        "createCollection"
-    )
+    const hasAllPermissions = useIsAllowedTo("Collection.addItems", "Collection.addFields", "Collection.removeFields")
```
Nick-Lucas, line 20: I don't understand why we're removing this, we need it right?

madebyisaacr: `createCollection` isn't used here so we don't need to check the permission. The "createCollection" permission is still checked in CollectionSelector.tsx, and if it's not allowed, the new collection option is hidden.

Nick-Lucas: _right_ I understand now! Yeah LGTM then

PR #519, opened 2026-01-13
CSV Import: allow using ignored fields as slug

madebyisaacr:
### Description

This pull request updates the CSV import plugin to allow using fields that are ignored/disabled as the slug field. This fixes an issue where if your CSV has a slug column, it would create duplicate fields because it would import as both the slug value and as a separate field.

I also made it check if there's a column that matches the slug field name, and if so it will be selected by default and the row will be ignored by default to avoid creating a duplicate slug field.

### Testing

Google Sheet for testing: https://docs.google.com/spreadsheets/d/1JuOKh_Wb0uIytKe4eE7uI6Js826i8V_uf2oFtyolDtA/edit?gid=1174080602#gid=1174080602

- [x] Check if the "Slug" column is the default selected slug field even though it is not the first column.
- [x] Disable the "Slug" column and check if it still allows you to use it as the slug field.

### DIFF
--- a/plugins/csv-import/src/routes/FieldMapper.tsx
+++ b/plugins/csv-import/src/routes/FieldMapper.tsx
@@ -51,35 +51,31 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
                 const fields = await collection.getFields()
                 setExistingFields(fields)
 
-                // Track which existing fields get mapped
-                const mappedFieldIds = new Set<string>()
-
                 const inferredFields = inferFieldsFromCSV(csvRecords)
 
-                // Determine which column should be the slug field
-                // Ensure the slug column is in inferredFields (exists in CSV) so it appears in possibleSlugFields
-                const slugColumnNameFromCollection =
-                    collection.slugFieldName &&
-                    inferredFields.find(field => field.columnName === collection.slugFieldName)?.columnName
-                const slugColumnName =
-                    slugColumnNameFromCollection ??
+                // Existing Slug field match against CSV column if any
+                const matchedSlugColumnName = collection.slugFieldName
+                    ? inferredFields.find(field => field.columnName === collection.slugFieldName)?.columnName
+                    : undefined
+
+                // Column we will suggest as slug field on the UI
+                const suggestedSlugColumnName =
+                    matchedSlugColumnName ??
                     inferredFields.find(field => isValidSlugColumn(field.columnName, csvRecords))?.columnName
 
                 // Create initial mappings based on name matching
                 const initialMappings: FieldMappingItem[] = inferredFields.map(inferredField => {
-                    // Only ignore slug field if it matches the collection's slugFieldName
+                    // ignore slug field if it matches the collection's slugFieldName
                     // If it was auto-detected, keep it enabled
-                    const isSlugField =
-                        slugColumnNameFromCollection && inferredField.columnName === slugColumnNameFromCollection
+                    const isSlugField = matchedSlugColumnName
+                        ? inferredField.columnName === matchedSlugColumnName
+                        : false
 
                     // Try to find an existing field with matching name
                     const matchingField = fields.find(f => f.name.toLowerCase() === inferredField.name.toLowerCase())
-
                     if (matchingField) {
                         const hasTypeMismatch = !isTypeCompatible(inferredField.inferredType, matchingField.type)
-                        if (!isSlugField) {
-                            mappedFieldIds.add(matchingField.id)
-                        }
+
                         return {
                             inferredField,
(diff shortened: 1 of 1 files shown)


plugins/csv-import/src/routes/FieldMapper.tsx
```diff
@@ -78,10 +78,31 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
...
-
                 const possibleSlugFields = calculatePossibleSlugFields(initialMappings, csvRecords)
-                setSelectedSlugFieldName(possibleSlugFields[0]?.columnName ?? null)
+                const slugField = collection.slugFieldName
+                    ? (possibleSlugFields.find(field => field.columnName === collection.slugFieldName) ??
+                      possibleSlugFields[0])
+                    : possibleSlugFields[0]
+
+                // If we found a column matching the collection's slugFieldName, set it to ignore to avoid duplicate slug fields
+                const finalMappings = initialMappings.map(mapping => {
```
Nick-Lucas: I think we can roll this logic into the original map, would be simpler and more performant

plugins/csv-import/src/routes/FieldMapper.tsx
```diff
@@ -78,10 +78,31 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
                     }
                 })
 
-                setMappings(initialMappings)
-
                 const possibleSlugFields = calculatePossibleSlugFields(initialMappings, csvRecords)
-                setSelectedSlugFieldName(possibleSlugFields[0]?.columnName ?? null)
+                const slugField = collection.slugFieldName
+                    ? (possibleSlugFields.find(field => field.columnName === collection.slugFieldName) ??
```
Nick-Lucas: If we match an existing slug mapped on its own column, won't that result in us ignoring the slug values for new rows? I need to test this out

plugins/csv-import/src/routes/FieldMapper.tsx
Nick-Lucas: I had a little play on `git switch csv-import-all-slug-fields-nick` and pretty sure it will work

madebyisaacr: Good idea. I merged your commit into this.

Nick-Lucas: Definitely retest it and tweak as needed, I just drafted it out with Claude and pushed

plugins/csv-import/src/routes/FieldMapper.tsx
madebyisaacr: I don't think so. Slug values are never ignored - you get an error in `additems()` if a row has no slug.

plugins/csv-import/src/routes/FieldMapper.tsx
```diff
@@ -54,34 +56,48 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
...
-                        mappedFieldIds.add(matchingField.id)
+                        if (!isSlugField) {
+                            mappedFieldIds.add(matchingField.id)
+                        }
                         return {
                             inferredField,
-                            action: "map",
-                            targetFieldId: matchingField.id,
-                            hasTypeMismatch,
+                            action: isSlugField ? "ignore" : "map",
```
Nick-Lucas, line 81: Hm I just tried an import on a new collection and it didn't auto-ignore the slug column, maybe the change I proposed needs a bugfix

madebyisaacr: Was the slug column's name the same as the slug field name on the collection, usually "Slug"? It only auto ignores it if the slug column name matches, because we don't want to ignore it by default if it's using a different column for the slug like "Title" for example.

madebyisaacr: If you copy and paste this table in with the Slug column, it should make the Slug field ignored by default. But if you copy without the first column then it should make Title the slug field and not make Title ignored.

https://docs.google.com/spreadsheets/d/1JuOKh_Wb0uIytKe4eE7uI6Js826i8V_uf2oFtyolDtA/edit?gid=1174080602#gid=1174080602

Nick-Lucas: Ah good point.

Image

So I think this might be confusing to be honest, not sure most users will understand why that column is disabled. And also if we're only covering the scenario where the slug name is the same as a column name, then displaying the mapping isn't a bad thing or incorrect right?

Maybe we should bring @johannes-ger in on the discussion?

Nick-Lucas (approved):

PR #533, opened 2026-01-21
Support (Secondary) Locations via collection references in Ashby Plugin

niekert:
### Description

- Add new Locations data source that extracts unique locations from jobs board
- Add location reference and secondary locations fields to jobs data source
- Allow data source selection in setup UI (Jobs vs Locations)
- Add CLAUDE.md for repository documentation

### Testing

  - [x] Setup Locations Collection                                                                                                                                                                           
    - Open Framer and create a new CMS collection                                                                                                                                                        
    - Launch the Ashby plugin                                                                                                                                                                            
    - Enter your Job Board Name (see slack)                                                                                                                                                   
    - Select "Locations" from the Collection dropdown                                                                                                                                                    
    - Click "Next"                                                                                                                                                                                       
    - Map the desired fields:                                                                                                                                                                            
    - Click "Sync" and verify locations are imported                                                                                                                                                     
    - Confirm unique locations are extracted from both primary and secondary job locations                                                                                                               
  - [x] Setup Jobs Collection with Location References                                                                                                                                                       
    - Create another new CMS collection                                                                                                                                                                  
    - Launch the Ashby plugin again                                                                                                                                                                      
    - Enter the same Job Board Name                                                                                                                                                                      
    - Select "Jobs" from the Collection dropdown                                                                                                                                                         
    - Click "Next"                                                                                                                                                                                       
    - Map standard job fields as needed                                                                                                                                                                  
    - Map location reference fields:                                                                                                                                                                     
        - Location Reference → should show the Locations collection as target                                                                                                                            
      - Secondary Locations → should show the Locations collection as target (multi-reference)                                                                                                           
    - Click "Sync"                                                                                                                                                                                       
  - [x] Verify Data Integrity                                                                                                                                                                                
    - Open a synced Job item                                                                                                                                                                             
    - Confirm "Location Reference" links to the correct Location item                                                                                                                                    
    - Confirm "Secondary Locations" contains all additional locations for that job                                                                                                                       
    - Cross-reference with the Ashby job board to verify accuracy                                                                                                                                        
    - Check that location slugs match between Jobs references and Locations collection IDs


CLAUDE.md
triozer, line 1: Mh, probably best to remove? Wdyt?

plugins/ashby/src/dataSources.ts
```diff
@@ -39,14 +50,95 @@ export type AshbyField = ManagedCollectionFieldInput &
...
+const locationsDataSourceName = "Locations"
+
+export function slugify(text: string): string {
+    return text
+        .toLowerCase()
+        .trim()
+        .replace(/[^\p{L}\p{N}\s-]/gu, "")
+        .replace(/[\s_-]+/g, "-")
+        .replace(/^-+|-+$/g, "")
+}
```
triozer: Should we re-use Framer's slugify method?

`slugify.ts`
```suggestion
// Note: We don't use the "slugify" package here because we want a very specific
// behavior for our CMS and web pages. Make sure to pick the right slugify
// function for your use case!

// Characters that are problematic in URLs and should be replaced with dashes:
// - Whitespace (\s)
// - Underscore (_) - Google specifically recommends using `-` instead https://developers.google.com/search/docs/crawling-indexing/url-structure#:~:text=example.com/%F0%9F%A6%99%E2%9C%A8-,Use%20hyphens%20to%20separate%20words,-We%20recommend%20separating
// - URL reserved delimiters: ? # [ ] @
// - Characters with special meaning: ! $ & ' * + , ; : = " < > % { } | \ ^ ` /
//   - : denotes paths on MacOS / reserved for drive letters on Windows
// All other characters (including unicode letters, numbers, symbols) are kept.
const unsafeSlugCharactersRegExp = /[\s_?#[\]@!$&'*+,;:="<>%{}|\\^`/]+/gu

/**
 * Trim leading and trailing dashes from a string.
 * We can't use a regexp, because matching e.g. /-+$/gu leads to polynomial backtracking with e.g. '-'.repeat(54773) + '\x00-'
 */
export function trimDashes(str: string): string {
	let start = 0
	let end = str.length
	while (start < end && str[start] === "-") start++
	while (end > start && str[end - 1] === "-") end--
	return str.slice(start, end)
}

/**
 * Takes a freeform string and converts it to a URL-safe slug.
 * Replaces problematic URL characters and combinations with dashes while preserving:
 * - Unicode letters and numbers
 * - Unicode symbols (±, §, etc.)
 * - URL-safe punctuation: - . ~ ( )
 */
export function slugify(value: string): string {
	return trimDashes(value.trim().toLowerCase().replace(unsafeSlugCharactersRegExp, "-"))
}
```

CLAUDE.md
niekert: Yeah. I think its nice to add a `claude.md` but this one was generated with `/init` and references unncessary plugin specific stuff

plugins/ashby/src/dataSources.ts
niekert: done!

PR #536, opened 2026-01-22
CSV Import: ignore divider field

triozer:
### Description

We should ignore divider fields when loading all fields in the mapping view.

### Testing

- [x] Dividers are skipped
  - Add a divider in a CMS Collection
  - Open this plugin
  - Ensure that the divider fields are not shown

### DIFF
--- a/plugins/csv-import/src/utils/prepareImportPayload.ts
+++ b/plugins/csv-import/src/utils/prepareImportPayload.ts
@@ -9,6 +9,7 @@ import {
 } from "framer-plugin"
 import type { FieldMappingItem } from "../components/FieldMapperRow"
 import { assert } from "./assert"
+import { getDataFields } from "./filterFields"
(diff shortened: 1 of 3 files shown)


plugins/csv-import/src/utils/prepareImportPayload.ts
```diff
@@ -194,7 +194,9 @@ export async function prepareImportPayload(opts: ProcessRecordsWithFieldMappingO
     }
 
     const existingItems = await opts.collection.getItems()
-    const fields = await opts.collection.getFields()
+    const allFields = await opts.collection.getFields()
+    // Filter out divider fields - they're visual separators and can't store data
+    const fields = allFields.filter(field => field.type !== "divider" && field.type !== "unsupported")
```
kaloyanvi: nit: can this be shared as helper in the other place too

PR #543, opened 2026-01-26
Required Fields section in CSV Import needs improvement

Nick-Lucas:
### Description

image

### Testing

[people.csv](https://github.com/user-attachments/files/24863174/people.csv)

- [ ] Check required fields validation
  - [ ] Upload people.csv
  - [ ] Make member_since and last_login required in Edit Fields UI
  - [ ] Upload people.csv and see there are 2 different validation warnings

### DIFF
--- a/plugins/csv-import/src/components/FieldMapperRow.tsx
+++ b/plugins/csv-import/src/components/FieldMapperRow.tsx
@@ -1,9 +1,9 @@
 import type { Field } from "framer-plugin"
-import { ChevronIcon } from "./ChevronIcon"
 import { labelByFieldType } from "../utils/fieldLabels"
 import type { InferredField } from "../utils/typeInference"
 import type { VirtualFieldType } from "../utils/virtualTypes"
 import { sdkTypeToVirtual } from "../utils/virtualTypes"
+import { ChevronIcon } from "./ChevronIcon"
 
 export type MappingAction = "create" | "map" | "ignore"
 


plugins/csv-import/src/routes/FieldMapper.tsx
```diff
@@ -367,23 +383,40 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
...
+                                >
+                                    <path
+                                        fill="transparent"
+                                        stroke="#999"
+                                        strokeLinecap="round"
+                                        strokeLinejoin="round"
+                                        strokeWidth="1.5"
+                                        d="m2.5 7 3-3-3-3"
+                                    />
+                                </svg>
```
madebyisaacr: This arrow icon is duplicated 3 times, twice in this file and once in FieldMapperRow.tsx. Make a reusable component.

plugins/csv-import/src/routes/FieldMapper.tsx
```diff
@@ -417,12 +450,7 @@ export function FieldMapper({ collection, csvRecords, onSubmit }: FieldMapperPro
                             <strong>{csvRecords.length}</strong> Items
                         </span>
                     </div>
-                    <button
-                        type="submit"
-                        className="framer-button-primary"
-                        disabled={!canSubmit}
-                        title={unmappedRequiredFields.length > 0 ? "Map all required fields to continue" : undefined}
-                    >
+                    <button type="submit" className="framer-button-primary" disabled={!canSubmit}>
```
madebyisaacr, line 423: Does title need to be removed?

Nick-Lucas: Yes I removed it, a tooltip is not a place to put a validation error. We have better places already

plugins/csv-import/src/routes/FieldMapper.tsx
Nick-Lucas: Good catch, I think the same is true of quite a bit of CSS too but I'll de-dup the icon as that's lower hanging fruit

PR #544, opened 2026-01-26
CSV import fixes

madebyisaacr:
### Description

- Remove default name when creating new collection.
- Fixes the home page quickly flashing between clicking the import button and import finishing.
- Deletes unused CSS
- Makes the new collection input text position perfectly match the select text position.

### Testing

- [x] Create a new collection
- [x] Import a CSV file

### DIFF
--- a/plugins/csv-import/src/components/CollectionSelector.tsx
+++ b/plugins/csv-import/src/components/CollectionSelector.tsx
@@ -27,7 +27,6 @@ export function CollectionSelector({ forceCreate, collection, onCollectionChange
     }, [isCreatingNew])
 
     // Validate for duplicates on initial render and when collections change
-    // Only show error if user has manually changed the name from the auto-generated one
     useEffect(() => {
         if (isCreatingNew) {
             const trimmedName = newCollectionName.trim()


plugins/csv-import/src/components/CollectionSelector.tsx
```diff
@@ -205,5 +217,27 @@ function useCollections(collection: Collection | null) {
...
     }, [collection])
 
-    return collections
+    return {
+        writableCollections: allCollections.filter(collection => collection.managedBy === "user"),
+        allCollections,
+    }
+}
+
+function findAvailableCollectionName(baseName: string, allCollections: Collection[]): string {
```
Nick-Lucas: Rather than this approach, can we just remove the default? 

There's not really any valid reason someone would use this plugin and want to use `Collection` as the name at all, and it's too easy to accidentally end up with that name and be stuck with it until after the import

plugins/csv-import/src/App.tsx
```diff
@@ -118,11 +111,15 @@ export function App({ initialCollection }: { initialCollection: Collection | nul
 
                 if (error instanceof ImportError || error instanceof Error) {
                     framer.notify(error.message, { variant: "error" })
-                    return
+                } else {
+                    framer.notify("Error processing CSV file. Check console for details.", {
```
Nick-Lucas: > Check console for details

I know this was here before, but I really don't think checking the console is a good UX. Could we just take `String(error)` or `error.message` and use it in place of "Check console for details"?

plugins/csv-import/src/components/CollectionSelector.tsx
```diff
@@ -205,5 +209,8 @@ function useCollections(collection: Collection | null) {
         }
     }, [collection])
 
-    return collections
+    return {
```
Nick-Lucas: Not sure I understand why so many changes here, why the distinction for writableCollections?

madebyisaacr: Those changes were necessary before when it needed to get all collections to generate a default name. But now that there's no default name, its not needed anymore so I reverted it.

Issue #90, opened 2024-10-23
Allow to pass `distDir` and `buildCommand` to `framer-plugin-tools` package


andirsun:
**Describe the bug**
We are creating a Plugin inside an NX mono-repo, when the folder structure is something like this
```
apps/
  another-app/
  my-plugin/ (This is my plugin code)
dist/
 apps/
   another-app/
   my-plugin/ (This is the directory of my compiled plugin files)
package.json (main package.json shared for all projects)
```

The problem is that the package [framer-plugin-tools](https://www.npmjs.com/package/framer-plugin-tools?activeTab=code) you recomend to package the compiled files into a ZIP is not receiving any parameters and in our case we should be able to modify
- **distDir**: The default distDir ir `const distPath = path.join(process.cwd(), "dist")` but in our project the distDir should be `dist/my-plugin/`

- **buildCommand**: the default build command is `const buildProcess = exec("npm run build")` but in our project is reserved to compile all projects, so we should be able to pass `npm run build:my-plugin`

**Expected behavior**

**Screenshots**

**Additional context**

PR #549, opened 2026-02-05
Move framer-plugin-tools into monorepo as workspace package

niekert:
### Description

This PR adds https://github.com/framer/framer-plugin-tools as a workspace package under `packages/plugin-tools`.

The CLI footprint the same

```
yarn framer-plugin-tools pack
```
(with some additional optional options)

Changes:
 - All plugins use the workspace version of plugin-tools
 - plugin-tools rewritten in TS
 - the package manager is now detected by the CLI command for running `build`
 - Addressed: https://github.com/framer/plugins/issues/90

### Pre-merge Checklist

- [x] Publish a new stable version of `framer-plugin-tools`
- [x] Archive the previous repo

### Testing

- [x] All plugins can pack succesfully
  - [x] Check out this branch
  - [x] Run `yarn turbo run pack` to run the pack command in all plugins
  - [x] No errors
- [x] The pack command still works on the published version (alpha)
  - [x] Create a new plugin `npm create framer-plugin test-tools-plugin`
  - [x] cd `./test-tools-plugin`
  - [x] `npx framer-plugin-tools@alpha pack`
  - [x] It creates a valid `plugin.zip`

### DIFF
--- a/plugins/code-link/package.json
+++ b/plugins/code-link/package.json
@@ -10,7 +10,7 @@
         "check-eslint": "run g:check-eslint",
         "check-typescript": "run g:check-typescript",
         "preview": "run g:preview",
-        "pack": "npx framer-plugin-tools@latest pack"
+        "pack": "run g:pack"
     },
     "dependencies": {
         "@code-link/shared": "workspace:*",


packages/plugin-tools/src/index.ts
```diff
@@ -0,0 +1,37 @@
+import { runBuildScript, zipDistFolder } from "./build-and-zip"
```
niekert: based on https://github.com/framer/framer-plugin-tools/blob/main/index.js#L10

packages/plugin-tools/src/index.ts
```diff
@@ -0,0 +1 @@
+export { runPluginBuildScript, zipPluginDistribution } from "./lib"
```
elmarburke, line 1: **nitpick**: no barrel files (as in the other code bases) plz

packages/plugin-tools/src/lib.ts
```diff
@@ -0,0 +1,85 @@
...
+import path from "path"
+
+/**
+ * Naive package manager detection by checking for lock files in the current directory and parent directories.
+ * @param cwd - The current working directory.
+ * @returns the package manager entrypoint CLI command.
+ */
+export function detectPackageManager(cwd: string): string {
+    let dir = cwd
+    for (let i = 0; i < 3; i++) {
```
elmarburke, line 15: **thought**: document why the magic number is 3

packages/plugin-tools/src/index.ts
niekert: Hmm what would you recommend since these are the exports for if you want to use `framer-plugin-tools` as a library. Just use `/lib.ts` as entrypoint?

elmarburke: Oooooh, ehm, yeah, that changes it a bit. I didn't think about that path. It's fine! Keep it like this!

PR #553, opened 2026-02-12
Code Link Docs & Folder UX

huntercaron:
### Description

Improves UX around folders by telling you in the CLI where its syncing to. Also updates readme with more info.

### Testing

- [x] In `plugins/code-link` run the plugin
- [x] In a Framer project, open the dev plugin, take note of the id in the command
- [x] In `code-link-cli` run `yarn dev [id from plugin]`
  - [x] Verify it mentions the folder name being created

### DIFF
--- a/packages/code-link-cli/src/controller.ts
+++ b/packages/code-link-cli/src/controller.ts
@@ -957,21 +957,20 @@ async function executeEffect(
                 return []
             }
 
-            const relativeDir = config.projectDir
-                ? "./" + (path.relative(process.cwd(), config.projectDir) || ".")
-                : null
+            const relative = config.projectDir ? path.relative(process.cwd(), config.projectDir) : null
+            const relativeDirectory = relative != null ? (relative ? "./" + relative : ".") : null
 
-            if (effect.totalCount === 0 && relativeDir) {
+            if (effect.totalCount === 0 && relativeDirectory) {
                 if (config.projectDirCreated) {
-                    success(`Created ${relativeDir} folder`)
+                    success(`Created ${relativeDirectory} folder`)
                 } else {
-                    success(`Syncing to ${relativeDir} folder`)
+                    success(`Syncing to ${relativeDirectory} folder`)
                 }
-            } else if (relativeDir && config.projectDirCreated) {
-                success(`Synced into ${relativeDir} (${effect.updatedCount} files added)`)
-            } else if (relativeDir) {
+            } else if (relativeDirectory && config.projectDirCreated) {
+                success(`Synced into ${relativeDirectory} (${effect.updatedCount} files added)`)
+            } else if (relativeDirectory) {
                 success(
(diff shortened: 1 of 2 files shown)


packages/code-link-cli/src/utils/project.ts
```diff
@@ -40,17 +40,17 @@ export async function findOrCreateProjectDir(
     projectHash: string,
     projectName?: string,
     explicitDir?: string
-): Promise<string> {
+): Promise<{ dir: string; created: boolean }> {
```
Nick-Lucas: nit: I would favour spelling out the whole word instead of shortening to `dir`, slight reduction in mental load when a codebase adheres to that principle. `directory` or `path` might produce more clarity downstream

packages/code-link-cli/src/controller.ts
```diff
@@ -681,7 +682,9 @@ async function executeEffect(
             if (!config.projectDir) {
                 const projectName = config.explicitName ?? effect.projectInfo.projectName
 
-                config.projectDir = await findOrCreateProjectDir(config.projectHash, projectName, config.explicitDir)
+                const result = await findOrCreateProjectDir(config.projectHash, projectName, config.explicitDir)
```
Nick-Lucas: nit: similarly, I suggest naming variables for what they are, `result` is obvious, is it `projectDirectory` and then you could have `projectDirectory.path`/`projectDirectory.created` later on?

packages/code-link-cli/src/controller.ts
```diff
@@ -1041,6 +1062,16 @@ export async function start(config: Config): Promise<void> {
         }
 
         void (async () => {
+            // Cancel any pending disconnect message (fast reconnect)
```
Nick-Lucas: nit: as a reader I'd rather see 'why' than 'what' in comments, or no comment at all. Why is`cancelDisconnectMessage()` important? what does it avoid breaking? The comment basically describes the function name currently

plugins/code-link/README.md
```diff
@@ -1,13 +1,30 @@
...
+3. Paste and run the command in your terminal
+4. Edit files in `{project}/files/` — changes sync to Framer
+
+## CLI Options
+
+| Flag | Description |
+| --- | --- |
+| `-n, --name <name>` | Project name for the created directory |
+| `-d, --dir <directory>` | Target project directory |
+| `-v, --verbose` | Enable debug logging |
```
Nick-Lucas, line 30: Great readme :)

packages/code-link-cli/src/controller.ts
huntercaron: good point, clear enough without it!

packages/code-link-cli/src/utils/project.ts
huntercaron: Gooood point. I should probably have followed the Framer convention to never shorten but would be a pretty big diff now. Ill improve in the areas is touching for now

PR #556, opened 2026-02-13
Code Link polish from feedback

huntercaron:
### Description

- Improves support when multiple projects have the same name
- Improves polling strategy when plugin is left open for a while
- Copy tweaks for pluralization
- Add log for when files are added during intial sync
- Log deletions that happen locally
- Small UI polish

### Testing

- [x] In `plugins/code-link` run the plugin
- [x] In a Framer project, open the dev plugin, take note of the id in the command
- [x] In `code-link-cli` run `yarn dev [id from plugin]` with a project named "Untitled"
  - [x] Verify a local folder named "Untitled" is created
- [x] In a second project also named "Untitled" repeat the steps to run `yarn dev [id from plugin]` at the same time
  - [x] Verify a local folder named "Untitled-[hash]" is created

### DIFF
--- a/packages/code-link-cli/package.json
+++ b/packages/code-link-cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "framer-code-link",
-    "version": "0.15.0",
+    "version": "0.17.0",
     "description": "CLI tool for syncing Framer code components - controller-centric architecture",
     "main": "dist/index.mjs",
     "type": "module",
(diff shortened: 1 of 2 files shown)


packages/code-link-cli/src/utils/project.ts
```diff
@@ -39,15 +39,16 @@ export async function getProjectHashFromCwd(): Promise<string | null> {
 export async function findOrCreateProjectDirectory(
     projectHash: string,
     projectName?: string,
-    explicitDirectory?: string
-): Promise<{ directory: string; created: boolean }> {
+    explicitDirectory?: string,
+    baseDirectory?: string
```
Nick-Lucas, line 43: We might be at the point where an opts object with named keys is preferred. 4 strings in a row doesn't do much to check positional correctness, and since they're mostly optional you also have to pass a bunch of undefineds to skip the middle ones

PR #573, opened 2026-02-27
Replace \ with / in file names when packing

madebyisaacr:
### Description

This pull request updates framer-plugin-tools to replace \ with / in file names.

[This plugin](https://www.framer.com/marketplace/plugins/cora-bridge/preview/)'s CSS file has the name:
`assets\index-DA6fbhnP.css`

Which gets mapped to this when URL encoded:
`assets%5Cindex-DA6fbhnP.css` ([link](https://591it18f7qouo0kvnmmcyvtll-ivhy3q5lhimk7kvvwa5k2baek5vmepew.plugins.framercdn.com/assets%5Cindex-DA6fbhnP.css))

But the stylesheet in the index.html file references `/assets/index-DA6fbhnP.css` so the file doesn't load.

I have never seen this before today, so it's probably an unusual quirk of using Windows and having an abnormal build config or something like that.

### Testing

- [x] Test packing a plugin with a file name including a \ and verify that when you unzip the plugin.zip file, it has a / in the file name

The way I tested on macOS is by opening the google-sheets plugin, moving the src/assets/hero.png file to the src/ folder and renaming it to "assets\hero.png" to simulate using a \ for directories like on Windows.

### DIFF
--- a/packages/plugin-tools/src/lib.ts
+++ b/packages/plugin-tools/src/lib.ts
@@ -35,6 +35,15 @@ interface ZipPluginDistributionOptions {
     zipFileName: string
 }
 
+// Normalize all entry paths to use forward slashes
+export function normalizeZipEntryPaths(zip: AdmZip): void {
+    for (const entry of zip.getEntries()) {
+        if (entry.entryName.includes("\\")) {
+            entry.entryName = entry.entryName.replace(/\\/g, "/")
+        }
+    }
+}
+
 export function zipPluginDistribution(options: ZipPluginDistributionOptions): string {
     const distPath = path.isAbsolute(options.distPath) ? options.distPath : path.join(options.cwd, options.distPath)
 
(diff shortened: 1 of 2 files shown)


packages/plugin-tools/src/lib.ts
```diff
@@ -54,6 +54,14 @@ export function zipPluginDistribution(options: ZipPluginDistributionOptions): st
     zip.addLocalFolder(distPath)
     zip.deleteFile(markerFileName)
     zip.addFile(markerFileName, Buffer.from("true", "utf-8"))
+
+    // Normalize all entry paths to use forward slashes
+    for (const entry of zip.getEntries()) {
```
niekert: Lets add a test for this in `lib.test.ts`

madebyisaacr: Added a test

Issue #528, opened 2026-01-19
Implement setCloseWarning in plugins


madebyisaacr:
Add the new setCloseWarning function to each of the CMS plugins. While syncing is in progress, add a close warning to prevent accidentally closing.

- [x] Notion
- [ ] Airtable
- [x] Google Sheets
- [ ] PR․co
- [ ] Recruitee
- [ ] Ashby
- [ ] Greenhouse
- [ ] HubSpot
- [x] CSV Import
- [ ] Locale Sync
- [ ] CMS Starter

PR #583, opened 2026-03-03
Add close warning to job board plugins

madebyisaacr:
### Description

This pull request implements `setCloseWarning` in Ashby, Recruitee, PR.co, and Greenhouse plugins. Based on the Notion plugin.

Part of https://github.com/framer/plugins/issues/528

Only Ashby and PR.co have been tested. The API keys I have for the Recruitee and Greenhouse plugins no longer work. 

### Changelog

- Added a warning when attempting to close the plugin while importing.

### Testing

- [ ] Ashby
- [ ] Greenhouse
- [ ] Recruitee
- [ ] PR.co

### DIFF
--- a/plugins/ashby/src/components/FieldMapping.tsx
+++ b/plugins/ashby/src/components/FieldMapping.tsx
@@ -232,6 +232,8 @@ export function FieldMapping({ jobBoardName, collection, dataSource, initialSlug
                 await framer.setCloseWarning(false)
                 setStatus("mapping-fields")
             }
+
+            await framer.setCloseWarning(false)
         }
 
         void task()
--- a/plugins/prco/src/components/FieldMapping.tsx
+++ b/plugins/prco/src/components/FieldMapping.tsx
@@ -226,14 +226,13 @@ export function FieldMapping({ pressRoomId, collection, dataSource, initialSlugF
             } catch (error) {
                 await framer.setCloseWarning(false)
                 console.error(error)
-                framer.notify(
-                    `Failed to sync collection “${dataSource.id}”. Check the logs for more details.`,
-                    {
-                        variant: "error",
-                    }
-                )
+                framer.notify(`Failed to sync collection “${dataSource.id}”. Check the logs for more details.`, {
+                    variant: "error",
+                })
                 setStatus("mapping-fields")
             }
+
+            await framer.setCloseWarning(false)
         }
 
         void task()
--- a/plugins/recruitee/src/components/FieldMapping.tsx
+++ b/plugins/recruitee/src/components/FieldMapping.tsx
@@ -235,6 +235,8 @@ export function FieldMapping({ companyId, token, collection, dataSource, initial
                 })
                 setStatus("mapping-fields")
             }
+
+            await framer.setCloseWarning(false)
         }
 
         void task()


plugins/ashby/src/data.ts
```diff
@@ -300,6 +300,8 @@ export async function syncExistingCollection(
         })
     }
 
+    void framer.setCloseWarning("Synchronization in progress. Closing will cancel the sync.")
```
Nick-Lucas: When inside an async method we should definitely be awaiting promises, not voiding them. Could we check that over?

plugins/ashby/src/components/SelectDataSource.tsx
```diff
@@ -43,10 +43,14 @@ export function SelectDataSource({
...
                     getDataSource(parsedJobBoardName, selectedDataSourceId)
                         .then(dataSource => {
+                            void framer.setCloseWarning(
+                                "Synchronization setup in progress. Closing will cancel the sync."
+                            )
                             onSelectDataSource(dataSource)
                             onSelectJobBoardName(parsedJobBoardName)
                         })
                         .catch((error: unknown) => {
+                            void framer.setCloseWarning(false)
```
Nick-Lucas: this file in general is in a poor state with respecting/awaiting promises appropriately, getDataSource isn't even being awaited when it could be. I think this is a good opprtunity to sort that and await setCloseWarning properly

plugins/ashby/src/components/SelectDataSource.tsx
```diff
@@ -42,15 +42,19 @@ export function SelectDataSource({
...
                             onSelectDataSource(dataSource)
                             onSelectJobBoardName(parsedJobBoardName)
                         })
-                        .catch((error: unknown) => {
+                        .catch(async (error: unknown) => {
                             console.error(error)
                             framer.notify(error instanceof Error ? error.message : "An unknown error occurred", {
                                 variant: "error",
                             })
+                            await framer.setCloseWarning(false)
```
Nick-Lucas, line 57: This probably needs to go in finally, else on a success it will never disable

plugins/greenhouse/src/components/SelectDataSource.tsx
```diff
@@ -32,11 +32,13 @@ export function SelectDataSource({
...
             getDataSource(boardToken, selectedDataSourceId)
-                .then(dataSource => {
+                .then(async dataSource => {
+                    await framer.setCloseWarning("Synchronization setup in progress. Closing will cancel the sync.")
                     onSelectDataSource(dataSource)
                     onSelectBoardToken(boardToken)
                 })
-                .catch((error: unknown) => {
+                .catch(async (error: unknown) => {
+                    await framer.setCloseWarning(false)
```
Nick-Lucas, line 41: also needs to go to finally right?

plugins/prco/src/components/FieldMapping.tsx
```diff
@@ -194,42 +194,49 @@ export function FieldMapping({ pressRoomId, collection, dataSource, initialSlugF
             return
         }
 
-        setStatus("syncing-collection")
+        const task = async () => {
+            setStatus("syncing-collection")
+            await framer.setCloseWarning("Synchronization in progress. Closing will cancel the sync.")
```
Nick-Lucas, line 199: where are we ending this aside from in catch blocks?

plugins/ashby/src/components/FieldMapping.tsx
```diff
@@ -195,44 +195,46 @@ export function FieldMapping({ jobBoardName, collection, dataSource, initialSlug
             return
         }
 
-        // try {
-        setStatus("syncing-collection")
+        const task = async () => {
+            setStatus("syncing-collection")
+            await framer.setCloseWarning("Synchronization in progress. Closing will cancel the sync.")
```
Nick-Lucas, line 200: where are we ending this aside from in catch blocks?

plugins/ashby/src/components/SelectDataSource.tsx
madebyisaacr: This close warning shouldn't be disabled because it needs to be shown on the field mapping page. That's how the Notion plugin does it.

If close warning is moved to finally, it will disable the close warning on the field mapping screen.

plugins/prco/src/components/FieldMapping.tsx
madebyisaacr: I added setCloseWarning(false) to the task function in FieldMapping.tsx in all plugins. But I don't think it will have any effect because the plugin closes itself instantly after syncing successfully.

PR #599, opened 2026-03-20
Add a `--once` flag to `framer-code-link`

iamakulov:
### Description

This pull request adds a `--once` flag to the `framer-code-link` command, to exit after the initial run. Useful to run syncs programmatically.

### Testing

- [x] Run locally with `--once`; confirm the changes are synced and the CLI exited

### DIFF
--- a/packages/code-link-cli/src/controller.ts
+++ b/packages/code-link-cli/src/controller.ts
@@ -705,11 +705,12 @@ async function executeEffect(
         installer: Installer | null
         fileMetadataCache: FileMetadataCache
         pendingRenameConfirmations: Map<string, PendingRenameConfirmation>
+        shutdown: () => Promise<void>
         userActions: PluginUserPromptCoordinator
         syncState: SyncState
     }
 ): Promise<SyncEvent[]> {
-    const { config, hashTracker, installer, fileMetadataCache, pendingRenameConfirmations, userActions, syncState } =
+    const { config, hashTracker, installer, fileMetadataCache, pendingRenameConfirmations, shutdown, userActions, syncState } =
         context
 
     switch (effect.type) {
@@ -1034,6 +1035,14 @@ async function executeEffect(
         }
 
         case "SYNC_COMPLETE": {
+            const shutdownIfOneOffSync = async () => {
+                if (!config.once) return false
+
+                status("Sync complete, exiting...")
+                await shutdown()
+                return true
+            }
+
             const wasDisconnected = wasRecentlyDisconnected()
 
             // Notify plugin that sync is complete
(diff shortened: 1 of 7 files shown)


packages/code-link-cli/src/controller.ts
```diff
@@ -1047,7 +1048,12 @@ async function executeEffect(
                     success(
                         `Reconnected, synced ${pluralize(effect.totalCount, "file")} (${effect.updatedCount} updated, ${effect.unchangedCount} unchanged)`
                     )
-                    status("Watching for changes...")
+                    if (config.once) {
+                        status("Initial sync complete, exiting...")
+                        await shutdown()
+                    } else {
+                        status("Watching for changes...")
+                    }
```
huntercaron: Since this code path is only for reconnections, Im not sure if it will ever happen?

packages/code-link-cli/src/controller.ts
```diff
@@ -1078,7 +1084,12 @@ async function executeEffect(
                 tryGitInit(config.projectDir)
             }
 
-            status("Watching for changes...")
+            if (config.once) {
+                status("Initial sync complete, exiting...")
```
huntercaron:
```suggestion
                status("Sync complete, exiting...")
```

Nit, saying initial redundant if its the only sync that will happen

packages/code-link-cli/src/controller.ts
huntercaron: If keeping both, might be nice to make it a small inline function

iamakulov: yeah i think my model just added it for completeness (and i didn’t oppose because it made sense)

dropping is easier, i’ll do

packages/code-link-cli/src/controller.ts
iamakulov: good one!

packages/code-link-cli/src/helpers/connection.test.ts
```diff
@@ -67,6 +67,37 @@ describe("initConnection", () => {
...
+
+        const closeReceived = new Promise<number>(resolve => {
+            client.once("close", code => {
+                resolve(code)
+            })
+        })
+
+        connection.close()
+
+        await expect(closeReceived).resolves.toBe(1005)
```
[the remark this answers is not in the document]
iamakulov: 1005 is standard when a status code is not provided, so I do not believe it will be flaky. But [1001](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) is a good one, will do that.

packages/code-link-cli/src/controller.ts
iamakulov: changed my mind, made it an inline function :) i don’t want to think whether a reconnect might happen before a connect succeeds (there are timers, etc), and especially whether that might change in the future

huntercaron: works for me!

Issue #510, opened 2026-01-08
Add time to Airtable date & time fields


madebyisaacr:
When generating fields in the Airtable plugin, [date & time](https://airtable.com/developers/web/api/field-model#dateandtime) fields should have `displayTime` enabled on the Framer date field.

Make sure this also works with formulas that return dates.

PR #601, opened 2026-03-21
Add time to Airtable date time fields

madebyisaacr:
### Description

This pull request adds displayTime support to the Airtable plugin. Based on https://github.com/framer/plugins/pull/600 by @mvanhorn, with a fix for the displayTime value not being used when creating fields and a "Date & Time" field type label to differentiate from regular date fields.

image

image

Closes https://github.com/framer/plugins/issues/510

### Changelog

- Add time support to Date & Time, Last Modified Time, and Created Time field types.

### Testing

- [x] Import an Airtable base with a date column with time enabled, a date column with time disabled, a created time column, and a last modified time column. Verify that all of those fields except the date column with time disabled have time enabled in the Framer collection.

### DIFF
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -1,4 +1,4 @@
-import type { Field, ManagedCollection, ManagedCollectionField, ManagedCollectionFieldInput } from "framer-plugin"
+import type { ManagedCollection, ManagedCollectionField, ManagedCollectionFieldInput } from "framer-plugin"
 import { FramerPluginClosedError, framer, useIsAllowedTo } from "framer-plugin"
 import { memo, useEffect, useMemo, useState } from "react"
 import type { DataSource } from "./data"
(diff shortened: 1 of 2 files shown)


plugins/airtable/src/FieldMapping.tsx
```diff
@@ -465,9 +448,43 @@ function processFields(fields: PossibleField[]): ManagedCollectionField[] {
...
+        case "number":
+            return "Number"
+        case "enum":
+            return "Option"
+        case "file":
+            return "File"
+        case "array":
+            return "Gallery"
+        default:
+            return "Unknown Type"
```
triozer, line 488: Ideally this would use ‎`assertNever`, so it acts as an exhaustive check when adding new field types.

```suggestion
            assertNever()
```

madebyisaacr: It won't let you add assertNever here. I think it's fine as it is - if a new field type is added in the future, the label being missing would be caught pretty quickly.

image

triozer: Ah yeah, `type` is a string! Okay, fine

PR #608, opened 2026-03-25
Add enum field support to Google Sheets sync

kaloyanvi:
## Summary
- add enum as a supported Google Sheets field type and derive enum cases from sheet values during field mapping
- refresh enum field definitions during sync so newly introduced values are recognized by the collection
- tighten the field-mapping page types to avoid unsafe field mutations and fix local type errors

## QA
- [ ] Import a sheet column as an enum field and confirm the generated cases match the unique non-empty cell values
- [ ] Re-sync after adding a new enum value in Google Sheets and confirm the collection field cases are updated before item sync
- [ ] Verify existing non-enum field imports still behave the same

### DIFF
--- a/plugins/google-sheets/src/sheets.ts
+++ b/plugins/google-sheets/src/sheets.ts
@@ -631,14 +631,14 @@ export async function syncSheet({
             const cases = mergeEnumCases(field.cases, getEnumCasesForColumn(rows, colIndex))
             if (cases.length === 0) {
                 throw new Error(
-                    `Enum field "${field.name}" requires at least one non-empty value before it can be synced.`
+                    `Option field "${field.name}" requires at least one non-empty value before it can be synced.`
                 )
             }
 
             if (!areEnumCasesEqual(field.cases, cases)) {
                 if (!canSetFields) {
                     throw new Error(
-                        "This sheet has enum columns with values that are not configured on the collection, " +
+                        "This sheet has option columns with values that are not configured on the collection, " +
                             "but the plugin does not have permission to update collection fields."
                     )
                 }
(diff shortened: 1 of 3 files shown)


plugins/google-sheets/src/sheets.ts
```diff
@@ -600,6 +616,51 @@ export async function syncSheet({
...
+        const updatedFields = fields.map(field => {
+            if (field.type !== "enum") return mapFieldToFramer(field)
+
+            const colIndex = uniqueHeaderRowNames.indexOf(field.id)
+            if (colIndex === -1) return mapFieldToFramer(field)
+
+            const cases = mergeEnumCases(field.cases, getEnumCasesForColumn(rows, colIndex))
+            if (cases.length === 0) {
+                throw new Error(
+                    `Enum field "${field.name}" requires at least one non-empty value before it can be synced.`
```
madebyisaacr:
```suggestion
                    `Option field "${field.name}" requires at least one non-empty value before it can be synced.`
```

plugins/google-sheets/src/sheets.ts
```diff
@@ -600,6 +616,51 @@ export async function syncSheet({
...
+                )
+            }
+
+            if (!areEnumCasesEqual(field.cases, cases)) {
+                if (!canSetFields) {
+                    throw new Error(
+                        "This sheet has enum columns with values that are not configured on the collection, " +
```
madebyisaacr:
```suggestion
                        "This sheet has option columns with values that are not configured on the collection, " +
```

plugins/google-sheets/src/sheets.ts
```diff
@@ -855,5 +916,18 @@ function mapFieldToFramer(field: SheetCollectionFieldInput): ManagedCollectionFi
         }
     }
 
+    if (field.type === "enum") {
+        assert(field.cases && field.cases.length > 0, `Enum field "${field.name}" requires at least one case`)
```
madebyisaacr:
```suggestion
        assert(field.cases && field.cases.length > 0, `Option field "${field.name}" requires at least one case`)
```

madebyisaacr (changes requested):
These errors are shown as notifications, so use "Option" instead of "Enum" to align with the field type name.

madebyisaacr (changes requested):
Looks good mostly, but I found two bugs:

- Options are not removed when they are no longer in use. For example, I have a sheet with 4 items. I changed the last one from enraged to inspired, and inspired was added but enraged was not removed despite none of the items using it.

Image

- If you clear an enum value in sheets, it stays the same in Framer after syncing. It should create an option with an empty name (or something like “None”) instead of skipping them.

PR #617, opened 2026-04-03
Add option field to Google Sheets

madebyisaacr:
### Description

This pull request adds the option field type to Google Sheets.

Originally I was working on finishing https://github.com/framer/plugins/pull/608 but after fixing the bugs, the code was too complex and the logic was hard to follow. So I started over from scratch with a simpler solution.

Notes:
- If there are any empty cells, a "None" option is added as the first case.
- Cases are sorted alphabetically.
- Cases are updated on each sync.

Slack thread: https://framer-team.slack.com/archives/C06L5H5ADK2/p1774388000311629

### Changelog

- Added option field type.

### Testing

Testing sheet - I used the status column as an option field: https://docs.google.com/spreadsheets/d/1qfTkbZqlAzmGmW3SpaYASdu8qWKmKFrxGCvO8YIdblo/edit?gid=0#gid=0

- [x] Import a sheet with an option field
- [x] Edit a cell and clear a cell in the column
- [x] Re-sync the sheet and make sure the option cases were updated

### DIFF
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -29,6 +29,7 @@ const fieldTypeOptions: FieldTypeOption[] = [
     { type: "color", label: "Color" },
     { type: "boolean", label: "Toggle" },
     { type: "number", label: "Number" },
+    { type: "enum", label: "Option" },
     { type: "file", label: "File" },
 ]
 
@@ -322,7 +323,7 @@ export function MapSheetFieldsPage({
                     <button
                         disabled={!isAllowedToManage}
                         title={isAllowedToManage ? undefined : "Insufficient permissions"}
-                        className="whitespace-nowrap inline-block"
+                        className={cx("whitespace-nowrap", !isPending && "inline-block")}
                     >
                         {isPending ? <div className="framer-spinner" /> : `Import from ${sheetTitle}`}
                     </button>
--- a/plugins/google-sheets/src/pages/MapSheetFields.tsx
+++ b/plugins/google-sheets/src/pages/MapSheetFields.tsx
@@ -240,6 +240,7 @@ export function MapSheetFieldsPage({
             ignoredColumns: Array.from(disabledColumns),
             slugColumn,
             lastSyncedTime: getLastSyncedTime(pluginContext, slugColumn),
+            configureFields: true,
         })
     }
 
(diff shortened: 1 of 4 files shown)


plugins/google-sheets/src/pages/MapSheetFields.tsx
```diff
@@ -322,7 +323,7 @@ export function MapSheetFieldsPage({
                     <button
                         disabled={!isAllowedToManage}
                         title={isAllowedToManage ? undefined : "Insufficient permissions"}
-                        className="whitespace-nowrap inline-block"
+                        className={cx("whitespace-nowrap", !isPending && "inline-block")}
```
madebyisaacr, line 327: This is a fix for the loading spinner being left aligned instead of centered. Not related to the other changes in this PR.

plugins/google-sheets/src/sheets.ts
```diff
@@ -576,6 +649,13 @@ export async function syncSheet({
     const [headerRow, ...rows] = sheet.values
 
     const uniqueHeaderRowNames = generateUniqueNames(headerRow)
+    const enrichedFields = enrichFieldsWithEnumCases(fields, uniqueHeaderRowNames, rows)
+    const needsFieldSchemaUpdate = !configureFields || enrichedFields.some(field => field.type === "enum")
```
kaloyanvi: `configureFields` is always false since we never pass it, so this const always evaluates to true.

plugins/google-sheets/src/pages/MapSheetFields.tsx
kaloyanvi, line 1: do we need to pass `configureFields: true` in the `onSubmit` handler on L235?

plugins/google-sheets/src/sheets.ts
```diff
@@ -306,6 +315,11 @@ export interface SyncMutationOptions {
     ignoredColumns: string[]
     colFieldTypes: VirtualFieldType[]
     lastSyncedTime: string | null
+    /**
+     * When false (e.g. sync-only mode), schema is only applied when enum fields need refreshed cases.
+     * When true (default), collection fields are always updated from the mapping.
+     */
+    configureFields?: boolean
```
kaloyanvi: lets make this non optional and add the default value as per the tsdoc here?

kaloyanvi (commented):
Left a few comments, as a whole it looks good if we fix those 💪🏻

madebyisaacr:
@kaloyanvi I fixed `configureFields`, can you check again?

kaloyanvi (approved):
Nice one!

PR #621, opened 2026-04-08
cleanup: use framer-spinner class in Unsplash

madebyisaacr:
### Description

This pull request replaces the spinner component in Unsplash with the built-in `framer-spinner` class. This change also makes the image loading spinner smaller to match other spinners in Framer, but if there's a reason to use the larger spinner size, that can be changed.

### Changelog

- Updated loading spinner style.

### Testing

- [x] Insert an image and look at the spinner. Verify that the spinner is white in both light and dark themes.
- [x] Click the random image button and look at the spinner


huntercaron (approved):
nice!

PR #628, opened 2026-04-17
Truncate button text in Notion and Airtable

madebyisaacr:
### Description

This pull request adds text truncation to field name buttons in Notion and Airtable plugins, and adds titles to the buttons for tooltips. https://github.com/framer/plugins/pull/605 did this for Google Sheets.

| Before | After |
| --- | --- |
| image | image |

### Changelog

- Improved text truncation on field name buttons.

### Testing

- [ ] Long column names are truncated with ellipsis

### DIFF
--- a/plugins/airtable/src/FieldMapping.tsx
+++ b/plugins/airtable/src/FieldMapping.tsx
@@ -8,7 +8,7 @@ import { isCollectionReference } from "./utils"
 
 function ChevronIcon() {
     return (
-        <svg xmlns="http://www.w3.org/2000/svg" width="8" height="16" aria-label="Chevron">
+        <svg xmlns="http://www.w3.org/2000/svg" width="8" height="16" aria-hidden="true">
             <path
                 d="M 3 11 L 6 8 L 3 5"
                 fill="transparent"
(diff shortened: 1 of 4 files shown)


plugins/airtable/src/FieldMapping.tsx
```diff
@@ -8,8 +8,7 @@ import { isCollectionReference } from "./utils"
 
 function ChevronIcon() {
     return (
-        <svg xmlns="http://www.w3.org/2000/svg" width="8" height="16">
-            <title>Chevron</title>
+        <svg xmlns="http://www.w3.org/2000/svg" width="8" height="16" aria-label="Chevron">
```
huntercaron: I think having a title on the chevron probably isn't helpful at all, I think aria-hidden would be better, and adding a title on the the actual interactive elements.

huntercaron (approved):
One nit re: the title changes, otherwise LGTM ty!

plugins/airtable/src/FieldMapping.tsx
madebyisaacr: Added `aria-hidden` to the chevrons.

PR #633, opened 2026-04-24
Refactor Code Link sync runtime for easier logic tracing

huntercaron:
## Summary
This refactor makes Code Link sync easier to trace by flattening the core path instead of adding more layers:

`SyncEvent -> transition -> Effect -> applyEffect -> SyncRuntime/SyncMemory`

The controller stays as the single orchestration surface, but the race-sensitive state now has clear homes:

- `SyncRuntime` owns lifecycle state: workspace paths, active connection, pending renames, active prompt sessions, disconnect UI, installer, and deferred prompt-sensitive state.
- `SyncMemory` owns file sync truth: agreed metadata, inbound content echoes, delete tombstones, and normalized path handling.
- `Effect` remains the flat controller plan step. The old intermediate execution layers (`EffectResult`, `RuntimeOp`, `SendIntent`, prompt refresh flags) are gone.
- `applyEffect` now performs each named side effect directly, so send/write/delete/prompt behavior can be read in one place.
- Leftover tiny scaffolding files were collapsed so the review path does not bounce through empty helpers.

## What changed
- Replaced blocking prompt flows with session-based sync events so the serial queue never waits on a user decision.
- Made conflict prompts a live unresolved-conflicts view instead of queued modal snapshots.
- Conflicts update in place as local or remote content changes, disappear when both sides converge, and clear the modal automatically when no conflicts remain.
- Delete prompts can merge additional pending deletes while stale path decisions are invalidated and ignored.
- Delete prompt paths that are already missing in Framer are immediately acknowledged back to the CLI, so hidden prompt entries do not get stranded.
- Stale or unknown prompt sessions are rejected without mutating runtime state.
- Kept unrelated files syncing while a prompt is visible.

## Race handling
The important sync commits are now success-driven and visible in the relevant `applyEffect` case:

- Remote writes arm content echo before disk write, commit metadata only after a successful write, and roll back the echo on failure.
- Remote deletes arm tombstones before unlink, record synced delete only after a successful or already-missing delete, and roll back on failure.
- Local file sends arm echoes and emit file-up/install work only after socket send success.
- Local delete sends record synced delete only after socket send success.
- Rename sends register pending renames only after socket send success, with deterministic ack/error cleanup.
- Once-mode conflict resolution clears prompt state and flushes metadata before sync-complete finalization can shut the CLI down.

## Plugin/UI updates
- Added session-aware prompt messages for clearing conflict and delete UI.
- Updated plugin reducer/message handling so conflict and delete prompts refresh instead of accumulating stale modal state.
- Scoped unguarded prompt-clearing actions so they cannot close unrelated prompt UI.
- Made plugin echo checks presence-aware and exact-content based so unseen or hash-colliding content is not mistaken for an echo.
- Preserved the existing plugin connection/reconnect/idle behavior.

## Testing
Added focused coverage using real collaborators where practical: real `SyncRuntime`, real `SyncMemory`, temp directories, metadata cache, and the existing in-memory websocket shape.

Covered scenarios include:

- send failure does not arm echoes, record deletes, or register pending renames
- write/delete failure rolls back guards and does not commit metadata
- live conflict prompt update and automatic clearing on convergence
- conflict convergence metadata commit
- delete prompt path invalidation and stale response rejection
- plugin acknowledgement for delete prompt paths already missing in Framer
- presence-aware and exact plugin echo suppression for empty/hash-colliding content
- scoped plugin prompt clearing for unrelated active prompts
- once-mode shutdown ordering after conflict resolution
- plugin reducer/message handling for session-aware prompt updates and clears

Verified:

- `yarn tsc --project packages/code-link-cli/tsconfig.json --noEmit`
- `yarn tsc --project plugins/code-link/tsconfig.json --noEmit`
- `yarn tsc --project packages/code-link-shared/tsconfig.json --noEmit`
- `yarn biome check ...`
- focused Code Link vitest suite: 14 files passed, 148 tests passed
- broad CLI/plugin vitest suite: 18 files passed, 190 tests passed


packages/code-link-cli/src/controller.ts
```diff
@@ -351,51 +207,57 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
+                    content: event.file.content,
+                    modifiedAt: event.file.modifiedAt,
+                })
+                return { state, effects }
+            }
+
+            if (
+                state.internalPhase === "snapshot_processing" ||
+                state.internalPhase === "handshaking" ||
+                state.internalPhase === "conflict_resolution"
```
niekert: Just a Q: if an event with remote file changes comes in while `snapshot_processing` or some of the other states, the event is ignored

How does it see those changes after transitioning from the current state?

packages/code-link-cli/src/controller.ts
```diff
@@ -427,110 +306,38 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
+                fileNames: event.fileNames,
+            })
             return {
-                state: {
-                    ...rest,
-                    mode: "watching",
-                },
+                state:
+                    state.internalPhase === "disconnected"
+                        ? state
```
niekert: nit: is this disconnected check like necessary? How could it be receiving transition events while disconnected? Or am I misunderstanding its meaning

packages/code-link-cli/src/controller.ts
```diff
@@ -540,8 +347,13 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
             const { kind, relativePath, content } = event.event
 
             // Only process changes in watching mode
-            if (state.mode !== "watching") {
-                effects.push(log("debug", `Ignoring watcher event in ${state.mode} mode: ${kind} ${relativePath}`))
+            if (state.internalPhase !== "watching") {
```
niekert: silly naming nit but can't this just be `state.phase` everywhere?

packages/code-link-cli/src/controller.ts
```diff
@@ -693,440 +536,511 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
-        hashTracker: ReturnType<typeof createHashTracker>
-        installer: Installer | null
-        fileMetadataCache: FileMetadataCache
-        pendingRenameConfirmations: Map<string, PendingRenameConfirmation>
-        shutdown: () => Promise<void>
-        userActions: PluginUserPromptCoordinator
-        syncState: SyncState
+// Apply: the only place that mutates or does I/O
+
+type LogLevel = "info" | "debug" | "warn" | "success" | "status"
```
niekert: nit: `LogLevel` already exists in logger as an enum, maybe just make it a string literal type there so you don't have duplicate LogLevels definitions

packages/code-link-cli/src/controller.ts
```diff
@@ -693,440 +536,511 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
-                    projectHash: config.projectHash,
-                    projectName,
-                    explicitDirectory: config.explicitDirectory,
-                })
-                config.projectDir = directoryInfo.directory
-                config.projectDirCreated = directoryInfo.created
+async function writeFiles(
+    files: Extract<Effect, { type: "WRITE_FILES" }>["files"],
+    ctx: ApplyCtx,
+    options: { silent?: boolean; skipEcho?: boolean } = {}
```
niekert: could be nice to add a comment what `skipEcho` means and why its useful, not sure I understand 🙏

packages/code-link-cli/src/controller.ts
```diff
@@ -693,440 +536,511 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
+        ctx.runtime.clearConflictPromptFiles(
+            prompt.session,
+            prompt.conflicts.map(conflict => conflict.fileName)
+        )
+        warn("Failed to send conflict prompt")
+    }
+}
+
+async function applyConflictChange(
+    change: ReturnType<SyncRuntime["updateActiveConflictLocal"]>,
```
niekert: nit:
```suggestion
    change: ConflictPromptChange,
```

nbd but the current type kinda couples it closely to that method when the type is already an exported interface

packages/code-link-cli/src/controller.ts
```diff
@@ -693,440 +536,511 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
+            })
+        } else if (relativeDirectory) {
+            logs.push({
+                level: "success",
+                message: `Synced into ${relativeDirectory} (${pluralize(effect.updatedCount, "file")} updated, ${effect.unchangedCount} unchanged)`,
+            })
+        } else {
+            logs.push({
+                level: "success",
+                message: `Synced ${pluralize(effect.totalCount, "file")} (${effect.updatedCount} updated, ${effect.unchangedCount} unchanged)`,
```
niekert: maybe too nitpicky but this is like 20+ lines of code for pushing a single success message to logs. A single `logs.push({ level "succes", message: syncSuccessMessage(runtime, effect)})` could make the function a bit less distracting. 

You can also `emitLog` directly or just call `success()` directly even instead of keeping an array

packages/code-link-cli/src/controller.ts
huntercaron: I remember adding that and already forget why so clearly a good point

packages/code-link-cli/src/controller.ts
```diff
@@ -693,440 +536,511 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
+                            content: conflict.remoteContent!,
+                            modifiedAt: conflict.remoteModifiedAt,
+                        })),
+                    ctx,
+                    { silent: true }
+                )
+                await deleteFiles(
+                    conflicts.filter(conflict => conflict.remoteContent === null).map(conflict => conflict.fileName),
+                    ctx
+                )
```
niekert:
```
const filesToDelete = []
const fileToWrite = []
for (const conflict of conflicts) {
    if (conflict.remoteContent === null) {
        filesToDelete.push(conflict.fileName)
        continue
    } 

    fileToWrite.push({ name: conflict.fileName, content: conflict.remoteContent!, modifiedAt: conflict.remoteModifiedAt })
}

await Promise.all([writeFiles(filesToWrite), deleteFiles(filesToDelete)])
```

imo something like this would be a big readability win

packages/code-link-cli/src/controller.ts
```diff
@@ -693,440 +536,511 @@ function transition(state: SyncState, event: SyncEvent): { state: SyncState; eff
...
-            const logFn = logFns[effect.level]
-            logFn(effect.message)
+        case "LOG":
+            emitLog({ level: effect.level, message: effect.message })
             return []
-        }
     }
 }
 
+async function executeEffect(effect: Effect, ctx: ApplyCtx): Promise<SyncEvent[]> {
```
niekert: nit: all these callers can just call `applyEffect` directly right?

packages/code-link-cli/src/runtime.ts
```diff
@@ -0,0 +1,439 @@
...
+}
+
+export interface ReadonlyFileMetadataCache {
+    get(fileName: string): FileSyncMetadata | undefined
+    has(fileName: string): boolean
+    size(): number
+    getPersistedState(): Map<string, PersistedFileState>
+}
+
+export interface ReadonlyRuntime {
```
niekert: Is this used besides being re-exported?

packages/code-link-cli/src/runtime.ts
```diff
@@ -0,0 +1,439 @@
...
+        this.memory.clearContentEcho(filePath)
+    }
+
+    armExpectedDeleteEcho(filePath: string): void {
+        this.memory.armExpectedDeleteEcho(filePath)
+    }
+
+    clearExpectedDeleteEcho(filePath: string): void {
+        this.memory.clearExpectedDeleteEcho(filePath)
+    }
```
niekert: This memory facade just pipes through memory methods directly which feels like its mostly plumbing?

It feels like calling `runtime.memory.doXX` in callers directly is fine since we're plumbing through almost all methods here. Unless you wanna avoid exposing some internal state?

plugins/code-link/src/plugin-base.ts
```diff
@@ -0,0 +1,84 @@
+import { normalizeCodeFilePathWithExtension } from "@code-link/shared"
+
+/**
+ * Plugin-side peer state: one map of normalized path → latest known content (snapshot + echo suppression).
+ */
+export class PluginBase {
```
niekert: nit imo this naming is a bit weird, its also only ever used within CodeFilesAPI, so i kinda feel like it should just be part of that class instead of introducing inheritance

huntercaron: thats true I had a similar thought, will fixy

packages/code-link-cli/src/controller.ts
huntercaron: oh truee thats much nicer thank ya sir

niekert (approved):
LGTM. Clearly an improvement in architecture 👏 It was a lil difficult to fully trace the behavioral changes but I did quite a bit of QA too

PR #638, opened 2026-04-30
Update framer-plugin 

huntercaron:
### Description

This pull request bumps `framer-plugin` to `3.11.0-alpha.12` for all workspaces under `plugins/`, and runs `yarn dedupe`

### Changelog

- Update Inter with support for alternatives

### Testing

- [ ] Smoke test a few Plugins
  - [ ] Verify they open and look correct
  - [ ] Verify Inter font loads correctly


plugins/ascii/package.tgz
madebyisaacr, line 1: What is this package.tgz file for? I unzipped it and it appears to be a copy of the ASCII plugin source code.

Image

huntercaron: Ah thank you, I must've run the wrong pack command once. removed it now

madebyisaacr (approved):

PR #639, opened 2026-04-30
Remove redundant Google Fonts Inter imports

huntercaron:
### Description

Some plugins were additionally importing Inter from Google Fonts in addition to our new self hosted Inter. This was breaking font alternatives.

This removes those redundant imports from `code-versions`, `global-search`, `hubspot`, and `unsplash`, plus an orphan `preconnect` to `fonts.googleapis.com` in `phosphor` that wasn't paired with any actual font load.

### Changelog

- Remove redundant font imports

### Testing

- [x] Each affected plugin still renders with Inter
  - [x] Open `code-versions` plugin and verify text uses Inter
  - [x] Open `global-search` plugin and verify text uses Inter
  - [x] Open `hubspot` plugin and verify text uses Inter
  - [x] Open `unsplash` plugin and verify text uses Inter
- [x] `phosphor` plugin still loads and renders correctly


plugins/global-search/src/styles.css
```diff
@@ -36,6 +35,14 @@
     --item-height: 30px;
 }
 
+[data-framer-styles="prerelease"] {
+    font-feature-settings:
+        "cv01" 1,
+        "cv05" 1,
+        "cv09" 1,
+        "cv11" 1;
+}
```
niekert, line 46: Could we add a comment what these do?

niekert (approved):
👏

plugins/global-search/src/styles.css
huntercaron: done!

PR #640, opened 2026-05-05
Remove aveta dependency from Google Search Console to avoid crasher

niekert:
### Description

This was causing some issues with Vite 8 production bundling, and should never have been a dependency in the first place

### Changelog

- Fixes a crasher when visitor numbers are formatted

### Testing

- [x] No more crashers
  - On a production build of the plugin
  - The plugin no longer crashes in the Performance component


huntercaron (approved):
seems chill!

PR #643, opened 2026-05-06
Add Code Link npm strategy modes

huntercaron:
### Description

Adds Code Link npm strategy handling so unsupported npm can run in `acquire-types` mode or `package-manager` mode. Package-manager mode reads `codeLinkNpmStrategy`/lockfiles, skips ATA writes into `node_modules`, and asks the plugin for Framer dependency versions to keep `package.json` dependencies current.

### Changelog

- Improve npm dependency type-fetching strategy

### Testing

- [ ] QA `acquire-types` mode
  - Run the Code Link plugin locally from `plugins/code-link`.
  - Use the alpha CLI with `--unsupported-npm=acquire-types` and verify unsupported package types are acquired.
- [ ] QA `package-manager` mode
  - Run the Code Link plugin locally from `plugins/code-link`.
  - Use the alpha CLI with `--unsupported-npm=package-manager`, verify `package.json` updates, then run your package manager.


packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -270,6 +294,126 @@ export class Installer {
...
+    private async collectPackageManagerPackageNames(): Promise<string[]> {
+        const packageNames = new Set([...CORE_LIBRARIES, ...PACKAGE_MANAGER_DEV_DEPENDENCIES])
+        await this.addPackageNamesFromDirectory(path.join(this.projectDir, "files"), packageNames)
+        return [...packageNames]
+    }
+
+    private async addPackageNamesFromDirectory(directory: string, packageNames: Set<string>): Promise<void> {
+        let entries: Dirent[]
+        try {
+            entries = await fs.readdir(directory, { withFileTypes: true })
```
iamakulov: **[potential bug]** Does this need `recursive: true`? (Code files can be in subdirectories.)

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -270,6 +294,126 @@ export class Installer {
...
+                    return
+                }
+
+                if (!entry.isFile() || JSON_EXTENSION_REGEX.test(entry.name)) {
+                    return
+                }
+
+                try {
+                    const content = await fs.readFile(entryPath, "utf-8")
+                    for (const imported of extractImports(content).filter(i => i.type === "npm")) {
```
iamakulov: **[simplicity]** Up to you whether to drive-by this but I noticed we only ever do `extractImports().filter(i => i.type === "npm")`. We can drop URL imports and simplify.

iamakulov: **[bug (pre-existing)]** We can also switch from fragile regexes (eg they don’t work with imports like

```ts
import Foo, {
  useFoo,
} from "foo"
```

) to https://www.npmjs.com/package/parse-imports.

packages/code-link-cli/src/helpers/installer.ts
iamakulov, line 184:
```suggestion
/**
    * Fire-and-forget processing of a component file to fetch missing types or add `package.json` dependencies.
    * JSON files are ignored.
    */
```

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -228,11 +245,17 @@ export class Installer {
 
         if (allImports.length === 0) return
 
+        if (this.npmStrategy === "package-manager") {
+            await this.enqueuePackageJsonRefresh(allImports.map(imp => getBasePackageName(imp.name)))
+            return
+        }
```
iamakulov: **[simplicity]** I think this would read much simpler if we rearrange this into:

```ts
    private async processImports(fileName: string, content: string): Promise<void> {
        const allImports = extractImports(content).filter(i => i.type === "npm")

        if (allImports.length === 0) return

        if (this.npmStrategy === "package-manager") {
            await this.enqueuePackageJsonRefresh(allImports.map(imp => getBasePackageName(imp.name)))
        } else if (this.npmStrategy === "acquire-types") {
            await this.acquireTypes(allImports)
        } else {
            const supportedImports = allImports.filter(i => this.isSupportedPackage(i.name))
            const unsupportedImports = allImports.filter(i => this.isSupportedPackage(i.name))
            if (unsupportedImports.length > 0) {
                debug(`Skipping unsupported packages: ${unsupported.join(", ")} (use --unsupported-npm to enable)`)
            }
            await acquireTypes(supportedImports)
        }
```

WDYT?

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -270,6 +294,126 @@ export class Installer {
...
+                } catch {
+                    // Best effort; file contents can change while the watcher is active.
+                }
+            })
+        )
+    }
+
+    private async enqueuePackageJsonRefresh(packageNames: string[]): Promise<void> {
+        const missingPackageNames = packageNames
+            .map(name => getBasePackageName(name))
```
iamakulov: **[simplicity]** This is weird; two parts of this code tell different stories:

- The param name (`packageNames`) tells that the function accepts sanitized package names (`foo`, `@bar/baz`)
- But `.map(name => getBasePackageName(name))` tells that the function accepts raw import paths that need to be sanitized into package names (`@bar/baz/dist/index.js`)

Looking at the usage, it seems the first story is true. Could we/should we simplify so that the story is consistent?


```suggestion
    private async enqueuePackageJsonRefresh(packageNames: string[]): Promise<void> {
        const missingPackageNames = packageNames
```

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -270,6 +294,126 @@ export class Installer {
...
+            return this.packageJsonRefreshPromise
+        }
+
+        this.packageJsonRefreshPromise = this.packageJsonRefreshPromise
+            .then(async () => {
+                await this.refreshPackageJsonFromPlugin(missingPackageNames)
+            })
+            .catch((err: unknown) => {
+                warn("Could not refresh package.json dependency versions", err)
+            })
```
iamakulov: **[simplicity]** If I’m honest, I’m struggling to follow what happens here. We `.filter()` but also modify the global state (which this `.filter()` depends on) in the process? We return a promise but also do `promise = promise.then`?

While I figured out what all of this does after staring into it for a few minutes, could I nudge you to find a KISS version of this? Eg:

- why store processed package names in the global state if we could treat `package.json` as a source of truth? not only this is less complex, this also eliminates a whole class of bugs when the `package.json` and local state go out of sync
- why do the whole enqueue thing? do we really need it? or, if we do, is there a simpler primitive we can use here? (debounce? an existing queue from npm?)

  **[bug]** something in the queueing logic leads to a bug where we reliably log the same message twice (repro: [this project](https://framer.com/projects/Tedious-Benefits--cfnzO4pofYd4337aOwNt-f5JCQ?node=augiA20Il) + this command: `npx framer-code-link@alpha --unsupported-npm=package-manager A8ekr3Wn`)), and I don’t even want to model the whole queueing thing in my head to figure out where this comes from

  Image

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -270,6 +294,126 @@ export class Installer {
...
+        }
+
+        if (!changed) {
+            return
+        }
+
+        pkg.dependencies = dependencies
+        pkg.devDependencies = devDependencies
+        await fs.writeFile(packagePath, JSON.stringify(pkg, null, 4))
+        success("Updated dependencies. Run your package manager to install them.")
```
iamakulov:
```suggestion
        success("Updated project dependencies in package.json. Run your package manager to install them.")
```

**[dx]** thoughts? otherwise unclear what “dependencies” this refers to (i, as a user, could read this as plugin dependencies)

packages/code-link-cli/src/helpers/npm-strategy.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (packageJsonStrategy) {
+        debug(`Using npm strategy from package.json ${CONFIG_FIELD}: ${packageJsonStrategy}`)
+        return packageJsonStrategy
+    }
+
+    const detectedLockfile = await detectLockfile(projectDir)
+    if (detectedLockfile) {
+        debug(`Using npm strategy package-manager from ${detectedLockfile}`)
+        return "package-manager"
+    }
```
iamakulov: **[dx]** Is it intentional that we’re doing this now?

1\) I thought you wanted to discourage people from using `--unsupported-npm`

2\) This is pretty magical; it’s not obvious that adding a `yarn.lock` would change how `code-link` works. I’d either not do this at all, or make a log like “yarn.lock detected, syncing project’s dependencies into package.json (pass --unsupported-npm=none to disable)”

3\) This doesn’t work great; eg it doesn’t detect monorepos where `yarn.lock` is outside of the current dir

packages/code-link-cli/src/index.ts
```diff
@@ -20,6 +20,18 @@ const { version } = require("../package.json") as { version: string }
...
+    if (mode === undefined) {
+        return "acquire-types"
+    }
+
+    if (mode === "acquire-types" || mode === "package-manager") {
+        return mode
+    }
+
+    throw new InvalidArgumentError("unsupported npm mode must be 'acquire-types' or 'package-manager'")
+}
```
iamakulov: **[bug]** Don’t we also support `"none"`? (See `npm-strategy.ts`)

packages/code-link-cli/src/index.ts
```diff
@@ -39,7 +51,11 @@ program
     .option("-v, --verbose", "Enable verbose logging")
     .option("--log-level <level>", "Set log level (debug, info, warn, error)")
     .option("--dangerously-auto-delete", "Automatically delete remote files without confirmation")
-    .option("--unsupported-npm", "Allow type acquisition for unsupported npm packages")
+    .option(
+        "--unsupported-npm [mode]",
+        "Handle unsupported npm packages (acquire-types or package-manager)",
```
iamakulov: **[dx]**

```suggestion
        "Handle unsupported npm packages (none or acquire-types or package-manager)",
```

packages/code-link-cli/src/index.ts
```diff
@@ -39,7 +51,11 @@ program
     .option("-v, --verbose", "Enable verbose logging")
     .option("--log-level <level>", "Set log level (debug, info, warn, error)")
     .option("--dangerously-auto-delete", "Automatically delete remote files without confirmation")
-    .option("--unsupported-npm", "Allow type acquisition for unsupported npm packages")
+    .option(
+        "--unsupported-npm [mode]",
```
iamakulov: **[bug]** Devin says (and I confirmed this is true):

**`--unsupported-npm` without argument silently has no effect due to Commander optional-arg semantics**

When `--unsupported-npm` is used without a value argument (the backward-compatible form), Commander does not call the `parseUnsupportedNpmMode` parse function. Instead, Commander sets the option value to `true` (boolean) because `presetArg` is `undefined` and the option uses `[mode]` (optional argument). This is confirmed by Commander's source code at `node_modules/commander/lib/command.js:692-716` where `val` is `null` for optional options without an argument, and since `presetArg` is `undefined`, the parse function is skipped and `val` is set to `true`.

The boolean `true` flows into `config.npmStrategy`, passes `resolveNpmStrategy`'s truthy check (`npm-strategy.ts:11`), and reaches the Installer where it doesn't match `"acquire-types"` or `"package-manager"`, silently defaulting to `"none"` behavior. Users upgrading from the old boolean `--unsupported-npm` flag would find type acquisition for unsupported packages no longer works.

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -260,7 +283,8 @@ export class Installer {
         debug(`Processing imports for ${fileName} (${imports.length} packages)`)
 
         // Build filtered content with only supported imports for ATA
-        const filteredContent = this.allowUnsupportedNpm ? content : await this.buildFilteredImports(imports)
+        const filteredContent =
+            this.npmStrategy === "acquire-types" ? content : await this.buildFilteredImports(imports)
```
iamakulov: **[bug (pre-existing)]** This is a pre-existing bug, so just thought to note: I think we should (not necessarily in this PR) pin versions with `this.npmStrategy === "acquire-types"` too. Otherwise, a project might use `foo@4.0.0`, but we might accidentally acquire types for `foo@5.0.0`. We didn’t have an ability to pin/resolve versions in the past, but we do have it now.

(As a side note, do we know if anyone uses `this.npmStrategy === "acquire-types"`? Do we need to keep it around now that `this.npmStrategy === "package-manager"` is here? Maybe the latter one is what everyone would prefer?)

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -270,6 +294,126 @@ export class Installer {
         }
     }
 
+    private async collectPackageManagerPackageNames(): Promise<string[]> {
+        const packageNames = new Set([...CORE_LIBRARIES, ...PACKAGE_MANAGER_DEV_DEPENDENCIES])
+        await this.addPackageNamesFromDirectory(path.join(this.projectDir, "files"), packageNames)
+        return [...packageNames]
+    }
+
+    private async addPackageNamesFromDirectory(directory: string, packageNames: Set<string>): Promise<void> {
```
iamakulov: **[simplicity]** Had to pause for a moment to figure out where exactly this function adds package names. Could we / should we make it pure, as in, make it return a set rather than amend an existing set?

Then, above, we can simply do:

```diff
diff --git a/packages/code-link-cli/src/helpers/installer.ts b/packages/code-link-cli/src/helpers/installer.ts
index 4dd88e0591..b874a5db05 100644
--- a/packages/code-link-cli/src/helpers/installer.ts
+++ b/packages/code-link-cli/src/helpers/installer.ts
@@ -295,9 +295,10 @@
     }

     private async collectPackageManagerPackageNames(): Promise<string[]> {
-        const packageNames = new Set([...CORE_LIBRARIES, ...PACKAGE_MANAGER_DEV_DEPENDENCIES])
-        await this.addPackageNamesFromDirectory(path.join(this.projectDir, "files"), packageNames)
-        return [...packageNames]
+        const defaultPackageNames = new Set([...CORE_LIBRARIES, ...PACKAGE_MANAGER_DEV_DEPENDENCIES])
+        const packageNamesFromDir = await this.getPackageNamesFromDirectory(path.join(this.projectDir, "files"))
+
+        return [...defaultPackageNames.union(packageNamesFromDir)]
     }

     private async addPackageNamesFromDirectory(directory: string, packageNames: Set<string>): Promise<void> {
```

(`set.union()` is supported as of Node.js 22.)

iamakulov (commented):
Sorry, this is a big review, but a big chunk of it is about code complexity or DX. Feel free to skip those if it blocks shipping!

Some extra notes from QA/Slack:
- **[dx]** the version ranges that we set in `package.json` should be pinned, not `^`. we pin your dependency versions in framer, and the `package.json` should reflect that
- **[bug]** with [this project](https://framer.com/projects/Tedious-Benefits--cfnzO4pofYd4337aOwNt-f5JCQ?node=augiA20Il), dependency syncing simply doesn’t happen. (reproduced with `npx framer-code-link@alpha --unsupported-npm=package-manager A8ekr3Wn`.) files sync, but deps aren’t added into `package.json`. haven’t debugged why.

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -380,7 +380,7 @@ export class Installer {
 
         let changed = false
         for (const packageName of uniquePackageNames) {
-            const version = versions[packageName] ?? this.pinnedTypeVersions[packageName]
+            const version = (versions[packageName] ?? this.pinnedTypeVersions[packageName])?.replace(/^\^/, "")
```
iamakulov: **[simplicity]** why do we have to strip the `^` here? this feels like the wrong place – we have code that adds a `^` somewhere, and we’re undoing that code’s work with more code here. we can just delete the original code, I think?

(is it `unstable_getDependencyVersion` that returns a version with a `^`? then I’d def kill the `^` there. i’d expect the `getDependencyVersion` method to return a version, not a version range.)

packages/code-link-cli/src/helpers/installer.ts
iamakulov: already done! ty!

packages/code-link-cli/src/helpers/installer.ts
huntercaron: The ^ is persisted in the actual `dependencies.json` records for all Framer projects sadly, its something that is technically incorrect but was rushed through years ago.

Would you expect it here or in the Plugin API functions? fixing the source should happen holistically with the other npm fixes IMO,  has to be monkey patched somewhere in this process for now.

huntercaron: Maybe doing it in the Plugin API/FramerStudio at least makes more sense so it can later have the knowledge if the project is fixed/migrated once that happens

packages/code-link-cli/src/helpers/npm-strategy.ts
huntercaron: yeah indeed underbaked, removing

packages/code-link-cli/src/helpers/installer.ts
huntercaron: Re: bug will pin the versions everywhere yup!

I still don't believe package manager is the best flow for most people. I use acquire types! I don't need to faff with another step and don't need anything beyond the types

packages/code-link-cli/src/helpers/installer.ts
iamakulov: ohhhh! didn’t realize that’s where it comes from

great q, tbh i feel i’d strip it on the framer side then – because if this was a public api, then that’d be the only approach that feels right

huntercaron: https://github.com/framer/FramerStudio/pull/31178 I agree!

packages/code-link-cli/src/index.ts
huntercaron: changed to just undefined so its clearer, acquire types is intended to just be a name for the default

packages/code-link-cli/src/helpers/files.ts
```diff
@@ -29,44 +30,37 @@ export const DEFAULT_REMOTE_DRIFT_MS = 2000
...
-            } catch (err) {
-                warn(`Failed to read ${entryPath}:`, err)
-            }
+            files.push({
+                name: sanitizedPath,
+                content,
+                modifiedAt: stats.mtimeMs,
+            })
+        } catch (err) {
+            warn(`Failed to read ${entryPath}:`, err)
```
huntercaron, line 60: Drive by switch to readdir with recursive as mentioned in another comment!

packages/code-link-cli/src/helpers/installer.ts
huntercaron: Ended up using the TS parser instead, seems to work fine (still much better than the regex) and we had the dep in there already

huntercaron:
@iamakulov ty for all the feedback! I've simplified it quite a bit by plugging into our existing lifecycles/queues and listFiles, etc functions rather than doing all that on its own

packages/code-link-cli/src/helpers/installer.ts
iamakulov: ah maybe reverted?

packages/code-link-cli/src/helpers/installer.ts
```diff
@@ -257,40 +280,83 @@ export class Installer {
...
-        const basePkg = pkgName.startsWith("@") ? pkgName.split("/").slice(0, 2).join("/") : pkgName.split("/")[0]
+        if (!changed) {
+            return
+        }
 
-        return SUPPORTED_PACKAGES.has(basePkg)
+        pkg.dependencies = sortDependencyMap(dependencies)
+        pkg.devDependencies = sortDependencyMap(devDependencies)
+        await fs.writeFile(packagePath, JSON.stringify(pkg, null, 4))
+        success("Updated dependencies. Run your package manager to install them.")
```
iamakulov:
```suggestion
        success("Updated project dependencies in package.json. Run your package manager to install them.")
```

(i think this was done but then maybe reverted? or maybe i misread something)

thoughts? otherwise unclear what “dependencies” this refers to (i, as a user, could read this as plugin dependencies)

packages/code-link-cli/src/index.ts
```diff
@@ -39,7 +39,14 @@ program
     .option("-v, --verbose", "Enable verbose logging")
     .option("--log-level <level>", "Set log level (debug, info, warn, error)")
     .option("--dangerously-auto-delete", "Automatically delete remote files without confirmation")
-    .option("--unsupported-npm", "Allow type acquisition for unsupported npm packages")
+    .addOption(
+        new Option(
+            "--unsupported-npm [mode]",
+            "Handle unsupported npm packages (default without mode: acquire-types; modes: acquire-types, package-manager)"
```
iamakulov: had a bit of trouble parsing this; maybe this?

```suggestion
            "Handle unsupported npm packages (modes: acquire-types or package-manager; if passed without explicit mode, assumes acquire-types)"
```

or even, to explain a bit more what this does:

```suggestion
            "Handle third-party unsupported npm packages. Modes: acquire-types (prefills node_modules with type definitions for these packages), package-manager (puts the packages’ versions into package.json). If passed without explicit mode, assumes acquire-types"
```

iamakulov (approved):
Skimmed a bit less deeply in my second review, didn’t look at complexity. Looks good, left a couple of copy nits ❤️

packages/code-link-cli/src/helpers/installer.ts
huntercaron: Hmm dang sorry been having troubles with models undoing things recently

packages/code-link-cli/src/index.ts
huntercaron: Did a variation of the second one, ty!

PR #646, opened 2026-05-19
Renamer prerelease styles

madebyisaacr:
### Description

This pull request adds prerelease styles to the Renamer plugin, including new icons and tabs design.

Screenshots: https://www.notion.so/framer/365adf6e8c968017adafebc517001a38

### Changelog

- Updated icons
- Added icon for breakpoint layers

### Testing

- [x] Verify styles are correct with experiment on and off


plugins/renamer/src/components/LayerIcon.tsx
```diff
@@ -1,65 +1,197 @@
...
+                        />
+                    </svg>
+                )
+                break
+
+            case "ComponentInstanceNode":
+                icon = (
+                    <svg
+                        xmlns="http://www.w3.org/2000/svg"
+                        width="11"
```
Nick-Lucas, line 178: I expect cases like this are why we had 12 set as the parent element width/height (maybe the icon had a large perceived size so we dropped by 1?) but we seem to have lost that in the change

This file may be worth keeping that behaviour unless we can demonstrate it was a mistake before

plugins/renamer/src/components/Tabs.tsx
```diff
@@ -11,9 +12,20 @@ interface Props {
...
+    return isPrerelease ? (
+        <div className="prerelease-tabs">
+            <div className="prerelease-tabs-container">
+                {items.map((item, index) => (
+                    <button key={index} className={cx("prerelease-tab", item.active && "active")} onClick={item.select}>
+                        {item.label}
+                    </button>
+                ))}
+            </div>
+        </div>
```
Nick-Lucas: these could probably be conditions over the className probably rather than duplicating the structure, we already have `cx` which is designed for this need

plugins/renamer/src/App.css
```diff
@@ -250,6 +292,10 @@
     padding: 15px;
 }
 
+[data-framer-styles="prerelease"] .container {
```
Nick-Lucas, line 289: an alternative for the tabs could be to use this pattern instead, so that the new classes apply when `data-framer-styles` is set in the root

plugins/renamer/src/App.css
```diff
@@ -364,6 +410,8 @@
 
 .icon {
     flex-shrink: 0;
+    width: 12px;
+    height: 12px;
```
Nick-Lucas, line 408: Ah I see now, ignore my other comment then

plugins/renamer/src/components/LayerIcon.tsx
Nick-Lucas: Nevermind I see you just moved it to the class which makes sense: https://github.com/framer/plugins/pull/646/changes#r3272172807

plugins/renamer/src/App.css
madebyisaacr: Switched to using a selector for tab styles.

Nick-Lucas (approved):

PR #667, opened 2026-06-23
Update marketplace URL and fix CLI version

kaloyanvi:
## Description

- Update the Framer Marketplace submission URL to `https://www.framer.com/community/marketplace/plugins/` in both the CLI `pack` output and the README.
- Read the CLI `--version` from `package.json` instead of hardcoding it, so it can't drift out of sync again (it was reporting `1.1.0` while the package was at `1.2.1`).

## QA

- [x] QA
  - `yarn build` succeeds.
  - `node dist/cli.js --version` prints `1.2.2` (matches `package.json`).
  - CLI `pack` output and README both point to the new marketplace URL.

### DIFF
--- a/packages/plugin-tools/src/cli.ts
+++ b/packages/plugin-tools/src/cli.ts
@@ -36,7 +36,7 @@ program
         const dirName = path.dirname(zipFilePath)
 
         console.log(
-            `\n⚡️ ${fileName} file has been created in ${dirName} \n Submit your Plugin on the Framer Marketplace: https://www.framer.com/marketplace/dashboard/plugins/`
+            `\n⚡️ ${fileName} file has been created in ${dirName} \n Submit your Plugin on the Framer Marketplace: https://www.framer.com/community/marketplace/plugins/`
         )
     })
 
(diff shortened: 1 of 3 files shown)


packages/plugin-tools/src/cli.ts
```diff
@@ -36,7 +37,7 @@ program
         const dirName = path.dirname(zipFilePath)
 
         console.log(
-            `\n⚡️ ${fileName} file has been created in ${dirName} \n Submit your Plugin on the Framer Marketplace: https://www.framer.com/marketplace/dashboard/plugins/`
+            `\n⚡️ ${fileName} file has been created in ${dirName} \n Submit your Plugin on the Framer Marketplace: https://www.framer.com/community/marketplace/plugins/`
```
andreugordillovazquez: Small nit, but maybe “Publish your Plugin” instead of “Submit your Plugin”?

andreugordillovazquez (approved):

packages/plugin-tools/src/cli.ts
huntercaron: ++

kaloyanvi: cool, updated here and in the readme

PR #684, opened 2026-08-10
Fix Notion sync checkpoint handling

iamakulov:
This PR fixes a Notion sync bug where, when one or more items would fail to sync, we’d still treat the sync as successful and advance the checkpoint. The result of this would be that the failed items would stay outdated but would be treated as fresh (until you edit them again).

QAed by blocking one of the requests that the plugin makes to the Notion API to fetch a record.

### Changelog

- Fixed Notion sync so failed items are retried instead of being incorrectly marked as up to date.

### DIFF
--- a/plugins/notion/src/data.ts
+++ b/plugins/notion/src/data.ts
@@ -114,6 +114,14 @@ export interface SyncError {
     error: unknown
 }
 
+export type SyncResult =
+    | { status: "completed" }
+    | {
+          status: "completed-with-errors"
+          succeeded: number
+          failed: number
+      }
+
 export async function syncCollection(
     collection: ManagedCollection,
     dataSource: DataSource,
@@ -123,7 +131,7 @@ export async function syncCollection(
     lastSynced: string | null,
     existingFields?: readonly ManagedCollectionFieldInput[],
     onProgress?: (progress: SyncProgress) => void
-) {
+): Promise<SyncResult> {
     const fieldsById = new Map(fields.map(field => [field.id, field]))
     const contentFieldEnabled = fieldsById.has(pageContentProperty.id)
     const reportProgress = (p: { current: number; total: number; hasFinishedLoading: boolean }) =>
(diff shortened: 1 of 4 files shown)


plugins/notion/src/data.ts
```diff
@@ -371,16 +375,28 @@ export async function syncCollection(
...
+    if (syncErrors.length === 0) {
+        pluginDataUpdates.push(collection.setPluginData(PLUGIN_KEYS.LAST_SYNCED, syncStartedAt))
+    }
+
+    await Promise.all(pluginDataUpdates)
+
+    if (syncErrors.length > 0) {
+        const pluralSuffix = syncErrors.length === 1 ? "" : "s"
+        throw new Error(`Failed to sync ${syncErrors.length} item${pluralSuffix}. Please try again.`)
+    }
```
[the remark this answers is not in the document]
iamakulov: meh

plugins/notion/src/data.ts
```diff
@@ -114,6 +114,15 @@ export interface SyncError {
     error: unknown
 }
 
+export type SyncResult =
+    | { status: "success" }
+    | {
+          status: "partial"
+          succeeded: number
+          failed: number
+          total: number
```
kaloyanvi: Nit: do we need all 3, we can derive the total right?

plugins/notion/src/data.ts
```diff
@@ -371,16 +384,34 @@ export async function syncCollection(
...
+    // previous checkpoint makes those items eligible for the next sync.
+    if (syncErrors.length === 0) {
+        pluginDataUpdates.push(collection.setPluginData(PLUGIN_KEYS.LAST_SYNCED, syncStartedAt))
+    }
+
+    await Promise.all(pluginDataUpdates)
+
+    if (syncErrors.length > 0) {
+        return {
+            status: "partial",
```
kaloyanvi: Nit: it's bit strange that if all items error we still have "partial" status even though none succeeded

kaloyanvi (approved):
LGTM! Left two nits

plugins/notion/src/data.ts
iamakulov: good one, ty! killed `total`

plugins/notion/src/data.ts
iamakulov: yup, tuned to `completed` and `completed-with-errors`. i thiiiiink it’s fine to treat “completed with 0 successful items” and “errored due to an exception” differently; the UX of both scenarios is pretty similar right now, and i’m not feeling like adding extra complexity to treat the former as the latter.

PR #686, opened 2026-08-18
Locale sync - fix possible out of memory export crash

djohalo2:
### Description

To fix a possible out of memory crash in the Locale Sync plugin export, this pull request implements the new `listLocalizationGroups` method to get localization groups using async iterator when exporting. While at it I also added a loading state to the button.

### Changelog

- Fixed possible out of memory crash for Locale Sync plugin export.

### Testing

- [x] Open project with a lot of localization data and verify export works as expected
    - Also verify you can see a loading spinner and the button is disabled while exporting


plugins/locale-sync/src/App.tsx
```diff
@@ -72,15 +68,26 @@ export function App() {
...
+        setIsExporting(true)
+
+        try {
+            if (!targetLocale) {
+                throw new Error(`Could not find locale with id ${selectedLocaleId}`)
+            }
 
-        void exportXliff(defaultLocale, targetLocale)
+            await exportXliff(defaultLocale, targetLocale, filename)
+            framer.notify(`Successfully exported ${filename}`)
```
elmarburke: suggestion: If the export is successful, you get the file directly. I think we don't need to notify then? Consider removing this.

plugins/locale-sync/src/App.tsx
```diff
@@ -72,15 +68,26 @@ export function App() {
...
+        } catch (error) {
+            console.error(error)
+            framer.notify(`Error exporting ${filename}`, { variant: "error" })
```
elmarburke: polish: the error message doesn't read good to me. Consider something like `Failed to save.`

plugins/locale-sync/src/App.tsx
```diff
@@ -72,15 +68,26 @@ export function App() {
...
+
+        try {
+            if (!targetLocale) {
+                throw new Error(`Could not find locale with id ${selectedLocaleId}`)
+            }
 
-        void exportXliff(defaultLocale, targetLocale)
+            await exportXliff(defaultLocale, targetLocale, filename)
+            framer.notify(`Successfully exported ${filename}`)
+        } catch (error) {
```
elmarburke, line 87: issue: The error that you throw a few lines above will be catched here. Should we handle that one differently? When will it trigger?

elmarburke (approved):

PR #687, opened 2026-08-21
Fix HubSpot blog author name

djohalo2:
### Description

This pull request fixes an issue where hubspot blog author name was synced with last editor name instead of the actual linked author name. The field `authorName` is a bit confusing, but according to their docs indeed something else than what you would expect (https://developers.hubspot.com/docs/api-reference/latest/cms/blogs/posts/create-post#body-author-name)

To keep this change backwards compatible, we keep using the original `authorName` field id, but populate it with the fetched blog author name.

### Changelog

- Fixed synced blog posts showing the last editor name instead of the linked author name.

### Testing

- [x] Sync a hubspot blog post and verify the author name is now populated with actual blog author that you linked to the post and not the last updated name

### DIFF
--- a/plugins/hubspot/src/blog.ts
+++ b/plugins/hubspot/src/blog.ts
@@ -242,15 +242,22 @@ export async function syncBlogs({ fields, includedFieldIds }: SyncBlogMutation) 
 
     const fieldsById = new Map(fields.map(field => [field.id, field]))
     const unsyncedItemIds = new Set(await collection.getItemIds())
+    const hasAuthorNameField = fieldsById.has(AUTHOR_NAME_FIELD_ID)
     const requestedProperties = new Set([...includedFieldIds, "id", "slug"])
 
-    if (fieldsById.has(AUTHOR_NAME_FIELD_ID)) requestedProperties.add(BLOG_AUTHOR_ID_FIELD_ID)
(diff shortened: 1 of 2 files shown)


plugins/hubspot/src/blog.ts
```diff
@@ -232,15 +242,20 @@ export async function syncBlogs({ fields, includedFieldIds }: SyncBlogMutation)
 
     const fieldsById = new Map(fields.map(field => [field.id, field]))
     const unsyncedItemIds = new Set(await collection.getItemIds())
+    const requestedProperties = new Set([...includedFieldIds, "id", "slug"])
+
+    if (fieldsById.has(AUTHOR_NAME_FIELD_ID)) requestedProperties.add(BLOG_AUTHOR_ID_FIELD_ID)
+
+    const { results: posts } = await fetchAllBlogPosts(MAX_CMS_ITEMS, Array.from(requestedProperties))
+
+    const authorIds = Array.from(new Set(posts.map(post => post.blogAuthorId)))
```
huntercaron: Should we also gate this to only when we are requesting the blog author id?

huntercaron (approved):
Silly that we have to do this at all but nice fix thank you!

djohalo2:
@huntercaron Yeah their property name is very misleading...

PR #688, opened 2026-08-24
HubSpot plugin - handle invalid blog author references

djohalo2:
### Description

This pull request gracefully handles invalid blog author references on a post and fallbacks to last updated by. According to HubSpot this shouldn't be possible and they enforce that a blog post must have a valid blog author. However, drafts are also synced and those might not have a valid blog author id.

### Changelog

- Gracefully handles invalid or missing blog author references on a blog post and fallbacks to last updated by.

### Testing

- [x] Validate that triggering a 404 on getting a blog post author should no longer fail the sync, but fallbacks to last updated at (authorName)

### DIFF
--- a/plugins/hubspot/src/blog.ts
+++ b/plugins/hubspot/src/blog.ts
@@ -251,8 +251,14 @@ export async function syncBlogs({ fields, includedFieldIds }: SyncBlogMutation) 
 
     const authorNamesById = new Map<string, string>()
     if (hasAuthorNameField) {
-        const authorIds = Array.from(new Set(posts.map(post => post.blogAuthorId)))
-        const authors = authorIds.length ? await Promise.all(authorIds.map(fetchBlogAuthor)) : []
+        // Blog posts should have a blogAuthorId, but drafts or unexpected data may not.
+        // Gracefully fall back to the last updated author name in those cases.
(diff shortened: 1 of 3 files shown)


plugins/hubspot/src/blog.ts
```diff
@@ -251,8 +251,14 @@ export async function syncBlogs({ fields, includedFieldIds }: SyncBlogMutation)
 
     const authorNamesById = new Map<string, string>()
     if (hasAuthorNameField) {
-        const authorIds = Array.from(new Set(posts.map(post => post.blogAuthorId)))
-        const authors = authorIds.length ? await Promise.all(authorIds.map(fetchBlogAuthor)) : []
+        // Blog posts should have a blogAuthorId, but drafts or unexpected data may not.
+        // Gracefully fall back to the last updated author name in those cases.
+        const authorIds = Array.from(
+            new Set(
```
elmarburke, line 257: **issue**: `Set`s (and `Map`s) are, as per specification, without a defined order and it's order can't be lied on. You already have a filter step where you can deduplicate from.

elmarburke: nvm, you don't rely on the order anyway…

elmarburke (approved):

